PR #3, opened 2021-01-20
Rename seq_id to contig

alimanfoo:
This PR proposes renaming the `seq_id` argument to `contig` instead, because it's a bit more user friendly (intuitive) and also a bit more consistent with the naming conventions that sgkit are following.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -276,12 +276,12 @@ class Ag3:
             self._cache_site_filters[key] = root
             return root
 
-    def site_filters(self, seq_id, mask, field="filter_pass", analysis="dt_20200416"):
+    def site_filters(self, contig, mask, field="filter_pass", analysis="dt_20200416"):
         """Access SNP site filters.
 
         Parameters
         ----------
-        seq_id : str
+        contig : str
             Chromosome arm, e.g., "3R".
         mask : {"gamb_colu_arab", "gamb_colu", "arab"}
             Mask to use.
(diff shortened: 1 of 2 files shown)


alimanfoo:
cc @hardingnj, @cclarkson, @leehart, let me know if any objections

leehart:
No objections, only curiosity about the semantic appropriateness.

We seem to be using `seq_id` and `contig` anywhere that we mean "chromosome arm".

Although I might expect `contig` to simply be an abbreviation of `contiguous`, in the context of sequences, the term's definition in this field does not seem completely intuitive (to me, at least) and it seems to have a somewhat muddy/cryptic etymology, judging from https://en.wikipedia.org/wiki/Contig

But then, I suppose the same can be said of `seq_id` and `chrom`.

alimanfoo:
Thanks @leehart. Yeah, no term is perfect here. 

In gambiae, 2L, 2R, 3L and 3R are chromosome arms, but X is a chromosome.

In most other anophelines where the reference genome has not been assembled to chromosome (or chromosome arm) level, the reference genome is said to be made up of contigs, which are just contiguous pieces of sequence.

The [GFF3 format](http://gmod.org/wiki/GFF3) tries to cope with this variability by using "seqid" to mean the name of a chromosome or scaffold or contig or whatever sequence you are using to anchor your coordinate system. That's why I started using "seq_id" as a variable name in python code.

But here I'm wondering about going back to using "contig" because it's a bit more intuitive, and can reasonably mean contig or chromosome.

PR #4, opened 2021-01-20
Replace loc based logic for species assignment with function

hardingnj:
Hi Alistair- 

I just quickly had a go at rewriting logic around species assignment. I think its easier to follow when contained in a function like this?


alimanfoo:
Hi @hardingnj, looks good. Nit, but I would suggest to let the consolidate_species() function take a pandas series, so you could avoid the intermediate lambda function. I.e., so you could do:

```
df["species"] = df.apply(consolidate_species)
```

hardingnj:
Sorry- missed this first time around. Will amend this now.

alimanfoo:
No worries, thanks :smile:

alimanfoo (approved):

PR #16, opened 2021-03-06
Accelerate tests 

alimanfoo:
This PR does the following:

* Accelerates tests by adding support for using simplecache layer 
* Postpones lookup of releases unless needed
* Simplifies handling of url


alimanfoo:
Hey @cclarkson, happy to talk this through on Monday.

cclarkson (approved):
Looks good to me :+1:

PR #17, opened 2021-03-12
Ag3 SNP xarray dataset enhancements

### LINKED ISSUE #13: Add data variables to Ag3.snp_dataset()
Proposed to add some additional variables to the dataset returned by the `Ag3.snp_dataset()` method, including:

* `sample_...` variables with sample metadata
* `call_GQ`, `call_AD`, `call_MQ`
* `variant_filter_pass_...` with site filters arrays

alimanfoo:
This PR makes some enhancements to the Ag3 method which returns an xarray dataset holding data on SNP calls. 

* Rename `Ag3.snp_dataset()` to `Ag3.snp_calls()` - in future I think we'll be returning xarray datasets from other methods, and we don't want to always have to add "_dataset" to the end of every method name that does so.
* Add additional data variables `call_GQ`, `call_AD`, `call_MQ`, `variant_filter_pass_...` to the dataset.
* `Ag3.snp_calls()` now sets up it's own arrays, rather than depending on `snp_sites()` and `snp_genotypes()` - this simplifies the handling of multiple sample sets, which uses `xarray.concat()`.

Btw I decided **not** to include any sample metadata variables for the time being. May revisit that in future.

Resolves #13.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -650,57 +650,66 @@ class Ag3:
 
         return d
 
-    def snp_dataset(
-        self,
-        contig,
-        sample_sets="v3_wild",
-        species_calls=("20200422", "aim"),
-        site_mask=None,
-        site_filters="dt_20200416",
-    ):
-        """TODO doc me"""
+    def _snp_calls_dataset(self, contig, sample_set):
 
-        # TODO support multiple contigs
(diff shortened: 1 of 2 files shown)


alimanfoo:
cc @hardingnj, @cclarkson, @leehart, @jonbrenas

PR #26, opened 2021-03-26
Ag3 snp effects updates

cclarkson:
- fixes bug in intron code (we weren't getting splice site effects due to `else` instead of `elif`.
- removes the single snp effect method/tests, on reflection this wasn't adding anything helpful.
- Adds intron tests.

### DIFF
--- a/malariagen_data/veff.py
+++ b/malariagen_data/veff.py
@@ -935,11 +935,10 @@ def _get_within_intron_effect(annotator, base_effect, intron, exons):
         # SNPs
 
         if intron_min_dist <= 2:
-
             # splice site variation
             effect = base_effect._replace(effect="SPLICE_CORE", impact="HIGH")
 
-        if intron_min_dist <= 7:
+        elif intron_min_dist <= 7:
 
             # splice site variation
             effect = base_effect._replace(effect="SPLICE_REGION", impact="MODERATE")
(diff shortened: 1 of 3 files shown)


malariagen_data/veff.py
```diff
@@ -935,11 +935,10 @@ def _get_within_intron_effect(annotator, base_effect, intron, exons):
         # SNPs
 
         if intron_min_dist <= 2:
-
             # splice site variation
             effect = base_effect._replace(effect="SPLICE_CORE", impact="HIGH")
 
-        if intron_min_dist <= 7:
+        elif intron_min_dist <= 7:
```
alimanfoo, line 941: Good catch.

alimanfoo (commented):
Thanks Chris, looks good. Just one comment re naming.

tests/test_ag3.py
```diff
@@ -503,81 +494,85 @@ def test_snp_effects():
...
+    assert expected_fields == df.columns.tolist()
+    assert df.shape == (2829, 11)
+
+    # check first, second, third codon position non-syn
+    assert df.iloc[701].aa_change == "E35*"
+    assert df.iloc[703].aa_change == "E35V"
+    # while we are here, check all columns for a position
+    assert df.iloc[706].position == 28600605
+    assert df.iloc[706].ref_allele == "G"
+    assert df.iloc[706].alt_alleles == "C"
```
alimanfoo, line 542: Can we rename this to singular "alt_allele".

cclarkson:
@alimanfoo - I've renamed the column. Should be good to go now.

alimanfoo (approved):

PR #30, opened 2021-04-21
Profile veff with snakeviz demo

alimanfoo:
Hi @cclarkson, here's the notebook I used to profile snp_effects with snakeviz. Note that there is still some weirdness here with snakeviz where it sometimes doesn't show the full icicle chart, but if so you can fix that by scrolling to the table below, sorting by cumtime, then clicking on ag3.snp_effects() row, which will set that as the root of the icicle chart.


alimanfoo:
Served it's purpose.

PR #32, opened 2021-05-05
Optimise snp effects

cclarkson:
We optimised the snp_effect method and removed the unused code.

PR #33, opened 2021-05-10
Optimise snp effects - suggestions alimanfoo 20210510

alimanfoo:
This PR contains suggestions to the #32 PR.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -651,11 +651,7 @@ class Ag3:
 
         # get feature direct from geneset
         gs = self.geneset()
-        gs.rename(
-            columns={"ID": "feature_id", "Parent": "parent_id", "end": "stop"},
-            inplace=True,
-        )
-        feature = gs[gs.feature_id == transcript].squeeze()
+        feature = gs[gs["ID"] == transcript].squeeze()
 
         # grab pos, ref and alt for chrom arm from snp_sites
         pos, ref, alt = self.snp_sites(
@@ -666,7 +662,7 @@ class Ag3:
         pos = allel.SortedIndex(pos.compute())
 
         # locate transcript range
-        loc = pos.locate_range(feature.start, feature.stop)
+        loc = pos.locate_range(feature.start, feature.end)
 
         # dask compute on the sliced arrays to speed things up
         pos = pos[loc]
@@ -690,9 +686,7 @@ class Ag3:
             )
         ann = self._cache_annotator
 
-        df_effects = ann.get_only_transcript_effects(
-            transcript=transcript, variants=df_variants
-        )
+        df_effects = ann.get_effects(transcript=transcript, variants=df_variants)
 
         return df_effects
 
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -631,15 +631,17 @@ class Ag3:
 
         return is_accessible
 
-    def snp_effects(self, transcript, site_mask):
+    def snp_effects(self, transcript, site_mask=None, site_filters="dt_20200416"):
         """Compute variant effects for a gene transcript.
 
         Parameters
         ----------
         transcript : str
             Gene transcript ID (AgamP4.12), e.g., "AGAP004707-RA".
-        site_mask : {"gamb_colu_arab", "gamb_colu", "arab"}
+        site_mask : {"gamb_colu_arab", "gamb_colu", "arab"}, optional
             Site filters mask to apply.
+        site_filters : str, optional
+            Site filters analysis version.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


malariagen_data/ag3.py
```diff
@@ -649,43 +651,42 @@ def snp_effects(self, transcript, site_mask):
...
-        df_in["alt_allele"] = [list(q.tobytes().decode()) for q in list(alt)]
-        # explode the alt alleles into their own rows
-        variants = df_in.explode("alt_allele").reset_index(drop=True)
+        cols = {
+            "position": np.repeat(pos, 3),
+            "ref_allele": np.repeat(ref.astype("U1"), 3),
+            "alt_allele": alt.astype("U1").flatten(),
+        }
+        df_variants = pandas.DataFrame(cols)
+
```
alimanfoo, line 679: Construct initial dataframe the same way as in snp_allele_frequencies().

malariagen_data/veff.py
```diff
@@ -121,25 +92,21 @@ def get_ref_allele_coords(self, chrom, pos, ref):
 null_effect = VariantEffect(*([None] * len(VariantEffect._fields)))
 
 
-def get_only_transcript_effects(ann, transcript, variants):
+def get_effects(ann, transcript, variants):
```
alimanfoo, line 95: Renamed for simplicity.

malariagen_data/veff.py
```diff
@@ -121,25 +92,21 @@ def get_ref_allele_coords(self, chrom, pos, ref):
...
 
-    children = ann.get_children(transcript)
+    children = ann.get_children(transcript).sort_values("start")
```
alimanfoo, line 97: Can sort here, then don't need to sort below.

malariagen_data/veff.py
```diff
@@ -213,18 +179,18 @@ def get_only_transcript_effects(ann, transcript, variants):
 
 
 # add as method
-Annotator.get_only_transcript_effects = get_only_transcript_effects
+Annotator.get_effects = get_effects
 
 
-def _get_within_transcript_effects(ann, base_effect, cdss, utr5, utr3, introns):
```
alimanfoo, line 219: Renamed to singular - only one effect is returned.

malariagen_data/veff.py
```diff
@@ -533,7 +496,7 @@ def _get_within_intron_effect(base_effect, intron):
     ref = base_effect.ref
     alt = base_effect.alt
     intron_start, intron_stop = intron
-    strand = base_effect.gene_strand
```
alimanfoo, line 536: Caught a bug here, we weren't actually populating this field, so was None. Have renamed to "strand" for simplicity and made sure we populate it above.

pyproject.toml
```diff
@@ -12,8 +12,6 @@ zarr = "*"
 dask = {version="*", extras=["array"]}
 fsspec = "*"
 gcsfs = "*"
-petl = "*"
-petlx = "*"
```
alimanfoo, line 16: No longer needed since pandarisation.

malariagen_data/veff.py
```diff
@@ -86,29 +79,7 @@ def get_ref_allele_coords(self, chrom, pos, ref):
...
-        "intron_stop",
-        "intron_5prime_dist",
-        "intron_3prime_dist",
-        "intron_exon_5prime",
-        "intron_exon_3prime",
-        "ref_cds_start",
-        "ref_cds_stop",
-        "ref_intron_start",
-        "ref_intron_stop",
-        "ref_start_phase",
```
alimanfoo, line 111: We aren't using any of these in the output dataframe, so I removed them here. If we need them later we can always put them back.

malariagen_data/veff.py
```diff
@@ -27,19 +25,14 @@ def __init__(self, genome, geneset):
         self._genome_cache = dict()
         self._geneset_cache = None
 
-        geneset.rename(
-            columns={"ID": "feature_id", "Parent": "parent_id", "end": "stop"},
-            inplace=True,
-        )
```
alimanfoo, line 33: I thought it would be simpler to do away with the renaming altogether, just use columns as named.

malariagen_data/ag3.py
alimanfoo, line 679: Changed to construct initial dataframe the same way as in snp_allele_frequencies().

malariagen_data/ag3.py
```diff
@@ -649,43 +651,42 @@ def snp_effects(self, transcript, site_mask):
 
         # get feature direct from geneset
         gs = self.geneset()
-        gs.rename(
-            columns={"ID": "feature_id", "Parent": "parent_id", "end": "stop"},
-            inplace=True,
-        )
```
alimanfoo, line 655: Deleted as this is not needed here.

malariagen_data/ag3.py
```diff
@@ -631,15 +631,17 @@ def is_accessible(self, contig, site_mask, site_filters="dt_20200416"):
 
         return is_accessible
 
-    def snp_effects(self, transcript, site_mask):
+    def snp_effects(self, transcript, site_mask=None, site_filters="dt_20200416"):
```
alimanfoo, line 634: This change makes parameters and defaults follow snp_allele_frequencies().

alimanfoo (commented):
Some commentary on the changes in this PR...

PR #38, opened 2021-05-12
Add methods for computing copy number variation by gene

alimanfoo:
This PR adds two methods for computing copy number variation by gene, based on the HMM data:

* Ag3.gene_cnv() - compute modal copy number for each sample by gene, for all genes in a contig
* Ag3.gene_cnv_frequencies() - compute the population frequencies of amplifications and deletions, by gene, for all genes in a contig

Also includes some updates to CI and dev environment.


alimanfoo:
Going ahead with merge but happy to address any comments in a followup.

PR #45, opened 2021-05-13
Maintenance on SNP effects and allele frequencies functions

### LINKED ISSUE #36: Veff maintenance
There's a couple of further pieces of maintenance/tidying we could do on the veff module:

* Currently if a variant falls within an intron, we're calling `_get_intron_effect()`, but we could shortcut and call directly to `_get_within_intron_effect()`. This should provide a little speedup, and we could probably also delete `_get_intron_effect()` completely.
* We currently define `get_effects()` as a function, then attach it as a method on the `Annotator` class manually. This is a bit unconventional and makes IDEs like pycharm complain unnecessarily. We could be a bit more conventional here and just move the functions to be methods.

### LINKED ISSUE #37: Consistify seqid/contig
Currently we use the variable name "contig" throughout to mean a sequence identifier from a reference genome. However, the GFF loading returns a dataframe with the column name "seqid". We could consistify here, by renaming the GFF dataframes to use "contig" instead.

### LINKED ISSUE #41: Renaming of geneset dataframe columns is unnecessary and altering cached dataframe
https://github.com/malariagen/malariagen-data-python/blob/44bb5ab31ca3693ed2d9cb743b3fd77a2e367048/malariagen_data/ag3.py#L735

### LINKED ISSUE #42: Ag3.geneset() add description to default columns
We often use this, add to defaults.

### LINKED ISSUE #43: Ag3.snp_allele_frequencies() rename "maximum" to "max_af" for clarity
Suggest to rename this column to avoid the very general "maximum".

alimanfoo:
This PR has several pieces of maintenance, which came up as worth doing while working on the SNP effects example notebook. Includes:

* clean up intron effects and make get_effects() a method (resolves #36)
* rename seqid to contig (resolves #37)
* accelerate snp_effects and snp_allele_frequencies with site mask
* fix unnecessary geneset renaming (resolves #41)
* add description to default cols (resolves #42)
* rename maximum to max_af (resolves #43)

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -406,7 +406,8 @@ class Ag3:
         contig : str
             Chromosome arm, e.g., "3R".
         field : {"POS", "REF", "ALT"}, optional
-            Array to access. If not provided, all three arrays POS, REF, ALT will be returned as a tuple.
+            Array to access. If not provided, all three arrays POS, REF, ALT will be returned as a
+            tuple.
         site_mask : {"gamb_colu_arab", "gamb_colu", "arab"}
             Site filters mask to apply.
         site_filters : str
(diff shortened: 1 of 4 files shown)


alimanfoo:
Hi @cclarkson, sorry to bundle a few things in together here, but hopefully not too difficult to parse. Let's talk through at 2.

alimanfoo:
Going ahead with merge but happy to address any comments in a follow-up.

PR #48, opened 2021-05-18
Add site filter columns to SNP dataframes

alimanfoo:
This PR makes a small change to the `snp_effects` and `snp_allele_frequencies` methods to add the site mask columns into the resulting dataframe. This allows the user to retrieve all SNPs but then see which pass which masks if needed. Useful especially where site filters are too strict and interesting resistance SNPs have been filtered out.

Also some refactoring to move shared code to set up a SNP dataframe to a private utility method.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -638,15 +638,22 @@ class Ag3:
 
         return is_accessible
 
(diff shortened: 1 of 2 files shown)


alimanfoo:
Hi @cclarkson, here's the work I mentioned this morning. Any comments or questions give me a shout.

PR #53, opened 2021-06-15
Reverse expected/actual in tests

### LINKED ISSUE #35: Reverse expected/actual in tests
The convention for pytest in assertions seemds to be to put actual first then expected second, the opposite of what I previously thought. Pycharm follows this in how it reports errors. So to make the pytest and pycharm outputs make more sense, we should swap round the compared variables in all test assertions.

cclarkson:
Resolves #35


alimanfoo (approved):
Thanks so much @cclarkson, looks much better now :-)

PR #58, opened 2021-08-12
Cohort access

cclarkson:
resolves #51 

Adds a method `Ag3.sample_cohorts(sample_sets, analysis)` to access the Ag3 defined cohorts metadata. Tests have also been written for the method, but currently these point at the staging bucket. Once the metadata is copied to the release bucket, the tests should be updated.

PR #61, opened 2021-09-24
Adds cohort access to SNP and CNV frequency methods

### LINKED ISSUE #57: Add cohort access to Ag3.snp_allele_frequencies
---
Split from issue #51 

Modify method `Ag3.snp_allele_frequencies(transcript, cohorts, cohorts_analysis, ...)`:

* Allow the `cohorts` parameter to be a dict (current behaviour continues to be supported) or a string. If a string, it must be the name of one of the columns in the cohorts dataframe.
* Add a `cohorts_analysis` parameter which is a string with default value set to latest cohorts analysis. This will be used if the `cohorts` parameter is given as a string.
* Other parameters left as-is.

Open questions:

* Should this enforce a minimum sample size? Or give that as an option?

Example:

```python
df_snp_af = ag3.snp_allele_frequencies(transcript="AGAP006028-RA", cohorts="admin1_month")
```

  * Use parameter name "cohort_analysis" throughout for consistency
  * Add "min_cohort_size" parameter to the snp_allele_frequencies() and gene_cnv_frequencies() method, default value 10

cclarkson:
resolves #57, #58

### DIFF
--- a/tests/test_ag3.py
+++ b/tests/test_ag3.py
@@ -679,7 +679,7 @@ def test_cohort_snp_allele_frequencies():
     )
 
     assert isinstance(df, pd.DataFrame)
-    assert df.shape == (16904, 101)
+    assert df.shape == (16639, 101)
 
 
 def test_snp_allele_frequencies_0_cohort():
--- a/tests/test_ag3.py
+++ b/tests/test_ag3.py
@@ -664,6 +664,25 @@ def test_snp_allele_frequencies():
     assert np.any(df.max_af == 0)
 
 
+def test_cohort_freq():
+    ag3 = setup_ag3()
+    cohorts = "admin1_month"
+
+    df = ag3.snp_allele_frequencies(
+        transcript="AGAP004707-RD",
+        cohorts=cohorts,
+        cohorts_analysis="20210702",
+        min_cohort_size=10,
+        site_mask="gamb_colu",
+        sample_sets="v3_wild",
+        drop_invariant=False,
+    )
+
+    assert isinstance(df, pd.DataFrame)
+    # assert df.columns.tolist() == expected_fields
+    # assert df.shape == (132306, len(expected_fields))
+
+
 def test_snp_allele_frequencies_0_cohort():
     ag3 = setup_ag3()
     cohorts = {
--- a/tests/test_ag3.py
+++ b/tests/test_ag3.py
@@ -1026,13 +1026,21 @@ def test_gene_cnv(contig, sample_sets):
 
 
 @pytest.mark.parametrize("contig", ["2R", "X"])
-def test_gene_cnv_frequencies(contig):
-    ag3 = setup_ag3()
-    cohorts = {
-        "ke": "country == 'Kenya'",
-        "bf_2012_col": "country == 'Burkina Faso' and year == 2012 and species == 'coluzzii'",
-    }
-    expected_cols = [
+@pytest.mark.parametrize(
+    "cohorts",
+    [
+        {
+            "ke": "country == 'Kenya'",
+            "bf_2012_col": "country == 'Burkina Faso' and year == 2012 and species == 'coluzzii'",
+        },
+        {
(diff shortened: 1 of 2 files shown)


tests/test_ag3.py
```diff
@@ -587,12 +587,21 @@ def test_snp_effects():
...
+            "ke": "country == 'Kenya'",
+            "bf_2012_col": "country == 'Burkina Faso' and year == 2012 and species == 'coluzzii'",
+        },
+        {
+            "bf_2050_col": "country == 'Burkina Faso' and year == 2050 and species == 'coluzzii'"
+        },
+        "admin1_month",
+    ],
+)
+def test_snp_allele_frequencies(cohorts):
```
alimanfoo, line 603: Perhaps this is where the wizardry thingy was suggesting using separate functions? There is no point in using parametrize here, because there are completely separate code branches for each of the different values of the `cohorts` parameter. It would be clearer to have three separate test functions, e.g., named `test_snp_allele_frequencies__no_samples()`, `test_snp_allele_frequencies__str_cohorts()` and `test_snp_allele_frequencies__dict_cohorts()` or something like that.

tests/test_ag3.py
```diff
@@ -1026,44 +1036,57 @@ def test_gene_cnv_frequencies(contig):
...
+                a = df[fa].values
+                d = df[fd].values
+                x = a + d
+                assert np.all(x >= 0)
+                assert np.all(x <= 1)
+                assert df.columns.tolist() == expected_fields
+                if contig == "2R":
+                    assert df.shape == (3668, 10)
+                elif contig == "X":
+                    assert df.shape == (1063, 10)
```
alimanfoo, line 1075: This can be removed, because it is already covered by better cross-checks above. I.e., the length of the dataframe is checked by comparing with the length of `df_genes`, and the columns are checked by comparing to `expected_fields`.

tests/test_ag3.py
```diff
@@ -1026,44 +1036,57 @@ def test_gene_cnv_frequencies(contig):
         "bf_2012_col_amp",
         "bf_2012_col_del",
```
alimanfoo, line 1037: These values depend on which `cohorts` are being provided. Would be better to define e.g., `common_fields` here, then add on what you expect based on what's in the `cohorts`.

tests/test_ag3.py
```diff
@@ -1026,44 +1036,57 @@ def test_gene_cnv_frequencies(contig):
...
+                    assert df.shape == (1063, 10)
+
+        if isinstance(cohorts, str):
+            df_coh = ag3.sample_cohorts(
+                sample_sets="v3_wild", cohorts_analysis="20210702"
+            )
+            coh_nm = "cohort_" + cohorts
+            all_uni = df_coh[coh_nm].dropna().unique()
+            coh_columns = np.asarray([a[:-4] for a in df.columns])
+            assert all(np.in1d(all_uni, coh_columns))
```
alimanfoo, line 1084: This is sketchy. You should be able to construct exactly the `expected_fields` from `all_uni` and then you could compare via equality.

tests/test_ag3.py
```diff
@@ -1026,44 +1036,57 @@ def test_gene_cnv_frequencies(contig):
...
+            if contig == "2R":
+                assert df.shape == (3668, 192)
+            elif contig == "X":
+                assert df.shape == (1063, 192)
```
alimanfoo, line 1088: You should be able to remove this, if you test for the expected fields exactly, and because the length is already checked above.

alimanfoo (commented):
Hi Chris, looking good. 

A few small suggestions in the implementation. 

On the testing side, re when to use `pytest.parametrize` the key question is whether it means you can avoid duplicating testing code. I.e., you should be able to run at least some code under all values of the variables you are parametrizing, other there is no point and witchery is right you're better off to break up into separate functions. I think this applies to the `test_snp_allele_frequencies()` function.

Re `test_gene_cnv_frequencies()` there's a couple of things could be tightened up. More than happy to discuss or do paired if not immediately obvious.

alimanfoo (approved):
Awesome :bouquet:

PR #69, opened 2021-10-11
Add pf7 module

kathrynmurie:


kathryn1995:
FYI @alimanfoo @cclarkson @podpearson @roamato

cclarkson:
Hi @kathryn1995, apologies this took so long to get to. 

I've a couple of suggestions:

- Now that the Pf7 data in staging bucket, test_pf7.py could point a real data rather than building mock data.
- It might be worth looking at the new [Amin1](https://github.com/malariagen/malariagen-data-python/blob/master/malariagen_data/amin1.py) part of the module. The `snp_calls` method looks particularly relevant to pf7.py, as this pulls together useful calldata into an xarray.

Give me a shout when you'd like another review, I promise it won't take as long as last time!

kathryn1995:
> 
Hi @cclarkson, thanks for taking a look. 
I have added a new test file to run the tests using the real data in the staging bucket. I wanted to keep it separate from the unit tests so it is clear whether it is the code that is failing or something in the process of getting the data, hope that's okay. Will take a look at the new part of the module shortly!

cclarkson:
Hi @kathryn1995,

Cool, getting the data is fairly straightforward, so I think using the your tests going forward makes sense. Give me a shout if you want to discuss Amin1 `snp_calls`, I think that encapsulates what the Pf7 API is trying to achieve quite neatly.

alimanfoo:
Hi folks, just to chip in here and say it might be worth a little discussion of the public API design, to get some feedback on different approaches. 

To surface some thinking around the [Amin1 API](https://malariagen.github.io/vector-data/amin1/api.html), that has methods as follows (leaving out a few non-essential details):

* `sample_metadata() -> pandas DataFrame`
* `snp_calls(contig) -> xarray Dataset`

The return value of `snp_calls()` is a dataset that follows the [sgkit naming conventions](https://pystatgen.github.io/sgkit/latest/getting_started.html#data-structures) for arrays and dimensions. Data for multiple contigs can be obtained by passing a list of strings as the value of the `contig` parameter.

Although I'm not using sgkit much yet, I'm trying to align with it so it's easy to start using it when we want to try it out more. Ultimately I'd like to migrate to sgkit instead of scikit-allel, because sgkit is where the development and maintenance effort is now. It will take a while though, and it's also quite straightforward to use scikit-allel with xarray, so it's a nice data structure that gives options. 

I'm also starting to use xarray more to handling indexing and selecting variants and/or samples, because the indexing API is a bit more intuitive, and you can apply selections to a whole dataset - [example](https://malariagen.github.io/vector-data/ag3/examples/snps.html#selecting-sites-and-samples).

The [Ag3 API](https://malariagen.github.io/vector-data/ag3/api.html) is essentially the same, except that these methods have a `sample_sets` parameter, but that doesn't apply to Pf7, so `Amin1` is probably a better comparison.

Would there be any objection to following a similar pattern for Pf7? One sticking point is probably the naming of the `snp_calls()` method, as with Pf7 there are indels too. Alternative naming suggestions very welcome.

kathryn1995:
Hi @alimanfoo @cclarkson, 

I have changed the code to use xarray in the same way as amin1. Instead of using `snp` in the names, I have changed this to `variant` as discussed. 
There are two test files, one containing unit tests for the functions and a separate one for integration tests dependent on the data on the cloud. 
I have also added an extra option to the variant_calls() method so users can choose to extend the xarray to contain everything from the zarr, without this, there is just a set of default variables that are most commonly used for analysis

malariagen_data/pf7.py
```diff
@@ -0,0 +1,278 @@
...
+            )
+        ]
+        ds = xarray.concat(
+            datasets,
+            dim=DIM_VARIANT,
+            data_vars="minimal",
+            coords="minimal",
+            compat="override",
+            join="override",
+        )
```
alimanfoo, line 276: There's no need to concatenate here, there's only ever one dataset.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,278 @@
...
+DIM_GENOTYPES = "genotypes"
+
+
+class Pf7:
+    def __init__(
+        self,
+        url,
+        data_config=None,
+        extended_calldata_variables=None,
+        extended_variant_fields=None,
```
alimanfoo, line 30: I'm not sure it's worth having these parameters. It would seem fine to me just to have a fixed set of extended fields, defined at the module level.

alimanfoo (commented):
Couple of initial comments, will follow up with a more complete review.

malariagen_data/pf7.py
kathryn1995: I added these to make testing the extended function a bit easier. Instead of mocking the entire extended zarr I could just add a few extra variables. Similar to the `data_config` parameter, I thought most users wouldn't touch them, but more advanced users could specify the subset if they wanted to. Happy to change if you think this overcomplicates it thought

malariagen_data/pf7.py
kathryn1995: Sorry, I meant to do this but forgot. I have removed the concatenation now. I still have `variant_calls` as a separate function that calls `_variant_dataset` though, do you think it makes sense to rename `_variant_dataset` to `variant_calls` and just have one function or keep them separate in case we add any options to filter the dataset?

alimanfoo: I'd say just have one function, no point in having two functions if they're not actually doing anything different. Can always split out later if it serves some purpose.

malariagen_data/util.py
```diff
@@ -119,10 +119,8 @@ class SiteClass(Enum):
 
 def da_from_zarr(z, inline_array, chunks="auto"):
     """Utility function for turning a zarr array into a dask array.
-
     N.B., dask does have it's own from_zarr() function but we roll
     our own here to get a little more control.
-
     """
     if chunks == "native" or z.dtype == object:
         # N.B., dask does not support "auto" chunks for arrays with object dtype
```
alimanfoo, line 126: Did you mean to remove this whitespace?

malariagen_data/util.py
```diff
@@ -140,19 +138,15 @@ def da_from_zarr(z, inline_array, chunks="auto"):
...
     ----------
     ds : xarray.Dataset
     indexer : str
     dim : str
-
     Returns
     -------
     xarray.Dataset
-
     """
```
alimanfoo, line 150: Did you mean to remove this whitespace?

alimanfoo (commented):
Couple of nits around whitespace.

malariagen_data/pf7.py
alimanfoo: I'd suggest to remove them, keep things simple. Would be good to have a conversation about the mocking tests, I'll put some thoughts about that in a separate comment.

alimanfoo:
Hi @kathryn1995, just planting a thought for discussion, I'm wondering whether the unit tests using mocking are worth the maintenance overhead. I'm guessing you wanted those while you were developing and before you had data in the cloud. But most of what they're testing is pretty basic stuff, and won't increase coverage over the tests against the real data. I know this might be painful because you put work into them, but can we consider dropping them?

malariagen_data/pf7.py
kathryn1995: I have merged _variant_dataset and variant_calls into one function

malariagen_data/util.py
kathryn1995: No, I think it might have been my formatter. I have added them back in now.

kathryn1995:
@alimanfoo I have removed some of the unit tests that were more or less replicating the code. I have also refactored the code a little bit as the variant_calls function was getting too complicated with more than one thing going on. I have pulled out adding the coordinates, the data variables, and extending the data set into separate functions. I have also cut down on the amount of repeated code.

@podpearson @ahernank @roamato
I have added the option to define the extended set that you want with a list. When calling the `variant_calls` function you can either set the extended parameter to...

- `[]` : This is the default and will just return the default set with none of the extended variables.
- `"*"` or `["*"]` :  Either of these options will return all of the extended set
- a list of variables, e.g. `["AN", "GQ"]`. This will return the default set of variables with just the variables selected in the list added on 

An example of using this can be seen in this [notebook](https://github.com/malariagen/parasite-data/blob/update_docs_with_xarray/docs/pf7/Data_exploration.ipynb) under the Variant Calls section. Please let me know if you think this is a useful feature or if you have any thoughts!

roamato:
@kathryn1995 I like the three options (default, all, and default+some) and I think they would pretty much cover all common use cases (1 and 3 being probably the most frequent ones). As for the implementation, I'm not the expert but I wonder if `*` can be misleading as it has a specific meaning as an argument?

alimanfoo:
Hi all, it's probably worth bearing in mind that if you want a dataset with a specific selection of variables then you can use xarray to do that, in a similar way to selecting columns from a pandas dataframe. Here's an example using the Ag3 module:

image

We could make our life easier here by just offering two options in the API, default and extended, then allow the user to whittle down extended to a subset via xarray if they want to.

To be specific, this would mean going back to the approach of having either `extended=False` or `extended=True`.

kathryn1995:
> 

Hi @alimanfoo, 

I appreciate your point about it being simpler to just have `extended = True/False` and then the user can filter after. 

I was leaning towards having the option to specify what you want for a few reasons. For the people who are using the extended parameter, I would expect that they know what they want for the analysis they are doing. The run to get the full set can vary from about 30-50s and then there is a lot to filter through (92 variables) to find the ones you want. On the other hand, if you only want a handful added on to the default, this takes about 4-6 seconds. 

I just think having the option to set the fields you want would streamline things for the user as you don't have to wait each time to load the full dask array and then filter for what you want, which feels a bit clunky, instead, you can do it all in one go. 

The impression I got was that people would mainly be using the default dataset, so never really deal with this parameter anyway. 

@podpearson @roamato @ahernank Any thoughts on this?

podpearson:
@kathryn1995 , @alimanfoo and @roamato , I think for me I would normally either bring in the default or everything and then pick out what I actually want to use, so `extended=True/False` would probably be fine for me personally. I completely get Alistair's argument but the 30-50s to create the xarray is enough to be annoying. I'm used to doing things like:
```
callset = zarr.open(zarr_fn)
list(callset['variants'])
```
to see all of the variant fields, which is super-quick, so having to wait 30s would be a bit of a pain, but I guess if I wanted to see a list of all the variant fields, this would be the only way if using xarray?

I was wondering if, rather than an `extended` argument, we could have a `fields` argument with the same logic as that in `allel.read_vcf` (https://scikit-allel.readthedocs.io/en/stable/io.html), and a default of the fields brought in by the current code. But I think keeping the code as simple and as consistent as possible between releases/species is perhaps more important than my concerns, so I think I'm now favouring Alistair's suggestion. At least this doesn't rule out any use cases.

roamato:
Given the diversity of info in those fields, I can't really see any significant use case where one REALLY needs to work with all of them simultaneously (versus I can see strong use cases for importing some critical extra ones) and the significantly longer computing time is probably enough for me to support an explicit list of fields to import. That said, I agree that in order to select a subset you'll still need a way to list what's in the zarr.

Honestly I don't think this is so controversial to the point of stopping development and there are pros and cons on either side: @kathryn1995 perhaps make an executive decision?

alimanfoo:
Hi folks, agree 30+ seconds is a long time to make people wait, I'd be happy for that to be the overriding consideration. 

Btw when putting zarr data in the cloud it's well worth [consolidating the metadata](https://zarr.readthedocs.io/en/stable/tutorial.html#consolidating-metadata). This takes the array and group metadata from all the .zarray and .zgroup files and packs them into a single .zmetadata file. The benefit is that all the metadata for a zarr hierarchy can be retrieved in a single read request. The characteristic of object storage is high latency (~100ms) per read request, so the current 30+s time is probably due at least in part to lots of small read operations to read metadata for each array, which would go away if using consolidated metadata.

Consolidating the metadata is pretty straightforward and can be done in-place on GCS. Something like this would do the trick:

```python
import gcsfs

# UNCOMMENT THIS TO AUTHENTICATE. YOU ONLY NEED TO RUN THIS ONCE.
# After running this once, your authentication token should then be cached in `~/.gcs_tokens`
# Once you have authenticated, you should comment this out again to avoid re-authenticating.
# Overriding the client_config avoids the blocked app error
# gcsfs.core.client_config['installed']['client_id'] = '160816776314-7h1p2h6bqhe77vs1su4rr22fblso9dre.apps.googleusercontent.com'
# gcsfs.core.client_config['installed']['client_secret'] = 'SrJUwD58yKgViyJtw-NP2sz8'
# gcs_browser = gcsfs.GCSFileSystem(project='malariagen-jupyterhub', token='browser')

# setup gcs using previously cached authentication token
gcs = gcsfs.GCSFileSystem(project='malariagen-jupyterhub', token='cache', cache_timeout=0)

# setup access to a zarr store
store = gcs.get_mapper("gs://pf7_staging/pf7.zarr")

# consolidate metadata
import zarr
zarr.consolidate_metadata(store=store)
```

After you've done this, you should see a new object appear at `gs://pf7_staging/pf7.zarr/.zmetadata`. Now you can open the zarr using the consolidated metadata:

```python
root = zarr.open_consolidated(store=store)
```

E.g., [here's example from amin1](https://github.com/malariagen/malariagen-data-python/blob/ce4c7a88cd1cea88a7ee8f8681ef422e19003561/malariagen_data/amin1.py#L132).

I'd suggest consolidating metadata, then seeing how fast you get an xarray dataset with the full variable list.

kathryn1995:
Hi all,

Thanks for all the feedback. After consolidating the metadata this has drastically reduced the time to collect everything into a dask array, so thanks a lot for that suggestion @alimanfoo! 

To get the default dataset now takes a couple hundred milliseconds and the full extended set takes about a second. With this in mind, I've decided to scrap the option to pass a list of the variables you want to return, as the main purpose of this was to reduce the time users were waiting for their data. 

It seems to get a subset of the extended dataset would have been a two-step process for most people anyway: finding the list of available variables in the Zarr, and then picking the ones they want to call `variant_calls()`. With the reduced time to get the extended set, I don't think this is much different to calling `variant_calls` first and then filtering the dask array you get back. 

If we see a need for it we can always add this functionality back in easily, but I think it's best to keep things simple at this stage of the MVP

alimanfoo:
Hi @kathryn1995, really glad the consolidation helped, code is looking nice and lean.

Re the tests, I think there's still an opportunity to travel lighter. I'm going to drop some line comments into a code review, these are going to sound a bit harsh but basically what I'm trying to get at is whether a test is testing any behaviour, or whether it's just duplicating code or pinning implementation details. I think it's only worth testing behaviour, anything else is excess baggage ... but very happy if you want to push back on anything :smile:

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+        ) as mock_load_config:
+            pf7 = Pf7(self.url_starts_with_gs)
+        self.assertEqual(pf7._cache_sample_metadata, None)
+        self.assertEqual(pf7._cache_zarr, None)
+        self.assertEqual(
+            pf7.extended_calldata_variables, self.test_extended_calldata_variables
+        )
+        self.assertEqual(pf7.extended_variant_fields, self.test_extended_variant_fields)
+        mock_load_config.assert_called_once_with(None)
+        mock_init_filesystem.assert_called_once_with(self.url_starts_with_gs)
```
alimanfoo, line 187: Excess baggage?

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+        ) as mock_load_config:
+            pf7 = Pf7(self.url_starts_with_gs, data_config=self.test_config_path)
+        self.assertEqual(pf7._cache_sample_metadata, None)
+        self.assertEqual(pf7._cache_zarr, None)
+        self.assertEqual(
+            pf7.extended_calldata_variables, self.test_extended_calldata_variables
+        )
+        self.assertEqual(pf7.extended_variant_fields, self.test_extended_variant_fields)
+        mock_load_config.assert_called_once_with(self.test_config_path)
+        mock_init_filesystem.assert_called_once_with(self.url_starts_with_gs)
```
alimanfoo, line 203: Excess baggage?

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def test_load_config(self):
+        config = self.test_pf7_class._load_config(self.test_config_path)
+        self.assertEqual(
+            config,
+            {
+                "metadata_path": "metadata/test_metadata.txt",
+                "zarr_path": "pf7.zarr/",
+            },
+        )
```
alimanfoo, line 213: Excess baggage?

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def test_sample_metadata_returns_extpected_df(self):
+        metadata_df = self.test_pf7_class.sample_metadata()
+        pd.testing.assert_frame_equal(metadata_df, self.test_metadata_df)
```
alimanfoo, line 217: This tests a public API method, and verifies loading of some test data from a file behaves correctly :+1:

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def test_read_general_metadata_returns_expected_df_with_cache_set(
+        self,
+    ):
+        self.test_pf7_class.sample_metadata()
+        metadata_df = self.test_pf7_class.sample_metadata()
+        pd.testing.assert_frame_equal(metadata_df, self.test_metadata_df)
```
alimanfoo, line 224: This tests a public API method and verifies correct behaviour after caching :+1:

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+
+    @patch("builtins.open", new_callable=mock_open)
+    @patch("malariagen_data.pf7.pd.read_csv", return_value=pd.DataFrame())
+    def test_read_general_metadata_uses_cache_when_set(self, mock_read_csv, mock_open):
+        self.test_pf7_class.sample_metadata()
+        self.test_pf7_class.sample_metadata()
+        mock_open.assert_called_once()
+        mock_read_csv.assert_called_once()
```
alimanfoo, line 232: This tests a public API method and verifies that caching is working :+1:

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+    @patch("malariagen_data.pf7.zarr.open_consolidated")
+    def test_open_zarr_calls_functions_correctly(self, mock_zarr, mock_safestore):
+        with patch(
+            "malariagen_data.pf7.init_filesystem",
+            return_value=["fs", self.test_data_path],
+        ):
+            pf7_mock_fs = Pf7(self.test_data_path, data_config=self.test_config_path)
+        pf7_mock_fs.open_zarr()
+        mock_safestore.assert_called_once_with(fs="fs", path=self.test_zarr_path)
+        mock_zarr.assert_called_once_with(store="Safe store object")
```
alimanfoo, line 244: Excess baggage?

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+    def test_open_zarr_uses_cache(self, mock_zarr, mock_safestore):
+        with patch(
+            "malariagen_data.pf7.init_filesystem",
+            return_value=["fs", self.test_data_path],
+        ):
+            pf7_mock_fs = Pf7(self.test_data_path, data_config=self.test_config_path)
+        pf7_mock_fs.open_zarr()
+        pf7_mock_fs.open_zarr()
+        mock_safestore.assert_called_once_with(fs="fs", path=self.test_zarr_path)
+        mock_zarr.assert_called_once_with(store="Safe store object")
```
alimanfoo, line 257: This verifies that caching is working :+1:

malariagen_data/pf7.py
```diff
@@ -0,0 +1,291 @@
...
+        return self._cache_sample_metadata
+
+    def open_zarr(self):
+        if self._cache_zarr is None:
+            path = os.path.join(self._path, self.CONF["zarr_path"])
+            store = init_zarr_store(fs=self._fs, path=path)
+            self._cache_zarr = zarr.open_consolidated(store=store)
+        return self._cache_zarr
+
+    def _subset_extended_dictionary(self, extended_variables):
```
alimanfoo, line 157: This function isn't needed any more.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,291 @@
...
+                subset_extended_variants[var_name],
+                field,
+            )
+        return data_vars
+
+    def variant_calls(self, extended=False, inline_array=True, chunks="native"):
+        """Access variant sites, site filters and genotype calls.
+        Parameters
+        ----------
+        inline_array : bool, optional
```
alimanfoo, line 258: Missing documentation of `extended` parameter.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,291 @@
...
+        -------
+        df : pandas.DataFrame
+        """
+        if self._cache_sample_metadata is None:
+            path = os.path.join(self._path, self.CONF["metadata_path"])
+            with self._fs.open(path) as f:
+                self._cache_sample_metadata = pd.read_csv(f, sep="\t", na_values="")
+        return self._cache_sample_metadata
+
+    def open_zarr(self):
```
alimanfoo, line 150: Suggest to rename to `open_variant_calls_zarr()` for clarity.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,291 @@
...
+        **kwargs,
+    ):
+
+        # setup filesystem
+        self._fs, self._path = init_filesystem(url, **kwargs)
+        self.CONF = self._load_config(data_config)
+
+        # setup caches
+        self._cache_sample_metadata = None
+        self._cache_zarr = None
```
alimanfoo, line 38: Suggest to rename to `_cache_variant_calls_zarr` for clarity.

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+            "ANN_Feature_Type": [DIM_VARIANT, DIM_ALT_ALLELE],
+        }
+        self.assertEqual(actual_variants_dictionary, expected_variants_dictionary)
+        self.assertEqual(actual_calldata_dictionary, expected_calldata_dictionary)
+
+    def test_subset_extended_dictionary_raises_error(self):
+        with self.assertRaises(ValueError):
+            self.test_pf7_class._subset_extended_dictionary(
+                ["DP", "AC", "BAD_VARIABLE"]
+            )
```
alimanfoo, line 283: Tests a function that isn't needed any more.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,291 @@
...
+            z = root[f"variants/{var_name}"]
+            var = da_from_zarr(z, inline_array=inline_array, chunks=chunks)
+            coords[f"variant_{var_names_for_outputs[var_name]}"] = [DIM_VARIANT], var
+
+        z = root["samples"]
+        sample_id = da_from_zarr(z, inline_array=inline_array, chunks=chunks)
+        coords["sample_id"] = [DIM_SAMPLE], sample_id
+        return coords
+
+    def add_data_vars(self, root, inline_array, chunks, var_names_for_outputs):
```
alimanfoo, line 200: Should this be a private method instead? I.e., name start with "_".

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+        )
+
+        values = [actual_coordinates[k] for k in actual_coordinates]
+        for value in values:
+            assert isinstance(value[1], da.Array)
+        dimensions = [i[0] for i in values]
+        self.assertEqual(
+            dimensions,
+            [["variants"], ["variants"], ["samples"]],
+        )
```
alimanfoo, line 306: This tests behaviour of `_add_coordinates()` private method which is fair enough, although personally I would stick to testing behaviour of public methods only, because these will give you the same coverage of behaviour, and you give yourself freedom to refactor internally later without having to rewrite tests.

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+            [
+                ["variants", "alleles"],
+                ["variants"],
+                ["variants"],
+                ["variants"],
+                ["variants"],
+                ["variants", "samples", "ploidy"],
+                ["variants", "samples", "alleles"],
+            ],
+        )
```
alimanfoo, line 344: This tests (what should be) a private method, see comment above re testing private methods.

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+        for value in values:
+            assert isinstance(value[1], da.Array)
+        dimensions = [i[0] for i in values]
+        self.assertEqual(
+            dimensions,
+            [
+                ["variants", "samples"],
+                ["variants"],
+            ],
+        )
```
alimanfoo, line 373: This tests a private method, see comment above re testing private methods.

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+            [
+                "variant_allele",
+                "variant_filter_pass",
+                "variant_is_snp",
+                "variant_numalt",
+                "variant_CDS",
+                "call_genotype",
+                "call_AD",
+            ],
+        )
```
alimanfoo, line 394: This tests the `variant_calls()` public method with default params :+1: 

Although you could do this without any mocking, similar to how `test_sample_metadata_returns_extpected_df` works without any mocking.

tests/test_pf7.py
```diff
@@ -0,0 +1,433 @@
...
+                "variant_filter_pass",
+                "variant_is_snp",
+                "variant_numalt",
+                "variant_CDS",
+                "call_genotype",
+                "call_AD",
+                "call_GQ",
+                "variant_AN",
+            ],
+        )
```
alimanfoo, line 429: This tests the `variant_calls()` public method with `extended=True` :+1: 

Although again you could do this without any mocking, similar to how `test_sample_metadata_returns_extpected_df` works without any mocking.

alimanfoo (commented):
Hi @kathryn1995, sorry this is a load of comments all at once, entirely meant to be helpful :smile:

Mostly the comments here are coming from a point of view where you (1) focus on testing the public API only, and (2) test behaviour (i.e., are the outputs correct for given inputs). The point of this approach is that you get the same coverage, but you travel much lighter in terms of lines of test code, the purpose of each test is easy to follow, and you avoid pinning down any implementation details, which gives yourself the freedom to refactor internally as much as you want.

Very happy to discuss and feel free to push back.

malariagen_data/pf7.py
kathryn1995: I have removed this function and the tests that go with it. We can always get it back from the history if we decide to use it at some point

malariagen_data/pf7.py
kathryn1995: changed this for the cache and also the name of the path in the config

malariagen_data/pf7.py
kathryn1995: Added this in and just improved documentation in general

tests/test_pf7.py
kathryn1995: See comment above about testing this private method

tests/test_pf7.py
kathryn1995: See comment above about testing this private method

tests/test_pf7.py
kathryn1995: I have removed this test

tests/test_pf7.py
kathryn1995: I have kept this test but removed a lot of the unnecessary asserts and mocking. I am just using this to check the config is loaded correctly now

tests/test_pf7.py
kathryn1995: I have kept the tests for load config as I want to make sure using the default or entering your own config path works correctly

tests/test_pf7.py
kathryn1995: I have removed this as the same thing is basically tested at the same time as testing the cache

malariagen_data/pf7.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+class Pf7:
+    """Provides access to data from the Pf7 release.
+
+    Parameters
+    ----------
+        url (str, optional): Base path to data. Default uses Google Cloud Storage "gs://pf7_release/",
+                             or specify a local path on your file system if data have been downloaded.
+        data_config (str, optional) : path to config for structure of Pf7 data resource. Defaults to config in repository.
```
alimanfoo, line 31: Please edit to match [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) style.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,314 @@
...
+        return config_content
+
+    def sample_metadata(self):
+        """Access sample metadata and return as pandas dataframe
+
+        Returns
+        -------
+            df (pandas.DataFrame) : A dataframe of sample metadata on the samples that were sequenced as part of this resource.
+                                    Includes the time and place of collection, quality metrics, and accesion numbers.
+                                    One row per sample.
```
alimanfoo, line 171: Please edit to match [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) style.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,314 @@
...
+            with self._fs.open(path) as f:
+                self._cache_sample_metadata = pd.read_csv(f, sep="\t", na_values="")
+        return self._cache_sample_metadata
+
+    def open_variant_calls_zarr(self):
+        """Open variant calls zarr
+
+        Returns
+        -------
+            root (zarr.hierarchy.Group): Root of zarr containing information on variant calls.
```
alimanfoo, line 189: Please edit to [numpydoc style](https://numpydoc.readthedocs.io/en/latest/format.html#returns).

malariagen_data/pf7.py
```diff
@@ -0,0 +1,314 @@
...
+        ----------
+            extended (bool, optional): If False only the default variables are returned. If True all variables from the zarr are returned.
+                                        Defaults to False.
+            inline_array (bool, optional): Passed through to dask.array.from_array(). Defaults to True.
+            chunks (str, optional): If 'auto' let dask decide chunk size. If 'native' use native zarr chunks.
+                                    Also can be a target size, e.g., '200 MiB'. Defaults to "native".
+
+        Returns
+        -------
+            ds (xarray.Dataset): Dataset containing either defualt or extended variables from the variant calls Zarr.
```
alimanfoo, line 280: Please edit to numpydoc style.

malariagen_data/pf7.py
```diff
@@ -0,0 +1,314 @@
...
+        """Load the config for data structure on the cloud into json format."""
+        if not data_config:
+            working_dir = os.path.dirname(os.path.abspath(__file__))
+            data_config = os.path.join(working_dir, "pf7_config.json")
+        with open(data_config) as pf7_json_conf:
+            config_content = json.load(pf7_json_conf)
+        return config_content
+
+    def sample_metadata(self):
+        """Access sample metadata and return as pandas dataframe
```
alimanfoo, line 165:
```suggestion
        """Access sample metadata and return as pandas dataframe.
```

malariagen_data/pf7.py
```diff
@@ -0,0 +1,314 @@
...
+            >>> pf7.sample_metadata()
+        """
+        if self._cache_sample_metadata is None:
+            path = os.path.join(self._path, self.CONF["metadata_path"])
+            with self._fs.open(path) as f:
+                self._cache_sample_metadata = pd.read_csv(f, sep="\t", na_values="")
+        return self._cache_sample_metadata
+
+    def open_variant_calls_zarr(self):
+        """Open variant calls zarr
```
alimanfoo, line 185:
```suggestion
        """Open variant calls zarr.
```

alimanfoo (changes requested):
Some documentation nits.

malariagen_data/pf7.py
kathryn1995: Should now all be in numpydoc style

malariagen_data/pf7.py
kathryn1995: Should now all be in numpydoc style

malariagen_data/pf7.py
```diff
@@ -0,0 +1,325 @@
...
+
+class Pf7:
+    """Provides access to data from the Pf7 release.
+    Parameters
+    ----------
+    url : str, optional
+        Base path to data. Default uses Google Cloud Storage "gs://pf7_release/",
+        or specify a local path on your file system if data have been downloaded.
+    data_config : str, optional
+        path to config for structure of Pf7 data resource. Defaults to config in
```
alimanfoo, line 32: Looks like sentence is truncated?
```suggestion
    data_config : str, optional
        Path to config for structure of Pf7 data resource. Defaults to config included 
        with the malariagen_data package.
```

alimanfoo (commented):
Looks good, just a couple of tiny nits.

malariagen_data/pf7.py
kathryn1995: Should now all be in numpydoc style

kathryn1995:
> Looks good @kathryn1995, thanks for sticking with this! Good to merge?

Amazing, thanks for reviewing this! I will merge now

PR #77, opened 2021-11-10
Fix bug in applying site_mask parameter in Amin1

alimanfoo:


alimanfoo:
cc @cclarkson, found and fixed a bug in handling of the site_mask parameter. Will merge and cut a micro release if CI passes.

PR #87, opened 2021-11-23
Fix chunks argument to Ag3.snp_genotypes() getting passed through

### LINKED ISSUE #86: Chunks parameter appears to be ignored
Running `Ag3.snp_genotypes()` changing the `chunks` parameter seems to make no difference.

alimanfoo:
Resolves #86

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -546,7 +546,9 @@ class Ag3:
                 [
                     da.concatenate(
                         [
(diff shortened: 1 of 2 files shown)


alimanfoo:
Tests are failing here due to problems with CNV discordant read data.

cclarkson (approved):
👍

alimanfoo:
I'm going to close this PR and include this additional test over in #88 instead. Just because #88 naturally resolves this and merging would be complex.

PR #88, opened 2021-11-23
Support multiple releases via the sample_sets parameter

### LINKED ISSUE #85: Ag3.snp_calls() cannot take multiple releases for sample_set parameter

### LINKED ISSUE #86: Chunks parameter appears to be ignored

alimanfoo:
This PR makes some changes to how the `sample_sets` parameter is handled. In particular:

* Make sure that multiple releases can be requested via all functions that support a `sample_sets` parameter.
* Change the default behaviour of these functions so that data from all samples sets from all available releases are returned by default.

Note that this second point makes a subtle change to how the API behaves. When interacting with public releases, this will currently be equivalent to requesting `sample_sets="v3"`. When accessing pre-releases, this will access everything from v3 to v3.5. The rationale here is that, generally, this change should mean that users can forget about the `sample_sets` parameter for most use cases, which will hopefully make things a little simpler.

Resolves #85. Resolves #86.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -116,11 +116,11 @@ class Ag3:
         """
 
         if release is None:
-            # all available releases
+            # retrieve sample sets from all available releases
             release = self.releases
 
         if isinstance(release, str):
-            # single release
+            # retrieve sample sets for a single release
 
             if release not in self.releases:
                 raise ValueError(f"Release not available: {release!r}")
@@ -137,7 +137,7 @@ class Ag3:
                 return df
 
         elif isinstance(release, (list, tuple)):
-            # multiple releases
+            # retrieve sample sets from multiple releases
             df = pandas.concat(
                 [self.sample_sets(release=r) for r in release],
                 axis=0,
@@ -231,13 +231,14 @@ class Ag3:
             return df
 
     def _prep_sample_sets_arg(self, *, sample_sets):
-        if sample_sets == "v3_wild":
+
+        if sample_sets is None:
+            # all available sample sets
+            sample_sets = self.sample_sets()["sample_set"].tolist()
+
+        elif sample_sets == "v3_wild":
             # convenience, special case to exclude crosses
             sample_sets = self.v3_wild
-
-        elif sample_sets is None:
-            # all sample sets
-            sample_sets = self.sample_sets()["sample_set"].tolist()
 
         elif isinstance(sample_sets, str) and sample_sets.startswith("v3"):
             # convenience, can use a release identifier to denote all sample sets
@@ -249,12 +250,12 @@ class Ag3:
 
         return sample_sets
 
-    def species_calls(self, sample_sets="v3_wild", analysis="20200422", method="aim"):
+    def species_calls(self, sample_sets=None, analysis="20200422", method="aim"):
         """Access species calls for one or more sample sets.
 
         Parameters
         ----------
-        sample_sets : str or list of str
+        sample_sets : str or list of str, optional
             Can be a sample set identifier (e.g., "AG1000G-AO") or a list of sample set
             identifiers (e.g., ["AG1000G-BF-A", "AG1000G-BF-B"] or a release identifier (e.g.,
             "v3") or a list of release identifiers.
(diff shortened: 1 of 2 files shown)


alimanfoo:
N.B., currently some tests are skipped due to problems in the discordant read calls data. The skip marks should be removed when the data are fixed, prior to merging this PR.

malariagen_data/ag3.py
```diff
@@ -1143,14 +1169,13 @@ def snp_calls(
                 [
                     xarray.concat(
                         [
-                            self._snp_calls_dataset(
+                            self.snp_calls(
```
cclarkson, line 1172: Hi @alimanfoo, I'm not completely clear why we no longer need to call `_snp_calls_dataset` here?

cclarkson (approved):
Other than my one comment, this all makes sense and looks great to me.

malariagen_data/ag3.py
alimanfoo: Calling recursively like this allows multiple releases to be provided as the `sample_sets` argument. E.g., `sample_sets = ['v3', 'v3.1', 'v3.2']`.

alimanfoo: An alternative approach could be to properly flatten out the `sample_sets` parameter within the `_prep_sample_sets_arg` function, i.e., turn something like `['v3', 'v3.1', 'v3.2']` into a list of sample sets. I might look into that.

alimanfoo:
Hi @cclarkson, thanks looking at this. I took another look and decided it would be better to handle all the polymorphism of the `sample_sets` parameter within the `_prep_sample_sets_arg()` method. 

So now, it doesn't matter whether the user provides a single sample set, or a list of sample sets, or a single release, or a list of releases, or None (implying all data please), `_prep_sample_sets_arg()` always returns a list of sample sets.

This then simplifies logic elsewhere. In particular, methods which are performing concatenation over sample sets are simpler, and no longer need to use recursion.

Hope that makes sense.

malariagen_data/ag3.py
```diff
@@ -100,34 +100,54 @@ def releases(self):
...
+                df["release"] = release
+                self._cache_sample_sets[release] = df
+                return df
+
+        elif isinstance(release, (list, tuple)):
+            # retrieve sample sets from multiple releases
+            df = pandas.concat(
+                [self.sample_sets(release=r) for r in release],
+                axis=0,
+                ignore_index=True,
```
cclarkson, line 144: can we cache here?

malariagen_data/ag3.py
```diff
@@ -484,10 +528,17 @@ def open_snp_genotypes(self, sample_set):
             self._cache_snp_genotypes[sample_set] = root
             return root
 
+    def _snp_genotypes(self, *, contig, sample_set, field, inline_array, chunks):
+        # single single contig, single sample set
```
cclarkson, line 532: nitpick "single single"

malariagen_data/ag3.py
```diff
@@ -484,10 +528,17 @@ def open_snp_genotypes(self, sample_set):
             self._cache_snp_genotypes[sample_set] = root
             return root
 
+    def _snp_genotypes(self, *, contig, sample_set, field, inline_array, chunks):
```
cclarkson, line 531: What does the `*` do here?

malariagen_data/ag3.py
alimanfoo: it's really single :) will fix

malariagen_data/ag3.py
alimanfoo: It disallows any positional arguments, i.e., forces the method to be called with keyword only arguments. This is defensive, a common source of bugs is where you rely on positional arguments but then later add more positional args or change the argument order, and forget to update calling code somewhere. Forcing all calls to use keyword only arguments means if you add more arguments then all previous code will still work. See also https://www.python.org/dev/peps/pep-3102/.

malariagen_data/ag3.py
alimanfoo: I don't think it's worth it, as the dataframes for each of the releases in the list will be cached.

cclarkson (approved):
LGTM

alimanfoo:
Thanks Chris :)

PR #97, opened 2021-11-25
Improve species metadata

### LINKED ISSUE #55: Consolidate arguments for specifying the species calling analysis
https://github.com/malariagen/malariagen-data-python/blob/149398a9697abba5e1888b1ec106f84097000a98/malariagen_data/ag3.py#L241

It's awkward to have two arguments here, both `method` and `analysis`. 

Suggest to combine into one, e.g., `analysis="aim_20200422"`.

### LINKED ISSUE #70: Move default values for analysis parameters into constants
Several "analysis" parameters have a default value, which may be repeated several times if the parameter appears in several function signatures. It would be better to have these default values declared once via a constant at the top of the module, to make maintenance easier.

### LINKED ISSUE #93: Modify naming of species metadata columns in Ag3
This issue proposes some modifications to how the species calls are handled and presented through sample metadata.

The main problem coming up is that there are three possible sources of species/taxon assignment, available through different means, and this can be confusing:

* AIM species calls, included by default when calling `Ag3.sample_metadata()`, providing the `species` column (and several other columns)
* PCA species calls, which can be included when calling `Ag3.sample_metadata()` if the `species_calls` parameter is given as `("20200422", "pca")`, providing the `species` column (and several other columns).
* Cohorts metadata, accessible via the `Ag3.sample_cohorts()` method, which provides the `taxon` column.

This is all potentially confusing for the user. Below are some proposed changes to improve this.

## AIM species columns

Propose that we change the naming of the AIM species columns, from:

```
    aim_cols = (
        "aim_fraction_colu",
        "aim_fraction_arab",
        "species_gambcolu_arabiensis",
        "species_gambiae_coluzzii",
        "species",
    )
```

...to:

```
    aim_cols = (
        "aim_fraction_colu",
        "aim_fraction_arab",
        "aim_species_gambcolu_arabiensis",
        "aim_species_gambiae_coluzzii",
        "aim_species",
    )
```

This would make it easier to always talk about the "AIM species" assignment, and have that always visible in dataframes and pivot tables. I.e., it's always clear where the species assignment as come from.

## PCA species columns

Propose that we change the naming of the AIM species columns, from:

```
    pca_cols = (
        "PC1",
        "PC2",
        "species_gambcolu_arabiensis",
        "species_gambiae_coluzzii",
        "species",
    )
```

...to:

```
    pca_cols = (
        "pca_species_pc1",
        "pca_species_pc2",
        "pca_species_gambcolu_arabiensis",
        "pca_species_gambiae_coluzzii",
        "pca_species",
    )
```

In general we don't recommend use of the PCA species assignment, but in case anyone ever does use them, these column names make it clear what is being used.

## Notes

Still open for discussion is whether the cohort metadata should get included as well when calling `Ag3.sample_metadata()`, which would then provide the `taxon` column, which is the best column to use as it provides the most refined view of taxa within the dataset. However, I'll raise a separate issue to discuss that.

After this change is implemented, there will be some downstream consequences, as the vector data user guide will likely need to be updated (at least rerun), and possibly also partner user guides.

alimanfoo:
* Resolves #55
* Resolves #70
* Resolves #93

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -33,6 +33,7 @@ GENOME_ZARR_PATH = (
 DEFAULT_SPECIES_ANALYSIS = "aim_20200422"
 DEFAULT_SITE_FILTERS_ANALYSIS = "dt_20200416"
 DEFAULT_COHORTS_ANALYSIS = "20211101"
+CONTIGS = "2R", "2L", "3R", "3L", "X"
 
 
 class Ag3:
@@ -59,8 +60,8 @@ class Ag3:
 
     """
 
-    PUBLIC_RELEASES = ("v3",)
-    contigs = ("2R", "2L", "3R", "3L", "X")
+    public_releases = PUBLIC_RELEASES
+    contigs = CONTIGS
 
     def __init__(self, url, **kwargs):
 
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -183,17 +183,19 @@ class Ag3:
             self._cache_general_metadata[sample_set] = df
             return df
 
-    def _read_species_calls(self, *, sample_set, analysis, method):
+    def _read_species_calls(self, *, sample_set, analysis):
         """Read species calls for a single sample set."""
-        key = (sample_set, analysis, method)
+        key = (sample_set, analysis)
         try:
             return self._cache_species_calls[key]
         except KeyError:
             release = self._lookup_release(sample_set=sample_set)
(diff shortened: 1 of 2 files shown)


alimanfoo:
Hi @cclarkson, first block of changes coming in here to handling of species calls. Any comments/questions very welcome :)

cclarkson (approved):
LGTM

alimanfoo:
Thanks Chris :pray:

PR #98, opened 2021-11-25
Check release manifest

### LINKED ISSUE #74: Check for manifest.tsv when discovering releases
https://github.com/malariagen/malariagen-data-python/blob/d5df2f5de023a2033ae0690b73e8729bb079e100/malariagen_data/ag3.py#L109

If manifest.tsv is not present, but the release is assumed to exist based on the presence of the directory alone, various things will break.

alimanfoo:
* Resolves #74 

Also simplifies looking up which release a sample set is in.


cclarkson (approved):
LGTM

PR #99, opened 2021-11-25
Decode sample_id to str in Ag3.snp_calls() dataset

### LINKED ISSUE #82: Decode sample IDs when building snp_calls dataset
For consistency with sample metadata, decode sample IDs so they are strings not bytes.

alimanfoo:
* Resolves #82


alimanfoo:
Merging this one soon if no objections, small change.

cclarkson (approved):
Great.

PR #100, opened 2021-11-25
Add effects to allele frequencies

### LINKED ISSUE #95: Simplify handling of SNP effects in Ag3
Currently if a user wants SNP allele frequencies together with SNP effects, they have to call 2 methods and join/merge the resulting dataframes themselves.

It would be simpler IMO if we just added in the effects to the SNP allele frequencies. 

E.g., we could change the signature of the `Ag3.snp_allele_frequencies()` to add an `effects=True` parameter. If `True` (default) then the effects would be automatically included in the resulting dataframe.

alimanfoo:
* Resolves #95

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -768,6 +768,14 @@ class Ag3:
 
         return contig, loc_feature, df_snps
 
+    def _annotator(self):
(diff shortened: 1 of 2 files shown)


cclarkson (approved):
Nice, like it :)

Issue #14, opened 2021-03-02
Support genome regions when accessing data for Ag3


alimanfoo:
Several methods in the Ag3 class, including `snp_sites()`, `site_filters()`, `snp_genotypes()` and `snp_dataset()` take a `contig` argument and return values for a whole contig.

Proposed to deprecate the `contig` argument and replace with a more generalised `region` argument which could be any of the following:

* A contig (e.g., "3L")
* A contig region (e.g., "3L:1000000-2000000")
* A gene (e.g., "AGAP004070")
* A list/tuple of any of the above, in which case regions get concatenated

Internally this would require support for locating the indices bounding a region.

alimanfoo:
This has been implemented for most functions via #106. 

There are a few functions left, notably the CNV-related functions. These need a bit more work as locating a region requires an interval overlap query. I'll close this issue and raise a new one to cover that.

PR #106, opened 2021-12-08
Support genomic regions

nkran:
Wanted to give #14 a go as it would be a neat feature to have in every day data wrangling.
Despite WIP I wanted to open a PR to check in with you regarding implementation and suggestions.

Still to-do:
- [ ] Add support for CNV and haplotype functions
- [x] Tests for `region_slice()`


alimanfoo:
Hi @nkran, this is great stuff and very much appreciated!

At a first glance this looks like a great direction. I'll give it a more thorough review asap but expect comments will be relatively minor.

malariagen_data/ag3.py
```diff
@@ -61,6 +63,7 @@ class Ag3:
     """
 
     contigs = CONTIGS
+    Region = namedtuple("Region", ["contig", "start", "end"])
```
alimanfoo, line 66: Suggest to move this outside the class definition, so it's defined at the module level.

malariagen_data/ag3.py
```diff
@@ -314,6 +317,74 @@ def _prep_sample_sets_arg(self, *, sample_sets):
...
+                or end > self.genome_sequence(region=contig).shape[0]
+            ):
+                raise ValueError("Provided genomic coordinates are not valid.")
+
+        else:
+            raise ValueError(f"Region {region} is not valid.")
+
+        return Ag3.Region(contig, start, end)
+
+    def region_slice(self, region):
```
alimanfoo, line 367:
```suggestion
    def locate_region(self, region):
```

Nit, minor naming suggestion.

malariagen_data/ag3.py
```diff
@@ -314,6 +317,74 @@ def _prep_sample_sets_arg(self, *, sample_sets):
...
+                raise ValueError(f"Contig {contig} does not exist in the dataset.")
+            elif (
+                start < 0
+                or end <= start
+                or end > self.genome_sequence(region=contig).shape[0]
+            ):
+                raise ValueError("Provided genomic coordinates are not valid.")
+
+        else:
+            raise ValueError(f"Region {region} is not valid.")
```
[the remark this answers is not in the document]
nkran: Great, thanks. Will commit the changes.

nkran:
Hi @alimanfoo, thanks for the first review :) I added your corrections and support for regions in the other functions. 

I saw that there were TODO tags for multiple contigs/regions support in `cnv_discordant_read_calls()`, `cnv_hmm()` and `cnv_coverage_calls()` so I followed suit from other methods to do the same there. Hope I didn't mess up.

A few combinations of tests for CNV methods with some regions are failing with `KeyError` from `allel.locate_region()`. I assume that is due to the empty dataset that gets returned if there are no CNVs in a given region. Any suggestions or preference about how to resolve that?

Have a nice weekend,
N

malariagen_data/ag3.py
```diff
@@ -1817,17 +2035,27 @@ def gene_cnv_frequencies(
...
 
         # get sample metadata
         df_meta = self.sample_metadata(sample_sets=sample_sets)
 
-        # get genes
-        df_genes = self.geneset().query(f"type == 'gene' and contig == '{contig}'")
+        # find genes in the region
+        if region.start and region.end:
+            df_genes = self.geneset().query(
+                f"type == 'gene' and contig == '{region.contig}' and start >= {region.start} and end <= {region.end}"
```
alimanfoo, line 2050:
```suggestion
                f"type == 'gene' and contig == '{region.contig}' and start <= {region.end} and end >= {region.start}"
```

Possible tweak to retrieve all genes overlapping the requested region.

tests/test_ag3.py
```diff
@@ -1121,8 +1144,16 @@ def test_gene_cnv(contig, sample_sets):
     df_samples = ag3.sample_metadata(sample_sets=sample_sets, species_calls=None)
     n_samples = len(df_samples)
     assert ds.dims["samples"] == n_samples
+
     df_geneset = ag3.geneset()
-    df_genes = df_geneset.query(f"type == 'gene' and contig == '{contig}'")
+    region = ag3._resolve_region(region)
+    if region.start and region.end:
+        df_genes = df_geneset.query(
+            f"type == 'gene' and contig == '{region.contig}' and start >= {region.start} and end <= {region.end}"
```
alimanfoo, line 1152: Will need to edit this to match above if changing to overlap query.

alimanfoo (commented):
Couple of small suggestions.

alimanfoo:
> A few combinations of tests for CNV methods with some regions are failing with `KeyError` from `allel.locate_region()`. I assume that is due to the empty dataset that gets returned if there are no CNVs in a given region. Any suggestions or preference about how to resolve that?

Yes that would be my guess, especially the discordant read calls, there are only a couple of regions with data there. Could just adapt the regions tested so they return some data and avoid KeyError, e.g., for the discordant read calls could use genes from the cyp6p/aa cluster on 2R?

alimanfoo:
Hi @nkran, on second thoughts, it might be better to back out the changes to the CNV methods and revisit those in a later PR. The reason is that CNVs span intervals, and so we'd need to do an overlap query for regions. I.e., the current approach of using the `POS` field (which holds the start coordinates) and `SortedIndex.locate_range()` will only find CNVs where the start occurs within the requested region.

nkran:
Hi @alimanfoo! Ah yes, that makes sense! I'll omit those two commits and update the branch.

alimanfoo:
Hi @nkran, thanks so much, this all looks good to me now, except for one .pyc file has crept in (`tests/__pycache__/__init__.cpython-37.pyc`). If you could remove that then it looks good to merge.

nkran:
Hi @alimanfoo, just to double check - my commit (a2362bd) removes that [.pyc file](https://github.com/malariagen/malariagen-data-python/blob/master/tests/__pycache__/__init__.cpython-37.pyc) that is currently in master. Do you want me to revert the commit that deletes it and keep the .pyc file or to remove the .pyc file? Thanks for your patience :) Just want to get it right.

alimanfoo:
> Hi @alimanfoo, just to double check - my commit ([a2362bd](https://github.com/malariagen/malariagen-data-python/commit/a2362bdf9807e306168ce2d76b18246ebe513d7d)) removes that [.pyc file](https://github.com/malariagen/malariagen-data-python/blob/master/tests/__pycache__/__init__.cpython-37.pyc) that is currently in master. Do you want me to revert the commit that deletes it and keep the .pyc file or to remove the .pyc file? Thanks for your patience :) Just want to get it right.

My bad, I wasn't looking carefully enough to realise the file was already in master and you were cleaning up :) 

OK to merge?

alimanfoo (approved):

nkran:
OK! :) 🚀

alimanfoo:
Done, thanks again :smile: :pray:

PR #109, opened 2021-12-20
Add cohorts to sample_metadata()

### LINKED ISSUE #94: Include cohorts in Ag3 sample metadata
The cohort metadata are very useful and it would make life easier if they were included with the sample metadata by default, rather than having to separately load and merge the sample metadata and cohort metadata dataframes. To implement this, the following changes could be made:

1. Modify the signature of the `Ag3.sample_metadata()` method to include a `cohorts_analysis` parameter with a default value to the latest cohorts analysis. 
2. Modify the implementation of the `Ag3.sample_metadata()` method to load the cohorts metadata and join it to the general sample metadata, if the `cohorts_analysis` parameter is not None.

Note that a potential problem with doing this is that some future releases may be released before a new cohorts analysis is performed including it. In this case the release would have sample metadata but no cohorts metadata. This could be mitigated by either (a) ensuring we always run a cohorts analysis immediately following each release and before partners are notified, and/or (b) building in some logic to handle missing cohort files.

cclarkson:
resolves #94 

Adds `cohorts_analysis` parameter to `sample_metadata()`, defaults to latest cohorts analysis. This adds the cohort metadata columns to the `sample_metadata()` output.

PR also adds tests.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -418,16 +418,26 @@ class Ag3:
 
         return df
 
-    def _sample_metadata(self, *, sample_set, species_calls):
+    def _sample_metadata(self, *, sample_set, species_calls, cohorts_analysis):
         df = self._read_general_metadata(sample_set=sample_set)
         if species_calls is not None:
             df_species = self._read_species_calls(
                 sample_set=sample_set, analysis=species_calls
             )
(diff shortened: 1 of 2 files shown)


alimanfoo (approved):

alimanfoo:
Nice one :snake:

PR #112, opened 2021-12-23
Misc minor improvements

### LINKED ISSUE #103: Default to GCS URL for Ag3
Mostly we use GCS to access Ag3 data, make it default so we can do:

```
ag3 = malariagen_data.Ag3()
```

### LINKED ISSUE #104: Migrate to using "3.0" instead of "v3" etc. for releases
Try to standardise on common naming for releases, in parameters and dataframes, common to what we say in documentation etc. This would mean, e.g., using "3.0" instead of "v3", "3.1" instead of "v3.1", etc. Would require some translation behind the scenes.

### LINKED ISSUE #111: Better repr for Ag3
We could make the repr better, e.g., return some text explaining that this provides access to data from MalariaGEN Vector Observatory, Ag3 releases, see https://malariagen.github.io/vector-data/ag3/api.html for API documentation. Also something about terms of use.

alimanfoo:
* Resolves #104
* Resolves #103
* Resolves #111 
* Standardise naming of parameters providing an analysis version identifier
  * `species_calls -> species_analysis`
  * `site_filters -> site_filters_analysis`

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -96,6 +96,7 @@ class Ag3:
 
     def __init__(self, url=DEFAULT_URL, **kwargs):
 
+        self._url = url
         self._pre = kwargs.pop("pre", False)
 
         # setup filesystem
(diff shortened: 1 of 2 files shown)


cclarkson (approved):
LGTM

alimanfoo:
Thanks @cclarkson.

PR #120, opened 2022-01-11
Drop cohorts columns below min cohort size when computing frequencies

### LINKED ISSUE #118: Drop cohorts columns below min cohort size when computing frequencies
Currently in methods `Ag3.snp_allele_frequencies()` and `Ag3.gene_cnv_frequencies()`, if a cohort size falls below the `min_cohort_size` parameter then the cohort still has a column in the output dataframe, but the values are all nan. It would be better as a user to just drop these cohorts, rather than expecting the user to drop them themselves, I think.

cclarkson:
resolves #118 

Removes nan freq columns and updates tests.


alimanfoo (approved):
Thanks @cclarkson, LGTM.

PR #121, opened 2022-01-11
Fix bug with optimised dask compress

### LINKED ISSUE #114: Inconsistency in array lengths before and after computation
See example below, where results are different when run with different binder environments. These different environments may contain different versions of the malariagen_data package, but it is still surprising that results change for this particular situation.

image

image

alimanfoo:
Adds test and fix for subtle bug found within the optimised dask compress function. Resolves #114.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -698,14 +698,30 @@ class Ag3:
         if field is None:
             # return POS, REF, ALT
             ret = tuple(
-                self.snp_sites(region=region, field=f, site_mask=None)
+                self.snp_sites(
+                    region=region,
+                    field=f,
+                    site_mask=None,
+                    chunks=chunks,
+                    inline_array=inline_array,
+                )
                 for f in ("POS", "REF", "ALT")
             )
 
         elif isinstance(region, (tuple, list)) and not isinstance(region, Region):
             # concatenate
             ret = da.concatenate(
-                [self.snp_sites(region=r, field=field, site_mask=None) for r in region]
+                [
+                    self.snp_sites(
+                        region=r,
+                        field=field,
+                        site_mask=None,
+                        chunks=chunks,
+                        inline_array=inline_array,
+                    )
+                    for r in region
+                ],
+                axis=0,
             )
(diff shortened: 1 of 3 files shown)


malariagen_data/ag3.py
```diff
@@ -698,14 +698,30 @@ def snp_sites(
...
             # return POS, REF, ALT
             ret = tuple(
-                self.snp_sites(region=region, field=f, site_mask=None)
+                self.snp_sites(
+                    region=region,
+                    field=f,
+                    site_mask=None,
+                    chunks=chunks,
+                    inline_array=inline_array,
+                )
```
alimanfoo, line 707: This is incidental, not part of the bug fix.

malariagen_data/ag3.py
```diff
@@ -717,7 +733,11 @@ def snp_sites(
 
         if site_mask is not None:
             loc_sites = self.site_filters(
-                region=region, mask=site_mask, analysis=site_filters_analysis
+                region=region,
+                mask=site_mask,
+                analysis=site_filters_analysis,
+                chunks=chunks,
+                inline_array=inline_array,
```
alimanfoo, line 740: This is incidental, not part of the bug fix.

malariagen_data/util.py
```diff
@@ -204,10 +204,15 @@ def da_compress(indexer, data, axis):
...
     assert indexer.shape[0] == data.shape[axis]
+
+    # useful variables
     old_chunks = data.chunks
     axis_old_chunks = old_chunks[axis]
     axis_n_chunks = len(axis_old_chunks)
 
+    # ensure indexer and data are chunked in the same way
+    indexer = indexer.rechunk((axis_old_chunks,))
+
```
alimanfoo, line 215: I believe this is the key fix, rechunking the indexer here rather than later, although why this makes a difference is unclear.

cclarkson (approved):
Nice! That was a weird a one.

alimanfoo:
> Nice! That was a weird a one.

Yeah, weird indeed!

PR #124, opened 2022-01-14
`ag3.aa_allele_frequencies()` method

### LINKED ISSUE #101: Ag3 method for computing amino acid substitution frequencies
Currently it's possible to compute SNP allele frequencies. But often it's useful to compute frequencies of amino acid substitutions, because (1) this is often what the user is directly interested in, and (2) it simplifies the case where multiple SNPs cause the same amino acid substitution.

Placeholder issue to discuss proposals for design and implementation of such a function.

cclarkson:
resolves #101 

Adds a method which is similar to `snp_allele_frequencies()` except it collapses multi-variant aa-changes and drops unneeded columns.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -1163,8 +1163,7 @@ class Ag3:
 
         Notes
         -----
-        NaNs will be returned for any cohorts with fewer samples than min_cohort_size,
-        these can be removed from the output dataframe using pandas df.dropna(axis='columns').
+        Cohorts with fewer samples than min_cohort_size will be excluded from output.
 
         """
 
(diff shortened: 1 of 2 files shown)


cclarkson:
Hi @alimanfoo - I have revised as suggested.

alimanfoo (approved):
LGTM

malariagen_data/ag3.py
```diff
@@ -2395,6 +2395,100 @@ def sample_cohorts(
...
+
+        Returns
+        -------
+        df_snps : pandas.DataFrame
+
+        Notes
+        -----
+        NaNs will be returned for any cohorts with fewer samples than min_cohort_size,
+        these can be removed from the output dataframe using pandas df.dropna(axis='columns').
+
```
alimanfoo, line 2449: This is out of date?

PR #125, opened 2022-02-02
ag3.plot_frequencies_heatmap()

### LINKED ISSUE #102: Ag3 method for plotting a heatmap from frequencies dataframes
For reporting it would be useful to have a method for turning a dataframe of frequencies into a heatmap style representation. E.g.:

image

### LINKED ISSUE #127: Remove nan columns from Ag3.gene_cnv_frequencies()
in line with `snp_allele_frequencies`.

cclarkson:
Input - pandas dataframe output by ag3.snp_allele_frequencies() or ag3.gene_cnv_frequencies().

Output - plotly.express heatmap of frequencies

resolves #102 
resolves #127


malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+        or ag3.gene_cnv_frequencies() that has been user-filtered to just rows of interest (fewer rows than max_len).
+
+        Parameters
+        ----------
+        df : pandas DataFrame
+           Filtered output from ag3.snp_allele_frequencies() or ag3.gene_cnv_frequencies.
+        index : list of str
+            A list of one or more column headers present in the input dataframe, these become
+            the heatmap y-axis row labels. When taken together, these columns must produce a unique index.
+        index_name : str
```
alimanfoo, line 2507: It would be good to make this optional. I.e., the user doesn't have to specify an `index_name`, in which case no Y axis label appears.

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+            This is the y-axis label that will be displayed on the heatmap
+        max_len : int, optional
+            Displaying large styled dataframes may cause ipython notebooks to crash.
+        width : int
```
alimanfoo, line 2511: It would be good to make this the width per column, so the size of the dataframe scales naturally if there are more or less columns. So to get the actual width you pass to plotly, multiple the `width` parameter passed by the user by the number of frequency columns.

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+            Width of styled output.
+        colorbar : bool, optional
+            If False, colorbar is not output.
+
+        Returns
+        -------
+        Plotly.express heatmap built from the index and "frq_" columns of input DataFrame, where
+        increasing frequency is denoted by intensity of red coloration.
```
alimanfoo, line 2519: Nothing is returned from this function, so can remove the "Returns" section.

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+
+        """
+        # check len of input
+        assert len(df) <= max_len
```
alimanfoo, line 2523: Suggest either raise a `ValueError` here with a helpful message, or slice the dataframe to the first `max_len` rows.

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+        for i in index:
+            df.loc[:, i] = df[i].astype(str)
+
+        # make a new column out of index list and give it the index_name
+        df.loc[:, index_name] = df[index].agg("_".join, axis=1)
+
+        # check that index is unique (otherwise style won't work)
+        assert df[index_name].nunique() == len(
+            df
+        ), f"{index} - does not produce a unique index"
```
alimanfoo, line 2538:
```suggestion
        if not df[index_name].is_unique:
            raise ValueError(f"{index} does not produce a unique index")
```

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+            df.loc[:, i] = df[i].astype(float)
+
+        # keep only freq cols
+        heatmap_df = df[frq_cols].copy()
+
+        # set index
+        heatmap_df.set_index(index_name, inplace=True)
+
+        # clean column names
+        heatmap_df.columns = heatmap_df.columns.str.lstrip("frq_")
```
alimanfoo, line 2554: nice

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+            aspect="auto",
+            width=width,
+            color_continuous_scale="Reds",
+        )
+
+        fig.update_xaxes(side="top", tickangle=270, title="cohorts")
+        fig.update_layout(
+            coloraxis_colorbar=dict(
+                title="frequency",
+                tickvals=[0, 0.2, 0.4, 0.6, 0.8, 0.99],
```
alimanfoo, line 2571: Why doesn't this end with 1.0?

alimanfoo (commented):
Hi @cclarkson, looks good, a few suggestions...

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,92 @@ def aa_allele_frequencies(
...
+                ticktext=["1%", "20%", "40%", "60%", "80%", "100%"],
```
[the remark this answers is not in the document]
cclarkson: Ah, yes, missed that, thanks. I was playing with moving the 100% down a bit, away from the colorbar title (so moved 0% up a bit to balance) but decided it wasn't time to play with minutiae like that!

malariagen_data/ag3.py
cclarkson: Aesthetics - I was trying to imperceptibly move the final tickval away from the colorbar title.

cclarkson: Completely, unnecessary of course!

malariagen_data/ag3.py
cclarkson: If I understand correctly, the colorbar/row labels/_etc._ are included in "width", so it isn't just a multiple of columns that we want, plus colorbar and row labels are moveable feasts size-wise. I guess the work around is default to no width, but use it if given?

cclarkson:
@alimanfoo, thanks for comments, they have all been addressed now.

cclarkson:
@alimanfoo, all works, thanks - should be good to merge now.

alimanfoo:
Hi @cclarkson, code looks good. Could you add the notebook to this PR? Suggest putting it at `notebooks/test_plot_frequencies_heatmap.ipynb` or something like that.

malariagen_data/ag3.py
```diff
@@ -2489,6 +2490,89 @@ def aa_allele_frequencies(
...
+
+        """Plot a heatmap from a pandas DataFrame of frequencies, e.g., output from
+            `Ag3.snp_allele_frequencies()` or `Ag3.gene_cnv_frequencies()`. It's recommended to
+            filter the input DataFrame to just rows of interest, i.e., fewer rows than `max_len`.
+
+        Parameters
+        ----------
+        df : pandas DataFrame
+           A DataFrame of frequencies, e.g., output from `snp_allele_frequencies()` or `gene_cnv_frequencies()`.
+        index : str or list of str
```
KellyLBennett, line 2505: To make it even easier for the user, you could potentially have the option to specify if the table is either for SNP frequencies or CNVs and decide the indexes based on this information or perhaps update the doc string with details of the standard input.

KellyLBennett (commented):
A great functionality to add to the package. I only had one comment which I added to the changed document.

malariagen_data/ag3.py
alimanfoo: Hi @KellyLBennett, this is a good thought and we've had a bit of debate around this. I wonder if an even simpler approach would be to have the outputs the functions that create frequencies already set an index. E.g., the output of `snp_allele_frequencies()` is a dataframe with the appropriate index already set (contig, position, ref, alt). E.g., the output of `gene_cnv_frequencies()` is a dataframe with index gene_id already set. Then the `index` parameter here to `plot_frequencies_heatmap()` could be optional, and if not given would just use whatever index was already set. If so, the user could then do something like:

```python
df_freq = ag3.snp_allele_frequencies(...)
df_freq = df_freq.query(...)  # do some filtering
ag3.plot_frequencies_heatmap(df_freq)
```

cclarkson: Thanks @KellyLBennett, @alimanfoo, I will have a look and see how that might work.

alimanfoo:
Hi @cclarkson, the notebook looks good, thanks.

Perhaps we should address @KellyLBennett's point about usability here before merging, I think we'll probably thank ourselves.

I think this would mean:

* Modify `snp_allele_frequencies()` to set `["contig", "position", "ref_allele", "alt_allele", "aa_change"]` as index on frequencies dataframe before returning.
* Modify `aa_allele_frequencies()` to set `"aa_change"` as index before returning.
* Modify `gene_cnv_frequencies()` to set `["ID", "Name"]` as index before returning.
* Modify `plot_frequences_heatmap()` to make the `index` parameter optional, and if not provided, use the index already set on the input dataframe, converting to str and flattening if multiindex.

What do you think?

cclarkson:
Sounds good, I'll sort that after lunch.

cclarkson:
@alimanfoo, should be ready to go now. I also fixed the `gene_cnv_frequencies()` NaN column issue here.

malariagen_data/ag3.py
```diff
@@ -2489,6 +2494,98 @@ def aa_allele_frequencies(
...
+        if index is not None:
+            df = df.reset_index().copy()
+            if isinstance(index, list):
+                index_col = (
+                    df[index]
+                    .astype(str)
+                    .apply(lambda row: "_".join(row), axis="columns")
+                )
+            if isinstance(index, str):
+                index_col = df[index].astype(str)
```
alimanfoo, line 2545:
```suggestion
        if index is None:
            index = list(df.index.names)

        df = df.reset_index().copy()
        if isinstance(index, list):
            index_col = (
                df[index]
                .astype(str)
                .apply(lambda row: "_".join(row), axis="columns")
            )
        elif isinstance(index, str):
            index_col = df[index].astype(str)
        else:
            raise TypeError("wrong type for index parameter, expected list or str")
```

Possible simplification.

alimanfoo:
Hi @cclarkson, there's some tests need updating as well here for the changes to indexes.

alimanfoo:
One more thing, would suggest instead of:

```
.apply(lambda row: "_".join(row), axis="columns")
```

...to do:

```
.apply(lambda row: ", ".join([o for o in row if o is not None]), axis="columns")
```

...as this will avoid ugly "None" appearing in axis labels.

cclarkson:
Hi @alimanfoo, thanks! Will fix in the morning.

malariagen_data/ag3.py
```diff
@@ -1241,6 +1242,17 @@ def snp_allele_frequencies(
             ann = self._annotator()
             ann.get_effects(transcript=transcript, variants=df_snps)
 
+            df_snps.set_index(
+                ["contig", "position", "ref_allele", "alt_allele", "aa_change"],
+                inplace=True,
+            )
+
+        if not effects:
```
alimanfoo, line 1250:
```suggestion
        else:
```

tests/test_ag3.py
```diff
@@ -730,10 +730,10 @@ def test_snp_allele_frequencies__str_cohorts():
...
     min_cohort_size = 10
     universal_fields = [
-        "contig",
-        "position",
-        "ref_allele",
-        "alt_allele",
+        # "contig",
+        # "position",
+        # "ref_allele",
+        # "alt_allele",
```
alimanfoo, line 736: I'd delete these rather than leave commented out. 

Also I'd suggest adding a test on the index names.

tests/test_ag3.py
```diff
@@ -821,10 +814,6 @@ def test_snp_allele_frequencies__str_cohorts__effects():
     cohorts = "admin1_month"
     min_cohort_size = 10
     universal_fields = [
-        "contig",
-        "position",
-        "ref_allele",
-        "alt_allele",
```
alimanfoo, line 827: Add test on index names.

tests/test_ag3.py
```diff
@@ -1266,7 +1254,6 @@ def test_gene_cnv_frequencies(contig, cohorts):
 
         assert isinstance(df, pd.DataFrame)
         assert len(df) == len(df_genes)
-        assert df.index.name == "ID"
```
alimanfoo, line 1269: Add test on index names.

alimanfoo (commented):
Hi @cclarkson, nearly there, just a couple of very minor things...

alimanfoo:
Hi @cclarkson, are you done here?

tests/test_ag3.py
cclarkson: Hi @alimanfoo, I removed the tests as multi-column-index `df.index.name` returns `None`. I will have a play to see what I can find.

cclarkson: Ha ha, it's just `.names` !

alimanfoo:
Thanks so much @cclarkson. I've resolved conflicts with the sample_query PR, will wait to confirm CI passes then merge here.

cclarkson:
@alimanfoo - still come issues, will have a look now.

alimanfoo:
Woohoo :sunflower:

PR #126, opened 2022-02-03
Add sample_query parameter to allele frequencies

### LINKED ISSUE #96: Add sample_query parameter to Ag3 methods that return frequencies tables
To simplify building tables of frequencies of either SNPs or CNVs, where often a user might only want to consider a subset of samples, proposed to add a `sample_query` parameter to the `Ag3.snp_allele_frequencies()` and `Ag3.gene_cnv_frequencies()` methods, which applies a pandas query and then only returns cohorts which contain samples from this selection.

Also might want to reconsider what happens when cohorts are too small to calculate frequencies, e.g., drop columns rather than return nans.

alimanfoo:
Resolves #96.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -1115,6 +1115,7 @@ class Ag3:
         self,
         transcript,
         cohorts,
+        sample_query=None,
         cohorts_analysis=DEFAULT_COHORTS_ANALYSIS,
         min_cohort_size=10,
         site_mask=None,
@@ -1135,6 +1136,8 @@ class Ag3:
             {"admin1_month", "admin1_year", "admin2_month", "admin2_year"}.
             If a dict, should map cohort labels to sample queries, e.g.,
             `{"bf_2012_col": "country == 'Burkina Faso' and year == 2012 and species == 'coluzzii'"}`.
+        sample_query : str, optional
+            @@TODO
         cohorts_analysis : str
             Cohort analysis identifier (date of analysis), default is latest version.
         min_cohort_size : int
@@ -1167,6 +1170,16 @@ class Ag3:
 
         """
 
+        # handle sample_query
+        loc_samples = None
+        if sample_query is not None:
+            df_samples = self.sample_metadata(
+                sample_sets=sample_sets,
+                cohorts_analysis=cohorts_analysis,
+                species_analysis=species_analysis,
+            )
+            loc_samples = df_samples.eval(sample_query).values
+
         # setup initial dataframe of SNPs
         region, loc_feature, df_snps = self._snp_df(
             transcript=transcript, site_filters_analysis=site_filters_analysis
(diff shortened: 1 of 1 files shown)


cclarkson:
Hi @alimanfoo , this works nicely for `snp_allele_frequencies()` but for some reason I can't see yet, `gene_cnv_frequencies` only works without              `if n_samples == 0:
     raise ValueError(f"no samples for cohort {coh!r}")`.

malariagen_data/ag3.py
```diff
@@ -1135,6 +1136,9 @@ def snp_allele_frequencies(
             {"admin1_month", "admin1_year", "admin2_month", "admin2_year"}.
             If a dict, should map cohort labels to sample queries, e.g.,
             `{"bf_2012_col": "country == 'Burkina Faso' and year == 2012 and species == 'coluzzii'"}`.
+        sample_query : str, optional
+            A query statement which maps to the sample metadata e.g.,
+            "species == 'coluzzii' and country == 'Burkina Faso'"
```
alimanfoo, line 1141:
```suggestion
        sample_query : str, optional
            A pandas query string which will be evaluated against the sample metadata e.g.,
            "species == 'coluzzii' and country == 'Burkina Faso'".
```

malariagen_data/ag3.py
```diff
@@ -2060,6 +2076,9 @@ def gene_cnv_frequencies(
...
```
alimanfoo, line 2081:
```suggestion
        sample_query : str, optional
            A pandas query string which will be evaluated against the sample metadata e.g.,
            "species == 'coluzzii' and country == 'Burkina Faso'".
```

malariagen_data/ag3.py
```diff
@@ -2121,16 +2150,17 @@ def gene_cnv_frequencies(
...
-                is_amp_coh = np.compress(loc_samples, is_amp, axis=1)
-                is_del_coh = np.compress(loc_samples, is_del, axis=1)
+        for coh, loc_coh in coh_dict.items():
+            # handle sample query
+            if loc_samples is not None:
+                loc_coh = loc_coh & loc_samples
+            n_samples = np.count_nonzero(loc_coh)
+            # not working with this chunk
+            # if n_samples == 0:
+            #     raise ValueError(f"no samples for cohort {coh!r}")
```
alimanfoo, line 2160:
```suggestion
```

Can delete this.

alimanfoo (commented):
Looks good, couple of tiny doco nits.

Do we need to add this to `aa_allele_frequencies()` also?

alimanfoo:
Also looks like tests need to be updated.

alimanfoo:
Hi @cclarkson, all LGTM, think we can merge this if CI passes. I can't approve the PR as I originally submitted it, but consider it approved :+1:

alimanfoo:
Merged :partying_face:

PR #134, opened 2022-02-09
Expose more imshow parameters in plot_frequencies_heatmap

### LINKED ISSUE #133: Vary height in Ag3.plot_frequencies_heatmap()
If there are quite a few rows, the plot doesn't look good, I think because the height is being fixed to a default plot height plot plotly. The plot is squished and some of the x axis tick labels are missed out. E.g.:

image

I think the fix would be to have a height parameter which is used together with the number of rows on the input dataframe to scale the plot height. Or at least have a height parameter which can be passed through directly to plotly.

alimanfoo:
Resolves #133.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -2539,7 +2539,18 @@ class Ag3:
         return df_aaf
 
     def plot_frequencies_heatmap(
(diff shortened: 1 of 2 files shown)


alimanfoo:
Hi @cclarkson, FYI.

cclarkson:
@alimanfoo - nice :)

PR #140, opened 2022-02-25
Ag3 functions for space-time analysis of variant frequencies

### LINKED ISSUE #139: Move delineation of amp/del from columns to rows in Ag3.gene_cnv_frequencies()
It would be more consistent with other functions producing frequencies dataframes if the frequency columns were restricted to denoting cohorts. For `Ag3.gene_cnv_frequencies()` this would mean adding the CNV type (amp/del) as an additional index variable.

### LINKED ISSUE #143: Add space-time frequencies analysis functions
Add functions for:

* Plotting frequencies time series
* Plotting frequencies on a map

alimanfoo:
This PR proposes some new functions for computing and visualising variant frequencies over space and time, resolves #143.

There are two new main plotting functions:

* `Ag3.plot_frequencies_time_series()` - creates a time series line plot of variant frequencies, with each panel of the plot showing data for a single cohort, and each line showing variant frequencies.

Peek 2022-02-25 10-16 SNP frequencies time series

* `Ag3.plot_frequencies_interactive_map()` - creates an ipyleaflet map with markers showing frequencies and widget controls allowing to select which variant, taxon and time period to show data for.

Peek 2022-02-25 10-13 SNP frequencies interactive map

These new plotting functions can be used with either SNP allele frequencies, amino acid change allele frequencies, or gene CNV frequencies. To make this possible, some new functions have been added which create the necessary input data. These functions compute frequencies similar to the existing functions for computing frequencies, but provide more flexibility in the parameters and provider richer ouput data.

* `Ag3.snp_allele_frequencies_advanced()`
* `Ag3.aa_allele_frequencies_advanced()`
* `Ag3.gene_cnv_frequencies_advanced()`

All the above functions return an xarray Dataset following a set of common conventions regarding variable names.

Also there are some changes to the basic `Ag3.gene_cnv_frequencies()` function for consistency. In particular, resolves #139 by splitting amplifications and deletions into separate rows rather than columns. Also adds a `drop_invariant` parameter which is True by default, similar to SNP and amino acid frequencies functions.

One other change, I found that it is occasionally the case that the same amino acid change can be caused by SNPs at different genome positions. This is potentially a problem for the `Ag3.aa_allele_frequencies...()` methods, because it would be incorrect to sum allele frequencies from SNPs at different positions. Both of these methods have been modified to avoid this, i.e., amino acid changes are only merged where they are due to SNPs at the same genomic position.

### DIFF
--- a/malariagen_data/veff.py
+++ b/malariagen_data/veff.py
@@ -1,6 +1,7 @@
 import collections
 import operator
 
+import pandas as pd
 from Bio.Seq import Seq
 
 VariantEffect = collections.namedtuple(
@@ -167,7 +168,7 @@ class Annotator(object):
         variants["impact"] = limpact
         variants["ref_codon"] = lref_codon
         variants["alt_codon"] = lalt_codon
-        variants["aa_pos"] = laa_pos
+        variants["aa_pos"] = pd.Series(laa_pos, dtype="Int64")
         variants["ref_aa"] = lref_aa
         variants["alt_aa"] = lalt_aa
         variants["aa_change"] = laa_change
--- a/malariagen_data/veff.py
+++ b/malariagen_data/veff.py
@@ -1,7 +1,6 @@
 import collections
 import operator
 
-import pandas as pd
 from Bio.Seq import Seq
 
 VariantEffect = collections.namedtuple(
@@ -120,7 +119,6 @@ class Annotator(object):
         feature_stop = feature.end
         feature_strand = feature.strand
         for row in variants.itertuples(index=True):
-
             # some parameters
             chrom = feature_contig
             pos = row.position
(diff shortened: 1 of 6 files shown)


alimanfoo:
Btw I'm not too happy about the naming of the functions with the "_advanced" suffix, but couldn't think of anything better just yet. Basically we just need some way to differentiate between the simpler functions like "snp_allele_frequencies()" which return a pandas DataFrame and can be used to plot a heatmap, and the more advanced/complex functions which return an xarray Dataset and can be used to plot time series and maps.

alimanfoo:
cc @cclarkson, @jonbrenas, @KellyLBennett, @leehart - this is still work in progress but any early feedback welcome.

malariagen_data/veff.py
```diff
@@ -162,11 +163,12 @@ def get_effects(self, transcript, variants):
             lalt_aa.append(effect.alt_aa)
             laa_change.append(effect.aa_change)
 
+        variants["transcript"] = transcript
         variants["effect"] = leffect
         variants["impact"] = limpact
         variants["ref_codon"] = lref_codon
         variants["alt_codon"] = lalt_codon
-        variants["aa_pos"] = laa_pos
+        variants["aa_pos"] = pd.Series(laa_pos, dtype="Int64")
```
alimanfoo, line 171: This allows for missing values without forcing use of floats.

alimanfoo:
Hi @cclarkson, @leehart, this PR has become a beast I'm afraid, but please do let me know if you have any suggestions or questions. As we're under some time pressure I may merge soon without review, but very happy to follow up any comments you have subsequently.

PR #160, opened 2022-03-04
Add pv4 module

kathryn1995:
Create module for pv4 data that mimics the module created for pf7


kathryn1995:
@podpearson @roamato @ahernank 
Here is the code for adding the plasmodium module for pv4 and pf7. Let me know if you have any comments

podpearson:
All looks good to me - thanks @kathryn1995

alimanfoo:
Hi @kathryn1995, I'd suggest here using class inheritance. The `PlasmodiumTools` class you've defined here is basically a base class, which `Pf7` and `Pv4` could both inherit from. If you did it right, then the sub-classes `Pf7` and `Pv4` would only have to implement `__init__()`, all other methods would be implemented on the base (super) class. This would simplify code and eliminate duplicate documentation. Happy to elaborate.

alimanfoo:
Basically, with a bit of tweaking to the method names on the `PlasmodiumTools` base class, you could end up with a sub-class implementation like this...

```
class Pf7(PlasmodiumTools):
    """Provides access to data from the Pf7 release.

    Parameters
    ----------
    url : str, optional
        Base path to data. Default uses Google Cloud Storage "gs://pf7_release/",
        or specify a local path on your file system if data have been downloaded.
    data_config : str, optional
        Path to config for structure of Pf7 data resource. Defaults to config included
        with the malariagen_data package.
    **kwargs
        Passed through to fsspec when setting up file system access.

    Examples
    --------
    Access data from Google Cloud Storage (default):

        >>> import malariagen_data
        >>> pf7 = malariagen_data.Pf7()

    Access data downloaded to a local file system:

        >>> pf7 = malariagen_data.Pf7("/local/path/to/pf7_release/")

    """

    def __init__(
        self,
        url=None,
        data_config=None,
        **kwargs,
    ):
        if data_config is None:
            working_dir = os.path.dirname(os.path.abspath(__file__))
            data_config = os.path.join(working_dir, "pf7_config.json")
        super().__init__(data_config=data_config, url=url)

```

kathryn1995:
Thanks @alimanfoo! I thought there must be a better way to do this, but didn't know this existed. I will convert this PR to a draft, make the changes and open it again when it's ready

kathryn1995:
Thanks @alimanfoo! I thought there must be a better way to do this, but didn't know this existed. I will convert this PR to a draft, make the changes and open it again when it's ready

malariagen_data/pv4.py
```diff
@@ -0,0 +1,43 @@
+import os
+
+from malariagen_data.plasmodium import PlasmodiumTools
+
+
+class Pv4(PlasmodiumTools):
+    """Provides access to data from the Pv4 release.
+    Parameters
```
alimanfoo, line 8:
```suggestion

    Parameters
```

Newline.

malariagen_data/pf7.py
```diff
@@ -1,27 +1,9 @@
...
 
-DIM_ALT_ALLELE = "alt_alleles"
-DIM_STATISTICS = "sb_statistics"
-DIM_GENOTYPES = "genotypes"
-
-
-class Pf7:
+class Pf7(PlasmodiumTools):
     """Provides access to data from the Pf7 release.
     Parameters
```
alimanfoo, line 8:
```suggestion

    Parameters
```

Newline.

malariagen_data/plasmodium.py
```diff
@@ -0,0 +1,190 @@
...
+    DIM_PLOIDY,
+    DIM_SAMPLE,
+    DIM_VARIANT,
+    da_from_zarr,
+    init_filesystem,
+    init_zarr_store,
+)
+
+
+class PlasmodiumTools:
```
alimanfoo, line 20: Suggest calling this "PlasmodiumDataResource" or "PlasmodiumDataRelease", seems a bit more natural and descriptive.

alimanfoo (commented):
Hi @kathryn1995, looks good. A couple of very minor suggestions, otherwise good to go I reckon.

alimanfoo:
Thanks @kathryn1995. By the way it's nice to be thinking about this now for Plasmodium, I'm expecting we'll need something similar on the Anopheles side when funestus data comes along.

kathryn1995:
Hi @alimanfoo, I have made those small changes. Nice to hear this might be useful for Ag too!

PR #198, opened 2022-04-01
PCA functions

### LINKED ISSUE #187: Add Ag3 pca function
Add a function for running a PCA, can probably use [this one](https://malariagen.github.io/vector-data/ag3/examples/pca.html#pca-functions).

### LINKED ISSUE #197: Add Ag3 PCA plotting function

### LINKED ISSUE #203: Simplify handling of analysis version parameters
Currently the analysis version parameters like `cohorts_analysis`, `species_analysis` and `site_filters_analysis` are offered on every Ag3 method for which they are needed. This however clutters the API and means we end up passing the values around a lot internally. Simplify this by setting the parameters in the `Ag3` constructor, rather than in each method.

alimanfoo:
My contribution to the PCA implementation showdown :)

* Adds functions for computing a PCA; resolves #187
* Adds functions for plotting PCA results; resolves #197  
* Supporting this, adds functions for computing SNP allele counts (to find suitable sites for analysis).
* Supporting this, also adds capabilities for automatically saving (caching) results of SNP allele counts and PCA analyses, so these don't have to be re-run.
* As part of this, rework how the analysis version parameters are handled - these are now set when instantiating the Ag3 class, rather than in each method, which simplifies things quite a lot as we no longer need to pass these around everywhere; resolves #203.
* Also add some support for logging info messages to notify when a long-running computation is about to begin, as well as debug logging.

TODO:

* [x] Tweak notebook so it runs faster, and tidy it up
* [x] Add unit tests


alimanfoo:
cc @cclarkson

cclarkson:
Hi @alimanfoo, looks like poetry is struggling to install?

alimanfoo:
Thanks @cclarkson, I've updated poetry and added some unit tests. Will merge and release tomorrow morning if CI passes.

PR #227, opened 2022-04-29
Plot SNPs and site filters

alimanfoo:
* Adds Ag3.plot_snps() and Ag3.plot_snp_tracks()
* Adds Ag3.snp_variants() to support the above plotting functions, with some refactoring inside Ag3.snp_calls()
* Upgrade poetry


alimanfoo:
image

PR #229, opened 2022-04-29
Prune sample sets if a sample query is given

### LINKED ISSUE #225: Possible small optimisation - prune sample sets using sample query
A common pattern is to not specify a value for the `sample_sets` parameters, which defaults to all available sample sets, then give a `sample_query`. This is convenient, but then can cause a few delays because not all sample sets are needed when accessing other data like SNP calls.

A possible small optimisation would be to use the sample query to prune the list of sample sets, before doing anything else.

alimanfoo:
* Resolves #225

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -594,9 +594,11 @@ class Ag3:
         allow this to be a single sample set, or a list of sample sets, or a
         release identifier, or a list of release identifiers."""
 
+        all_sample_sets = self.sample_sets()["sample_set"].tolist()
+
         if sample_sets is None:
-            # all available sample sets
-            sample_sets = self.sample_sets()["sample_set"].tolist()
+            # use all available sample sets
+            sample_sets = all_sample_sets
 
         elif isinstance(sample_sets, str):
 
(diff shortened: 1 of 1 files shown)


alimanfoo:
Hm, how to avoid sample sets progress being shown twice?

malariagen_data/ag3.py
```diff
@@ -1004,7 +1018,9 @@ def snp_genotypes(
         debug = self._log.debug
 
         debug("normalise parameters")
-        sample_sets = self._prep_sample_sets_arg(sample_sets=sample_sets)
+        sample_sets = self._prep_sample_sets_arg(
+            sample_sets=sample_sets, sample_query=sample_query
```
cclarkson, line 1022: I'm not sure where `sample_query` is coming from here, it doesn't appear a parameter in `def species_calls()`

malariagen_data/ag3.py
```diff
@@ -638,6 +643,15 @@ def _prep_sample_sets_arg(self, *, sample_sets):
                     f"Bad value for sample_sets parameter, {k:!r} selected more than once."
```
cclarkson, line 643: could we just take unique sample sets here?

cclarkson, line 643: could we just take unique sample sets here?

alimanfoo:
Probably should benchmark to confirm this does save time.

Also sample metadata gets loaded twice, so two progress bars get shown, which is not so great.

PR #232, opened 2022-05-09
Plasmodium access annotations

kathryn1995:


kathryn1995:
This PR also fixes a bug in ag3. When specifying that you would like all of the annotations fields returned with "*", it used to add an empty column named "*". This now retrieves all of the fields and puts them into columns in the df

alimanfoo:
Hi @kathryn1995, CI is fixed in master, so you should be able to fix CI on this PR with something like:

```
git checkout master
git pull
git checkout plasmodium_access_annotations
git merge master
git push
```

malariagen_data/plasmodium.py
```diff
@@ -188,3 +191,34 @@ def variant_calls(self, extended=False, inline_array=True, chunks="native"):
...
+        Returns
+        -------
+        df : pandas.DataFrame
+
+        """
+        # Attributes
+        if type(attributes) not in [tuple, list] and attributes != "*":
+            raise TypeError("'attributes' must be a list, tuple, or '*'")
+        if attributes is not None:
+            attributes = tuple(attributes)
```
alimanfoo, line 212: Could you factor this out into a separate function so we can use it in Ag3 and Amin1. E.g., in malariagen_data.util:

```python
def _prep_geneset_attributes_arg(attributes):
    if type(attributes) not in [tuple, list] and attributes != "*":
        raise TypeError("'attributes' must be a list, tuple, or '*'")
    if attributes is not None:
        attributes = tuple(attributes)
    return attributes
```

Then code here (and in Ag3 and Amin1) could become:

```suggestion
        attributes = _prep_geneset_attributes_arg(attributes)
```

alimanfoo (commented):
Hi @kathryn1995, looks good, just two small comments.

malariagen_data/plasmodium.py
```diff
@@ -188,3 +191,34 @@ def variant_calls(self, extended=False, inline_array=True, chunks="native"):
         ds = xarray.Dataset(data_vars=data_vars, coords=coords)
 
         return ds
+
+    def genome_features(self, attributes=("ID", "Parent", "Name", "alias")):
```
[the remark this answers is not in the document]
alimanfoo: Btw if you feel strongly that "genome_features()" is a better name then we could change Ag3 and Amin1. It probably is a better name, "geneset" is a term for genome annotations that comes from vectorbase, but doesn't look like it's used much elsewhere.

If we do change Ag3 and Amin1 it would be good to provide backwards compatibility, which would mean keeping a geneset() method as an alias for genome_features(). There's an example of how we did this when the method "snp_dataset()" was renamed to "snp_calls()", see [here](https://github.com/malariagen/malariagen-data-python/blob/5ff3481c515fd7ef6efb93b2481d2186f053a396/malariagen_data/ag3.py#L1895).

Whatever we do, we should be consistent across data releases where we can.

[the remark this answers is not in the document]
kathryn1995: Hi, when we discussed the name I think we were all in agreement about changing to genome_features as our annotations don't just contain genes, so it was a more accurate description. 

I should have changed it everywhere it needs changing and have added an alias for amin1 and ag3

malariagen_data/plasmodium.py
kathryn1995: I've moved this out into util.py now

kathryn1995:
@alimanfoo Thanks for reviewing this, sorry it took me a while to get round to it. I should have addressed your comments

malariagen_data/ag3.py
```diff
@@ -1112,7 +1112,13 @@ def genome_sequence(self, region, inline_array=True, chunks="native"):
 
         return d[loc_region]
 
-    def geneset(self, region=None, attributes=("ID", "Parent", "Name", "description")):
+    def geneset(self, *args, **kwargs):
+        """Deprecated, this method has been renamed to snp_calls()."""
```
alimanfoo, line 1116:
```suggestion
        """Deprecated, this method has been renamed to genome_features()."""
```

malariagen_data/amin1.py
```diff
@@ -117,7 +117,11 @@ def genome_sequence(self, region, inline_array=True, chunks="native"):
 
         return d[loc_region]
 
-    def geneset(self, attributes=("ID", "Parent", "Name", "description")):
+    def geneset(self, *args, **kwargs):
+        """Deprecated, this method has been renamed to snp_calls()."""
```
alimanfoo, line 121:
```suggestion
        """Deprecated, this method has been renamed to genome_features()."""
```

alimanfoo (commented):
Just two nits, otherwise LGTM.

alimanfoo:
Hi @kathryn1995, sorry again for long hiatus. All looks good here to me. Feel free to merge if CI passes.

kathryn1995:
Hi @alimanfoo, thanks for picking that up. All checks have passed, I just need you to approve the request so I can merge

alimanfoo (approved):

PR #235, opened 2022-05-23
Use blank DataFrame when cohorts metadata are missing

leehart:


malariagen_data/ag3.py
```diff
@@ -3029,21 +3029,49 @@ def _read_cohort_metadata(self, *, sample_set):
...
             path = f"{path_prefix}/cohorts_{self._cohorts_analysis}/{sample_set}/samples.cohorts.csv"
-            with self._fs.open(path) as f:
-                df = pd.read_csv(f, na_values="")
+            # N.B., not all cohorts metadata exist, need to check
+            if self._fs.exists(path):
 
-            # ensure all column names are lower case
-            df.columns = [c.lower() for c in df.columns]
+                with self._fs.open(path) as f:
+                    df = pd.read_csv(f, na_values="")
```
alimanfoo, line 3036: Because of the relatively high latency of GCS, it would be faster here to ask for forgiveness rather than permission. I.e., change this to something like:

```python
try:
    with self._fs.open(path) as f:
        df = pd.read_csv(f, na_values="")
        # ...

except FileNotFoundError:
    # create blank dataframe
```

This is better because it removes the existence checks, each of which will take some time, which will add up if accessing metadata for lots of sample sets.

alimanfoo (changes requested):
Hi @leehart,

The test looks good, thanks for adding that.

Looking again at the main implementation, I'd suggest changing the approach slightly for performance reasons, see comment below.

alimanfoo (approved):
Thanks @leehart, LGTM.

PR #251, opened 2022-06-14
Use blank DataFrame when species calls are missing

### LINKED ISSUE #252: Handle missing species calls
In some cases the species calls for a given release and/or species analysis may be missing. We need to handle this gracefully. Currently species calls are expected to be always present, and so an exception is raised if missing.

leehart:
* Resolves #252


alimanfoo (approved):
This all looks great to me. OK to merge?

PR #261, opened 2022-07-20
Fix sample metadata dtype

### LINKED ISSUE #260: Ag3.plot_pca_coords() triggers TypeError: can only concatenate str (not "int") to str
Some sample sets appear to be triggering this error. Suggests some sample metadata fields (location?) are being inferred by pandas as numeric when they should always be strings.

alimanfoo:
* Resolves #260


alimanfoo:
Hi @ericrlucas, this should fix the problem with plot_pca_coords(), will merge and make a release when CI passes.

EricRLucas:
Cool, thanks

PR #263, opened 2022-07-21
Fix poetry update with explicit pandas version constraints

alimanfoo:
This PR enables poetry update to complete in a reasonable time frame. Previously update was running for >10 mins with no sign of completing. I don't know why this trick makes poetry better able to resolve dependencies, but when debugging the update I noticed that a lot of the messages related to pandas, and so tried this.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        poetry-version: ["1.1.14"]
+        poetry-version: ["1.2.0b3"]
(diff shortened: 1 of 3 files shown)


alimanfoo:
cc @cclarkson

alimanfoo:
Gosh darn it, looks like this is not so clever, poetry install now failing on Pythons 3.8 and 3.9.

alimanfoo:
xref https://github.com/python-poetry/poetry/issues/5121

alimanfoo:
OK, looks like we've found a happy place here, at least for the moment. Will require using poetry 1.2.0b3 or higher, but I reckon that's OK. Merging.

PR #267, opened 2022-07-27
Misc Ag3 tweaks

alimanfoo:
* Add params to Ag3.plot_aim_heatmap()
* Add longitude and latitude to outputs from Ag3 diversity stats functions
* Add Ag3 plot_diversity_stats() function


cclarkson (approved):
LGTM

PR #270, opened 2022-08-03
config.json supplies defaults for ag3

### LINKED ISSUE #255: Ag3 - load current (latest) releases and analysis versions from external resource
Currently the default version for analyses, like the cohorts analysis, are hard-coded in the malariagen_data.ag3 module. This is a bit cumbersome at the moment, because it means that every time we generate a new cohorts analysis, we have to also release a new version of the malariagen_data package with the default analysis version updated. This is particularly annoying where conda environments are being used, and so new environments have to then be created to update the version of malariagen_data package in use.

It would potentially be simpler and more flexible to host a file, e.g., a JSON or YAML file, alongside the data in the release bucket, which has the latest versions of all analyses. This could then be retrieved when the Ag3 class is instantiated to obtain default values for the latest analyses.

cclarkson:
resolves #255

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -1,3 +1,4 @@
+import json
 import sys
 import warnings
 from bisect import bisect_left, bisect_right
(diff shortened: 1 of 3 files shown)


leehart (approved):
LGTM Thanks @cclarkson

PR #272, opened 2022-08-03
GWSS method

### LINKED ISSUE #271: Add GWSS methods to ag3
**Supporting functions**
- load_haplotypes()
Consider moving downsampling upstream to Ag3.haplotypes() by adding a cohort_size parameter, analogous to Ag3.snp_calls()
If so, probably don't need a load_haplotypes() function
- h12_calibration()
needed? 
caching strategy? use on disk results_cache?
set window_sizes default value? 
- plot_h12_calibration()
- h12_gwss()
add support for saving results to on disk results cache (see pca() or snp_allele_counts() for examples of how to do this)
- plot_h12_gwss()
make a bokeh plot of H12 GWSS

cclarkson:
resolves #271 

- [x] add cohort_size parameter to haplotypes() method
- [x] add h12 calibration method
- [x] add h12 calibration plotting method
- [x] add h12 method
- [x] add h12 plotting method
- [x] tests


cclarkson:
When we add a `cohort_size` parameter to ag3.haplotypes in the same way as ag3.snp calls, we are randomly sampling at the level of the individual. In ag3.haplotypes this means that `cohort_size=30` returns 60 haplotypes. Is this desired behaviour? To work at the haplotype level we would need to invoke a new parameter e.g.` haplotype_count` to avoid confusion.

alimanfoo:
> When we add a `cohort_size` parameter to ag3.haplotypes in the same way as ag3.snp calls, we are randomly sampling at the level of the individual. In ag3.haplotypes this means that `cohort_size=30` returns 60 haplotypes. Is this desired behaviour? To work at the haplotype level we would need to invoke a new parameter e.g.` haplotype_count` to avoid confusion.

Hi Chris, yes I think it's fine to keep this at the level of the individual. I.e., cohort_size parameter always means the number of individuals, and so cohort_size=30 will return 60 haplotypes.

cclarkson:
@alimanfoo - this is ready for review. Took a while to figure out a weird caching bug, but it looks like it is sorted now.

cclarkson:
Hi @alimanfoo  - this should be good to go now.

malariagen_data/ag3.py
```diff
@@ -7951,6 +7972,502 @@ def plot_roh(
...
+            sample_query=sample_query,
+            sample_sets=sample_sets,
+            window_sizes=window_sizes,
+            cohort_size=cohort_size,
+            random_seed=random_seed,
+        )
+
+        # compute summaries
+        q50 = [
+            np.median(calibration_runs[window]) for window in calibration_runs.keys()
```
alimanfoo, line 8133: Using `calibration_runs.keys()` here makes me nervous, because it relies on the keys always coming out in the same order as the `window_sizes` parameter was originally provided. This may in fact happen, but I would prefer to guarantee the correct order, e.g., by doing something like:

```suggestion
            np.median(calibration_runs[str(window)]) for window in window_sizes
```

...and similarly below.

malariagen_data/ag3.py
```diff
@@ -7951,6 +7972,502 @@ def plot_roh(
...
+        cohort_size : int, optional
+            If provided, randomly down-sample to the given cohort size.
+        random_seed : int, optional
+            Random seed used for down-sampling.
+
+        Returns
+        -------
+        x, h12 : numpy.ndarray, numpy.ndarray
+            An array containing the window centre point genomic positions (x) and
+            an array with h12 statistic values for each window (h12).
```
alimanfoo, line 8219:
```suggestion
        x : numpy.ndarray
            An array containing the window centre point genomic positions.
        h12 : numpy.ndarray
            An array with h12 statistic values for each window.
```

malariagen_data/ag3.py
```diff
@@ -7951,6 +7972,502 @@ def plot_roh(
...
+        bkplt.show(fig)
+
+    def h12_gwss(
+        self,
+        contig,
+        analysis,
+        window_size,
+        sample_sets,
+        sample_query,
+        cohort_size=20,
```
alimanfoo, line 8187: Suggest 30.

malariagen_data/ag3.py
```diff
@@ -7951,6 +7972,502 @@ def plot_roh(
...
+        return results
+
+    def plot_h12_gwss_track(
+        self,
+        contig,
+        analysis,
+        window_size,
+        sample_sets,
+        sample_query,
+        cohort_size=20,
```
alimanfoo, line 8287: Suggest 30.

malariagen_data/ag3.py
```diff
@@ -7951,6 +7972,502 @@ def plot_roh(
...
+        return fig
+
+    def plot_h12_gwss(
+        self,
+        contig,
+        analysis,
+        window_size,
+        sample_sets,
+        sample_query,
+        cohort_size=20,
```
alimanfoo, line 8396: Suggest 30.

malariagen_data/ag3.py
```diff
@@ -7951,6 +7972,502 @@ def plot_roh(
...
+            size=3,
+            line_width=0.5,
+            line_color="black",
+            fill_color=None,
+        )
+
+        # tidy up the plot
+        fig.yaxis.axis_label = "H12"
+        fig.yaxis.ticker = [0, 1]
+        fig.xaxis.visible = False
```
alimanfoo, line 8382: The pattern I've followed elsewhere is that, for the `plot_foo_track()` functions, the X axis is visible by default, and is styled using the `_bokeh_style_genome_xaxis()` convenience function. It is then the job of the outer `plot_foo()` function to hide the X axis of any track figures. See `plot_cnv_hmm_coverage_track()` and `plot_cnv_hmm_coverage()` for an example.

alimanfoo: Feel like I am nit picking, but consistency does help.

tests/test_ag3.py
```diff
@@ -3134,3 +3174,64 @@ def test_aim_calls(sample_sets, sample_query, aims):
...
+    assert isinstance(x, np.ndarray)
+    assert isinstance(h12, np.ndarray)
+
+    # check dimensions
+    assert len(x) == 11354
+    assert len(x) == len(h12)
+
+    # check some values
+    assert x[0] == 27701.195
+    assert h12[11353] == 0.17875
```
alimanfoo, line 3237: Probably worth using `assert_allclose()` here if comparing floating point values.

alimanfoo (commented):
Hey @cclarkson, looking great. A few nits...

malariagen_data/ag3.py
```diff
@@ -7,6 +7,9 @@
 from textwrap import dedent
 
 import allel
+import bokeh.layouts as bklay
+import bokeh.models as bkmod
+import bokeh.plotting as bkplt
```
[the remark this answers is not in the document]
cclarkson: I've added the imports into the function body.

cclarkson:
Hi @alimanfoo, thanks for comments. I've also added width=DEFAULT_GENOME_PLOT_WIDTH and genes_height=DEFAULT_GENES_TRACK_HEIGHT, in keeping with your previous "track" plots. Should be good to go now.

alimanfoo (approved):

PR #274, opened 2022-09-01
Fixes DEBUG bug

### LINKED ISSUE #273: Ag3 DEBUG logging appears by default
Several users are reporting that DEBUG log messages are appearing under the default configuration of the Ag3 class. This should not happen, DEBUG logging should be off by default.

cclarkson:
Adds logic to logger, if `debug=False`, logger is set to INFO rather than DEBUG.

resolves #273

### DIFF
--- a/malariagen_data/util.py
+++ b/malariagen_data/util.py
@@ -530,7 +530,11 @@ class LoggingHelper:
 
         # set up a logger
         logger = logging.getLogger(name)
-        logger.setLevel(logging.DEBUG)
+        if debug is not False:
+            logger.setLevel(logging.DEBUG)
+        else:
+            logger.setLevel(logging.INFO)
+
         self._logger = logger
 
         # set up handler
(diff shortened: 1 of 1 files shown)


malariagen_data/util.py
```diff
@@ -530,7 +530,11 @@ def __init__(self, name, out, debug=False):
 
         # set up a logger
         logger = logging.getLogger(name)
-        logger.setLevel(logging.DEBUG)
+        if debug is not False:
+            logger.setLevel(logging.DEBUG)
+        else:
+            logger.setLevel(logging.INFO)
```
alimanfoo, line 536:
```suggestion
        if debug:
            logger.setLevel(logging.DEBUG)
        else:
            logger.setLevel(logging.INFO)
```

alimanfoo:
Thanks @cclarkson. Were you able to confirm this makes debug logging stop happening by default in colab?

alimanfoo (approved):

PR #278, opened 2022-09-12
Client location, haplotype progress bar, H12 optional parameters

### LINKED ISSUE #269: Update warning message about client location
The colab menu has changed to "Disconnect and delete runtime", we need to update out warning message about non-US client location.

### LINKED ISSUE #276: Progress not shown when loading haplotypes via Ag3.plot_h12_gwss()
When running an H12 selection scan via `Ag3.plot_h12_gwss()` would be good to see a progress bar while the haplotypes are loaded.

### LINKED ISSUE #277: Make sample_sets parameter optional in Ag3.plot_h12_gwss()
Make the `sample_sets` parameter optional in the `Ag3.plot_h12_gwss()` function. Default value should be `None`, which means default to all available sample sets, following behaviour in other functions.

cclarkson:
* Resolves #269 
* Resolves #276 
* Resolves #277

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -3279,11 +3279,7 @@ class Ag3:
         for r in region:
             ly = []
 
-            haplotypes_iterator = self._progress(
-                sample_sets, desc="Load sample haplotypes"
-            )
-
-            for s in haplotypes_iterator:
+            for s in sample_sets:
                 y = self._haplotypes_dataset(
                     contig=r.contig,
                     sample_set=s,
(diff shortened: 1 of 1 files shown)


cclarkson:
@alimanfoo, I'm not 100% I have the haplotype progress bar in the correct place.

alimanfoo:
Thanks @cclarkson. 

To show progress of a dask computation, you need to do this slightly differently. Basically, at the point where you call `.compute()` (or do something which triggers a dask compute()) you need to put that inside a `with` block using `self._dask_progress(desc="Load haplotypes")`. In this case, the line you need to put under a `with` block is [here](https://github.com/malariagen/malariagen-data-python/blob/7afb2d58f2e0fffa8e43bddbccdd6e35940134c6/malariagen_data/ag3.py#L8418).

An analogous example is [here](https://github.com/malariagen/malariagen-data-python/blob/7afb2d58f2e0fffa8e43bddbccdd6e35940134c6/malariagen_data/ag3.py#L1506), where we show progress of computing a genotype array.

cclarkson:
hi @alimanfoo, I think it is good to go now.

alimanfoo (approved):
Thanks @cclarkson, LGTM.

alimanfoo:
Feel free to merge and make a micro release (6.1.1).

PR #279, opened 2022-09-14
UPDATE TO 6.1.1

cclarkson:
README.md and pyproject.toml updated for version bump with bug fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ from the following locations:
 ## Release notes
 
 
+### 6.1.1
+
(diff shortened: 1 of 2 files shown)


alimanfoo (approved):
Thanks Chris. It looks like one of the CI runs on the release PR has failed due to a connection error, should be an intermittent thing, I've just relaunched that job. Should take around 20 mins to complete. If CI passes then feel free to merge. We can add release notes later.

PR #283, opened 2022-10-01
Implement H1X statistic for detecting shared selective sweeps

alimanfoo:
This PR adds new public methods `Ag3.h1x_gwss`, `Ag3.plot_h1x_gwss` for performing a genome-wide scan for signals of shared selective sweeps between two cohorts.

These functions compute a summary statistic in windows which is similar to Garud's H1 statistic but applied to two cohorts rather than one. Because of this I've called the statistic H1X where the "X" stands for being across two cohorts. For each haplotype, we compute the product of the frequencies in each cohort. We then sum these products to give an H1X value for a window. The interpretation is straightforward, in that H1X provides an estimate for the probability that any two haplotypes sampled one from each cohort are identical.

As with H12, the choice of window size is important for this statistic. However, I think it should be fine to use the existing H12 calibration functions. I.e., first follow the normal process to choose a suitable window size for each cohort for H12. Then for H1X choose something that is at least larger than the smaller of these two window sizes.


alimanfoo:
Here's an example plot showing shared sweep at Vgsc between gambiae and coluzzii cohorts in Mali:

image

alimanfoo:
Hi @cclarkson, @sanjaynagi, here's an implementation of a simple approach to detecting genome regions with share sweeps (i.e., adaptive gene flow). Comments welcome :lotus_position:

cclarkson (approved):
Looks great.

PR #285, opened 2022-10-05
Implement interactive haplotype cluster plot 

### LINKED ISSUE #284: Implement interactive haplotype clustering function
Need to write a function in plotly (or bokeh) which clusters haplotypes and produces an interactive dendrogram, in which one can hover over leaves and retrieve metadata about that haplotype. 

proposed function arguments - 

- region
- sample_sets
- sample_query
- analysis (gamb_colu, gambcolu_arab)

optionally - 
- a height to cut tree (total SNP differences in given region) with which we could colour leaves as being part of a haplotype cluster?

any things I've missed?

I assume we will use hamming distance for the hierarchical clustering. What would be the preferred name for this function?

p.s. in general requesting feedback on my issues, pull requests and code where possible :)

sanjaynagi:
This PR adds a new public method `Ag3.plot_haplotype_clustering`to resolve #284 which hierarchically clusters haplotypes and produces an interactive dendrogram with plotly, allowing users to hover over leaves of the dendrogram and retrieve metadata about that haplotype. The function allows the leaves to be coloured by columns in the sample metadata (default='taxon'), select a linkage method (default='single') and supports downsampling with `cohort_size` parameter. 

Ideally, we could also make use the count_sort parameter of scipy.cluster.hierarchy.dendrogram which would make for clearer dendrograms, however, it is not exposed in plotly.figurefactory.create_dendrogram. mentioned here https://github.com/plotly/plotly.py/issues/3598 

**Importantly**, due to an incompatibility error with flake8 version 3.9.0 and importlib-metadata, I had to change the pre-commit hook configuration file to flake8 version 5.0.1. This may not be acceptable for malariagen_data (?), but it was the only way I could commit..... I think it is related to this issue https://github.com/PyCQA/flake8/issues/1701. I wonder if it could separately be fixed by pinning importlib-metadata to an earlier version?


sanjaynagi:
Here is an example of haplotype clustering at the VGSC in the AG1000G-GH sample set. Leaves are coloured by taxon, highlighting a large haplotype cluster which is a mixture of gambiae/coluzzii haplotypes.

screen_clust

malariagen_data/ag3.py
```diff
@@ -8628,6 +8628,184 @@ def plot_h12_gwss(
...
+        if color == "taxon":
+            _setup_taxon_colors(plot_kwargs)
+
+        debug("apply any user overrides")
+        plot_kwargs.update(kwargs)
+
+        debug("Create dendrogram with plotly")
+        leaf_labels = np.arange(
+            ht.shape[1]
+        )  # set labels as the index which we extract to reorder metadata
```
alimanfoo, line 8741: Nit: for longer comments like these, I'd usually put the comment on it's own line before the code it applies to. Otherwise it causes some slightly odd reformatting of code.

malariagen_data/ag3.py
```diff
@@ -8628,6 +8628,184 @@ def plot_h12_gwss(
...
+        max_dist = _hamming_to_snps(
+            ht.T
+        ).max()  # get the max distance, required to set xmin, xmax, which we need xmin to be slightly below 0
```
alimanfoo, line 8744: See above re placement of long comments.

malariagen_data/ag3.py
```diff
@@ -8628,6 +8628,184 @@ def plot_h12_gwss(
...
+        )
+
+        # Repeat the dataframe so there is one row of metadata for each haplotype
+        df_samples_phased_haps = pd.DataFrame(
+            np.repeat(df_samples_phased.values, 2, axis=0)
+        )
+        df_samples_phased_haps.columns = df_samples_phased.columns
+        df_samples_phased_haps = df_samples_phased_haps[
+            hover_data
+        ]  # select only columns in hover_data
```
alimanfoo, line 8773: See above re placement of long comments.

.pre-commit-config.yaml
```diff
@@ -21,6 +21,6 @@ repos:
     hooks:
       - id: blackdoc
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.9.0
+    rev: 5.0.1
```
alimanfoo, line 24: :+1:

alimanfoo (commented):
Hey @sanjaynagi, looks great, just some minor comments.

Only other thing is... I really would like to be able to use the count_sort parameter, it does make the dendrograms much neater and easier to read. I reckon we should just vendor the whole of the plotly.figure_factory.create_dendrogram module for now and hack it to add the count_sort parameter. I.e., copy source code of plotly.figure_factory.create_dendrogram module into a new module in this package (e.g., malariagen_data.plotly_dendrogram), then hack it. Would you be up for that?

alimanfoo (approved):
Looks great to me, very happy with it.

@cclarkson any comments before merge?

alimanfoo:
Oh, just thought, we probably need to add a legend if either color or symbol parameters are given.

alimanfoo:
Thanks @sanjaynagi for adding the legend, looks good.

Running on a larger number of haplotypes, I'm noticing the Y limits seem to be much larger than necessary. E.g. ...

image

Here the largest distance is less than 70 SNPs, but the Y axis goes up to 160 for some reason. Any idea what's going on there?

sanjaynagi:
Ahh, well spotted. Sorry, should have checked more thoroughly. 

It's because I had used the same distance function (`_hamming_to_snps`) to find the maximum distance (ymax), but this does not take into account the linkage method used. So before, the ymax was only correct if we used 'complete' linkage. 

create_dendrogram requires separate linkage and distance functions, so Ive added another commit which adds a separate function `_get_max_hamming_distance`(using scipy.heirarchy.cluster.linkage), which takes a linkage_method argument and returns the correct value of ymax.

cclarkson:
This looks great, nice work @sanjaynagi. I've compared results using the function to previous analyses and it matches nicely. The only minor things I noted were that the population legend order seemed a bit random, and needing to scroll through the legend when analysing lots of populations is a bit tricky, but both these things might just be plotly "features"!

alimanfoo:
Hi @sanjaynagi, just to say I pushed a commit here which resolves merge conflicts with master. Make sure to pull before you do any further work.

alimanfoo:
Suggest we merge here and deal with adding support for ordering legend items via a follow up. Will merge if CI passes.

PR #287, opened 2022-10-12
Haplotype median-joining networks implementation

### LINKED ISSUE #286: Implement interactive haplotype network function
Create a function which builds a median-joining network from haplotypes from a given gene or locus of interest, then creates a visualisation of the network which allows panning and zooming etc.

alimanfoo:
This PR resolves #286 by adding a new function Ag3.plot_haplotype_network() which constructs a median-joining network from haplotypes and then displays using dash-cytoscape.


alimanfoo:
image

alimanfoo:
image

alimanfoo:
image

alimanfoo:
N.B., I think we have to label this as a bit "experimental" for the moment, as there are a few usability wrinkles. Particularly when running my local computer, if I rerun the same cell I get an error which I think is due to the fact that jupyter-dash is failing to shut down the previous app before running a new app on the same port. This doesn't seem to happen on colab for some reason. Anyway, this can be worked around by providing the server_port parameter.

cclarkson:
Oooh, looks cool!

alimanfoo:
Hi @cclarkson, @sanjaynagi, this one's ready for review. Some gnarly stuff in here for sure but very happy to try and explain what's going on.

alimanfoo:
Here's [a colab notebook](https://colab.research.google.com/drive/1qLP6jvk3qld8kVp8PqYO7Acu09kz_v1r?usp=sharing) with some more examples if you want to take it for a spin.

sanjaynagi (commented):
Awesome! looks great to me. 

some of it does indeed look gnarly to work through!

alimanfoo:
Thanks chaps. I'll add the docstring then merge soon if no objections.

PR #289, opened 2022-10-18
Af1.0

leehart:


leehart:
@alimanfoo I still need to resolve conflicts, test notebooks ("Community Edition" PyCharm doesn't have that, btw) and resolve any feedback.

@jorels1 mentioned that he might be able to get the ref genome seq annotations zarr produced for Af1 soon, so we might be able to squeeze that in too, if there's time.

alimanfoo:
> test notebooks

Thanks Lee. To do this you can just do:

```
poetry run jupyter nbconvert --execute notebooks/*  --inplace
```

alimanfoo:
> > test notebooks
> 
> Thanks Lee. To do this you can just do:
> 
> ```
> poetry run jupyter nbconvert --execute notebooks/*  --inplace
> ```

(This is what is run via github actions.)

malariagen_data/anopheles.py
```diff
@@ -0,0 +1,2213 @@
...
+    unpack_gff3_attributes,
+    xarray_concat,
+)
+
+DEFAULT_SITE_FILTERS_ANALYSIS = "dt_20200416"
+DEFAULT_GENOME_PLOT_WIDTH = 800  # width in px for bokeh genome plots
+DEFAULT_GENES_TRACK_HEIGHT = 120  # height in px for bokeh genes track plots
+
+
+class Anopheles(ABC):
```
alimanfoo, line 50: Suggest rename to "AnophelesDataResource" for naming consistency with plasmodium modules.

malariagen_data/af1.py
```diff
@@ -0,0 +1,747 @@
...
+
+    def _site_mask_ids(self):
+        if self._site_filters_analysis == "dt_20200416":
+            return ["funestus"]
+        elif self._site_filters_analysis == "sc_20220908":
+            return ["funestus"]
+        else:
+            raise ValueError
+
+    def _snp_variants_dataset(self, *, contig, inline_array, chunks):
```
alimanfoo, line 245: Could this method be moved to the base class?

malariagen_data/af1.py
```diff
@@ -0,0 +1,747 @@
...
+
+        debug("set up attributes")
+        attrs = {"contigs": self.contigs}
+
+        debug("create a dataset")
+        ds = xr.Dataset(data_vars=data_vars, coords=coords, attrs=attrs)
+
+        return ds
+
+    def wgs_data_catalog(self, sample_set):
```
alimanfoo, line 289: Could this method be moved to the base class?

malariagen_data/af1.py
```diff
@@ -0,0 +1,747 @@
...
+        self,
+        region,
+        sample_sets=None,
+        sample_query=None,
+        site_mask=DEFAULT_SITE_MASK,
+        width=800,
+        height=120,
+        max_snps=200_000,
+        show=True,
+    ):
```
alimanfoo, line 542: There are several functions like this, where the implementation is fully generic and could be moved to the base class, but there is a parameter (in thise case `site_mask`) where the default value needs to be different in the child classes.

It would be good to move these to the base class somehow. There's several ways I could think of doing it. Perhaps the simplest would be to set the default value of the `site_mask` parameter to the string "default". Then the implementation could check if the value is "default" and replace it with the value of a class or instance attribute (e.g., `self._default_site_mask`) which is set in the child class.

malariagen_data/af1.py
```diff
@@ -0,0 +1,747 @@
...
+
+        taxon_palette = px.colors.qualitative.Plotly
+        taxon_color_map = {
+            "funestus": taxon_palette[0],
+            "gcx1": taxon_palette[1],
+            "gcx2": taxon_palette[2],
+            "gcx3": taxon_palette[3],
+        }
+        plot_kwargs["color_discrete_map"] = taxon_color_map
+        plot_kwargs["category_orders"] = {"taxon": list(taxon_color_map.keys())}
```
alimanfoo, line 747: The taxon values "gcx1", "gcx2" and "gcx3" are only applicable to Ag3. 
```suggestion
    @staticmethod
    def _setup_taxon_colors(plot_kwargs):
        import plotly.express as px

        taxon_palette = px.colors.qualitative.Plotly
        taxon_color_map = {
            "funestus": taxon_palette[0],
        }
        plot_kwargs["color_discrete_map"] = taxon_color_map
        plot_kwargs["category_orders"] = {"taxon": list(taxon_color_map.keys())}
```

malariagen_data/af1.py
```diff
@@ -0,0 +1,747 @@
...
+        This computation may take some time to run, depending on your computing
+        environment. Results of this computation will be cached and re-used if
+        the `results_cache` parameter was set when instantiating the Ag3 class.
+
+        """
+        debug = self._log.debug
+
+        # change this name if you ever change the behaviour of this function, to
+        # invalidate any previously cached data
+        name = PCA_RESULTS_CACHE_NAME
```
alimanfoo, line 493: Perhaps this could be a class variable, rather than a module variable? That would make it easier to move this function to the base class?

alimanfoo (commented):
Hi @leehart, looking great, thanks so much for tackling this! I've looked through the implementations and direction all looks good. Just a couple of suggestions about opportunities to move some more code into the base class.

I'll review the tests separately.

alimanfoo (commented):
Hi @leehart, just to say the approach for the unit tests LGTM, happy to go with this. There may be some opportunities to factor out more common tests in future but happy to improve that in future.

malariagen_data/af1.py
leehart: Of course, thanks! Will fix.

malariagen_data/af1.py
leehart: Yes, that shouldn't be there, thanks. Will fix.

malariagen_data/af1.py
leehart: The complication with generalising this one is the "special handling for 3.0 as data catalogs have a different format", but I'll take another look. I could move Af1's version to the parent class and have Ag3 override it. I suppose we can assume that Af1's version will/should be the standard, going forward.

leehart: ^ will do.

malariagen_data/af1.py
leehart: Yes, I agree, only I feel it might be slightly cleaner to default to `None` in the parent class, then check for `None` and replace that with the value an abstract property set by the child class, as you suggest. Thanks!

leehart: In the case of `plot_snps_track`, the subclasses actually have non-generic `cols` and `tooltips` too, but perhaps it will be OK to solve those in a similar way. (I wanted to avoid refactoring any functions that required non-trivial work to generalise, to avoid opening cans of worms and delaying the Af1 release, but now I appreciate that these cases are not very far off. Will do.)

malariagen_data/af1.py
leehart: Agreed, will do. Thanks!

alimanfoo:
Hi @leehart, thanks so much for following up. In general happy to postpone any nontrivial refactoring, your call. I'll approve here and let you merge when you're happy to.

malariagen_data/af1.py
alimanfoo: > Yes, I agree, only I feel it might be slightly cleaner to default to `None` in the parent class, then check for `None` and replace that with the value an abstract property set by the child class, as you suggest. Thanks!

Not sure exactly what you mean here, but just flagging that we still need to allow for user to supply None meaning apply no site mask at all.

alimanfoo (approved):

malariagen_data/af1.py
leehart: That's a good point, thanks. (Sorry, I didn't mean to prompt any responses this week, so extra thanks!)

leehart:
> Hi @leehart, thanks so much for following up. In general happy to postpone any nontrivial refactoring, your call. I'll approve here and let you merge when you're happy to.

Thanks @alimanfoo . Sorry I didn't meant to prompt you. I will be updating the PR shortly with resolved merge conflicts and additional changes. I believe that @cclarkson has kindly agreed to take a look in your stead.

malariagen_data/af1.py
leehart: FWIW, I'll use "default" as the default value instead of `None`, as you originally suggested, and I've checked that specifying `site_mask=None` when calling does indeed override the default of "default" to be `None`, rather than defaulting to "default" instead of `None`, if that makes sense.

leehart:
It looks like the recently added `tests/test_af1.py::test_site_annotations` is failing for some reason. Will investigate/fix.

leehart:
It looks like a memory issue (exit code 137).

leehart:
OK, CI checks have passed, after temporarily avoiding the memory issue.  

@cclarkson -- It would be very reassuring to have your eyes on this too, please? Just because there has been quite a lot of reshuffling of the ~cabinet~ code, and I wouldn't want to break the ~economy~ package.

cclarkson:
Hi @leehart, it all looks sensible to me. I haven't had a chance to run the code today, but if the tests are passing then I reckon it is good to go.

leehart:
> Hi @leehart, it all looks sensible to me. I haven't had a chance to run the code today, but if the tests are passing then I reckon it is good to go.

OK, thanks @cclarkson . I'll bite the bullet and merge this in now.

PR #292, opened 2022-10-27
v7.0.0 release prep

leehart:


leehart:
@cclarkson Merging and releasing soon, unless objections.

PR #296, opened 2022-11-07
adds title to plot_haplotype_clustering function 

sanjaynagi:
Adding a title to `plot_haplotype_clustering()` function with information on selected sample set/sample query, the genomic region and the number of SNPs found in that window. 

I added a line for sample query, similar to H12/H1X functions. Not sure if that is really desirable, let me know and ill change it. Also have added prefix to title like `sample query:` or `number of SNPs:`, happy to remove if preferred. 

Currently looks like this:

image

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -6138,9 +6138,20 @@ class Ag3(AnophelesDataResource):
             hoverinfo="y",
             line=dict(width=0.5, color="black"),
         )
+
+        if not sample_sets and not sample_query:
+            title = f"genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+        elif not sample_sets:
+            title = f"sample query:    {sample_query}<br>genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+        elif not sample_query:
+            title = f"sample sets:    {sample_sets}<br>genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+        else:
+            title = f"sample sets:    {sample_sets},    sample query:    {sample_query}<br>genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+
         fig.update_layout(
             width=width,
             height=height,
+            title=title,
             autosize=True,
             hovermode="closest",
             plot_bgcolor="white",


malariagen_data/ag3.py
```diff
@@ -6138,9 +6138,20 @@ def plot_haplotype_clustering(
...
+
+        if not sample_sets and not sample_query:
+            title = f"genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+        elif not sample_sets:
+            title = f"sample query:    {sample_query}<br>genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+        elif not sample_query:
+            title = f"sample sets:    {sample_sets}<br>genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+        else:
+            title = f"sample sets:    {sample_sets},    sample query:    {sample_query}<br>genomic window:    {region}<br>number of SNPs:    {ht.shape[0]}"
+
```
alimanfoo, line 6150: This could avoid code duplication, e.g.:

```suggestion

        title_lines = []
        if sample_sets is not None:
            title_lines.append(f"sample sets: {sample_sets}")
        if sample_query is not None:
            title_lines.append(f"sample query: {sample_query}")
        title_lines.append(f"genomic region: {region}")
        title_lines.append(f"number of SNPs: {ht.shape[0]}")
        title = "<br>".join(title_lines)

```

alimanfoo (commented):
Thanks @sanjaynagi, small suggestion...

sanjaynagi:
awesome, thanks!

sanjaynagi:
Actually, an issue with this is it leads to titles with four lines, which end up encroaching the plot. I had previously put sample set and sample query on the same line if they are both specified. thoughts? could move the number of SNPs to the same line as genomic region? this might be nicer than having three lines for title anyway 

image

sanjaynagi:
for example: 
image

sanjaynagi:
pushed the change but happy to modify if preferred.

alimanfoo:
THanks @sanjaynagi, three line solution looks good to me. Very minor suggestion would be to put number of SNPs in brackets so it stands out a little more, e.g.:

**genomic region: 2L:2,410-000-2,430,000 (3,351 SNPs)**

Also note that the sample sets list can get very long, but can figure out how to deal with that later.

sanjaynagi:
Thanks! did think it needed to stand out more. Done.

alimanfoo (approved):

alimanfoo:
Thanks @sanjaynagi :sunglasses:

PR #303, opened 2022-11-29
Fix WGS data catalogs for Ag3.0 release to use Sanger S3 

alimanfoo:
* Resolves https://github.com/malariagen/malariagen-data-python/issues/302


alimanfoo:
Currently broken as CORS headers not set on all buckets. Have raised that internally, will revisit here when CORS is fixed.

Issue #306, opened 2022-12-20
Af1.geneset() does not include AFUN identifiers 


sanjaynagi:
The gff table from `Af1.geneset()` does not include AFUN identifiers anywhere, which most users will need to search for genes (and for me, prevents use for cDNA primers in AgamPrimer). 

If possible, it would be better to do this in the original google cloud/package data than some sort of post-load table merging, realistically anyone using the geneset is going to need AFUN gene/transcript identifiers. 

I don't quite know how these tables are initially generated, but ideally the ID column will contain AFUN identifiers, as in ag3.

sanjaynagi:
@alimanfoo @jorels1 @leehart not knowing exactly how the google cloud storage works, is this a reasonable suggestion? (to edit the table and reupload to GCS) 

also i see now that `geneset()` has been deprecated in favour of `genome_features()`. 

happy to do this myself if I have access.

alimanfoo:
Hi @sanjaynagi, the Af1.0 data release uses a new funestus assembly ("idAnoFuneDA-416_04" - catch name :smile:). I actually don't know where the GFF came from - @jorels1? 

This assembly does seem to be [on vectorbase](https://vectorbase.org/vectorbase/app/record/dataset/TMPTX_afunidAnoFuneDA416_04) now ([gff download](https://vectorbase.org/vectorbase/app/downloads/Current_Release/AfunestusidAnoFuneDA416_04/gff/data/)) but I don't think it has AFUN-like gene identifiers yet. Situation actually looks a bit messy as the version we've got in the release bucket is somewhat different from what vectorbase currently has.

sanjaynagi:
cool, cc'ing alex @amakunin who I assume produced the assembly as well

jorels1:
[...]

PR #307, opened 2022-12-22
Fix Af1 genome features

alimanfoo:
The main purpose of this PR is to override the implementation of `genome_features()` in the `Af1` class to use different default GFF attributes, to deal with the fact that the "Name" attribute is not present in the GFF, as described [here](https://github.com/malariagen/malariagen-data-python/issues/306#issuecomment-1361422819).

Also some other maintenance:

* Update the tests to cope with the change to the source GFF where gene identifiers are now slightly different (just "LOC..." instead of "gene-LOC...").
* Slim down the test parameters to reduce test run time a bit.
* Fix up some typos and docstrings.
* Some maintenance on github actions.


tests/test_af1.py
```diff
@@ -96,18 +96,17 @@ def test_site_filters(region):
     assert filter_pass.dtype == bool
 
 
-@pytest.mark.parametrize("chunks", ["auto", "native"])
```
alimanfoo, line 99: No need to cover different values of `chunks` parameter here I think, should already be well covered in tests on Ag3 class.

alimanfoo:
Went ahead with merge in order to cut a release with the fixes for Af1 genome features, but happy to address any comments in a follow up.

PR #327, opened 2023-01-31
Add extra sample metadata

### LINKED ISSUE #319: User-provided sample metadata
Consider adding support for users to provide their own custom sample metadata, which then gets joined into standard metadata.

alimanfoo:
* Adds a `add_extra_metadata()` method to the `AnophelesDataResource` class, which enables users to add extra (custom) sample metadata columns. Also adds a `clear_extra_metadata()` method in case this needs to be reset. Resolves #319. 
* Also speed up tests in `test_anopheles` by using caching of GCS data.


leehart (approved):
nice

PR #330, opened 2023-01-31
Docstrings for Anopheles extra metadata methods

alimanfoo:
Forgot to add docstrings for the extra metadata methods.

Also cleaned up some TODO/FIXME comments which don't seem to need fixing or resolving any more, at least not in my dev environment.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -3791,7 +3791,6 @@ class Ag3(AnophelesDataResource):
             ht.T,
             distfun=lambda x: _hamming_to_snps(x),
             linkagefun=lambda x: linkage(x, method=linkage_method),
-            # FIXME: expected type 'list', got 'ndarray'
             labels=leaf_labels,
             color_threshold=0,
             count_sort=count_sort,
@@ -3941,7 +3940,6 @@ class Ag3(AnophelesDataResource):
 
         from itertools import cycle
 
-        # FIXME: unresolved references
         import dash_cytoscape as cyto
         import plotly.express as px
         from dash import dcc, html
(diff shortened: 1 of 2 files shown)


leehart (approved):
Thanks @alimanfoo .
I added some of those comments after seeing warnings in my own dev environment, during the Af1 refactoring. I have already removed some (in my own working) after seeing that some error suppressions had recently been added, and some warnings had simply disappeared.
It would be good to understand why some of these issues are being flagged and whether they pose actual problems or require any remedies. I suppose we could further suppress warnings or at least provide a comment about it, in case these warnings worry other developers. On the other hand, if there are no real concerns, then it's good to remove such ghost comments, which might only contribute to confusion or clutter.

alimanfoo:
> It would be good to understand why some of these issues are being flagged and whether they pose actual problems or require any remedies. I suppose we could further suppress warnings or at least provide a comment about it, in case these warnings worry other developers. On the other hand, if there are no real concerns, then it's good to remove such ghost comments, which might only contribute to confusion or clutter.

These warnings that come up in pycharm generally seem to be edge cases where pycharm's type checker doesn't work correctly, or where pycharm fails to correctly find functions or methods in some external module. Basically, pycharm does generate spurious warning from time to time. With each new release, pycharm's linting resolves some of these problems, and so spurious warnings may go away after upgrading pycharm. But it's still not perfect, and so spurious warnings will come up from time to time. If they do, I think it's fine to mark them to be ignored via a `# noinspection` comment. 

Ultimately we have the unit tests, plus we have linting via flake8 as a precommit hook and in CI, so those should catch any real problems.

PR #345, opened 2023-03-01
concatenate chromosomal arms in Ag3 and add iHS method 

sanjaynagi:
In this PR I have added functionality to concatenate contigs in `genome_sequence()`, `genome_features()`, `snp_genotypes()`, `snp_calls()` and `haplotypes()`. This functionality has been added solely for Ag3 in `ag3.py`, overwriting these methods and using the `super()` method to call the `AnophelesDataResource` methods. 

A few points:

- How is best to get the maximum value for the right chromosomal arm? - mostly, I have used `genome_sequence('2/3R').shape[0]`. Im assuming here we don't need to worry about any off by one errors, please correct if wrong. 

- To allow 2/3RL input in Ag3, I added a chromosome property (in addition to contig) and added a check in  `resolve_regions` in util.py. However, this means I had to add chromosome properties for Amin, Af1, and plasmodium (in which I just copied values for contigs). I dont know if this is ideal and there may have been an simpler way. 

- The way I've done it, it is not possible to use an input region list mixing contigs and chromosomes such as ['2R', '3RL']. 
- I still need to add tests and add documentation. 

- PS. The docstring for `genome_sequence()` previously stated it can take a list of regions, but I'm pretty sure it can't. I didn't change this yet. 

I have also added a method to perform iHS GWSS scans, with an optional MAF filter and optional custom function to summarise iHS across windows (np.nanmax is default). 

apologies for the lack of separate commits.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -356,7 +356,7 @@ class Ag3(AnophelesDataResource):
             genome = self.open_genome()
             contigs = (
                 region.contig[0] + region.contig[1],
-                region.contig[0] + region.contig[1],
+                region.contig[0] + region.contig[2],
             )
 
             d = da.concatenate(
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -32,6 +32,7 @@ from .util import (
     Region,
     da_from_zarr,
     init_zarr_store,
+    locate_region,
     xarray_concat,
 )
 
@@ -60,6 +61,7 @@ GENOME_REF_ID = "AgamP4"
 GENOME_REF_NAME = "Anopheles gambiae (PEST)"
 
 CONTIGS = "2R", "2L", "3R", "3L", "X"
+CHROMOSOMES = "2RL", "3RL", "X"
 DEFAULT_MAX_COVERAGE_VARIANCE = 0.2
 
 PCA_RESULTS_CACHE_NAME = "ag3_pca_v1"
@@ -68,8 +70,10 @@ FST_GWSS_CACHE_NAME = "ag3_fst_gwss_v1"
 H12_CALIBRATION_CACHE_NAME = "ag3_h12_calibration_v1"
 H12_GWSS_CACHE_NAME = "ag3_h12_gwss_v1"
 H1X_GWSS_CACHE_NAME = "ag3_h1x_gwss_v1"
+IHS_GWSS_CACHE_NAME = "ag3_ihs_gwss_v1"
 
 DEFAULT_SITE_MASK = "gamb_colu_arab"
+DEFAULT = "default"
 
 
 class Ag3(AnophelesDataResource):
@@ -129,6 +133,7 @@ class Ag3(AnophelesDataResource):
     """
 
     contigs = CONTIGS
+    chromosomes = CHROMOSOMES
     _major_version_int = MAJOR_VERSION_INT
     _major_version_gcs_str = MAJOR_VERSION_GCS_STR
     _genome_fasta_path = GENOME_FASTA_PATH
@@ -143,6 +148,7 @@ class Ag3(AnophelesDataResource):
     _h12_calibration_cache_name = H12_CALIBRATION_CACHE_NAME
     _h12_gwss_cache_name = H12_GWSS_CACHE_NAME
     _h1x_gwss_cache_name = H1X_GWSS_CACHE_NAME
+    _ihs_gwss_cache_name = IHS_GWSS_CACHE_NAME
     site_mask_ids = ("gamb_colu_arab", "gamb_colu", "arab")
     _default_site_mask = DEFAULT_SITE_MASK
     _site_annotations_zarr_path = SITE_ANNOTATIONS_ZARR_PATH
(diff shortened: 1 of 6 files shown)


malariagen_data/ag3.py
```diff
@@ -314,6 +320,482 @@ def _repr_html_(self):
...
+                region, inline_array=inline_array, chunks=chunks
+            )
+            return d
+
+        elif region.contig in ["2RL", "3RL"]:
+            genome = self.open_genome()
+            contigs = (
+                region.contig[0] + region.contig[1],
+                region.contig[0] + region.contig[1],
+            )
```
alimanfoo, line 360:
```suggestion
            contigs = (
                region.contig[0] + region.contig[1],
                region.contig[0] + region.contig[2],
            )
```

malariagen_data/ag3.py
```diff
@@ -314,6 +320,482 @@ def _repr_html_(self):
...
+
+            d = da.concatenate(
+                [
+                    da_from_zarr(
+                        genome[contig], inline_array=inline_array, chunks=chunks
+                    )
```
alimanfoo, line 366:
```suggestion
                    super().genome_sequence(region=contig, inline_array=inline_array, chunks=chunks)
```

Small nit, but generally nice to keep things tidy by calling superclass method where possible.

alimanfoo (commented):
Review of the `genome_sequence()` implementation...

Main comment is, would be great to add some tests! (It's often a good approach to write the tests first, especially for tricky refactoring problems like this, but also not a problem add them after.)

Couple of small comments below...

malariagen_data/ag3.py
```diff
@@ -314,6 +320,482 @@ def _repr_html_(self):
...
+                        start=lambda x: x.start + max_r, end=lambda x: x.end + max_r
+                    )
+                    df_part = pd.concat([df_r, df_l], axis=0)
+
+                    debug("apply region query")
+                    if r.end is not None:
+                        df_part = df_part.query(f"start <= {r.end}")
+                    if r.start is not None:
+                        df_part = df_part.query(f"end >= {r.start}")
+                    parts.append(df_part)
```
alimanfoo, line 447: Seemingly small point, but this code is duplicated from the superclass implementation.

It would be nice to avoid code duplication wherever possible, even if small. (Should be possible to avoid any code duplication.)

malariagen_data/ag3.py
```diff
@@ -314,6 +320,482 @@ def _repr_html_(self):
...
+            region = self.resolve_region(region)
+            debug("normalise to list to simplify concatenation logic")
+            if isinstance(region, Region):
+                region = [region]
+
+            if all([r.contig in self.contigs for r in region]):
+                df = super().genome_features(region=region, attributes=attributes)
+                return df
+
+            elif all([r.contig in ["2RL", "3RL"] for r in region]):
```
alimanfoo, line 422: What if someone gives `region=["2RL", "X"]`?

A bit weird, but what if someone gives `region=["2RL", "3L"]`?

alimanfoo (commented):
Couple of thoughts on the `genome_features()` implementation. Feels like the approach isn't quite right but need to think on it some more.

malariagen_data/ag3.py
```diff
@@ -314,6 +320,482 @@ def _repr_html_(self):
...
+
+        if region.start:
+            slice_start = region.start - 1
+        else:
+            slice_start = None
+        if region.end:
+            slice_stop = region.end
+        else:
+            slice_stop = None
+        loc_region = slice(slice_start, slice_stop)
```
alimanfoo, line 379: Also just noting, this code is duplicated from the superclass. Again, would be good to find a way to do this without any code duplciation.

alimanfoo (commented):
Another comment on `genome_sequence()`.

malariagen_data/ag3.py
sanjaynagi: Yeah the first case in particular, jesus. need to write tests. will try a slightly different approach

malariagen_data/ag3.py
sanjaynagi: Yeah, i think I need to split and convert the coordinates instead like 2RL:50,000,000-70,000,000 into 2R:50,000,000-max , 2L:0-xxx

sanjaynagi: I was trying to avoid this before, but it should simplify the code.

alimanfoo:
Hi @sanjaynagi, after thinking about this a bit, I have a suggestion for a slightly different approach. The main issue with the approach in this PR is that there is some duplication of code between parent and child classes. There is also some duplication of docstrings between parent and child classes. Duplication of code or docstrings is generally worthwhile avoiding where possible, to reduce maintenance burden and reduce chances of bugs or inconsistencies creeping in.

I've made a PR #346 which illustrates how this could be done for the `genome_sequence()` method. The general idea is to isolate the logic for obtaining data for a complete contig into a separate private method in the base class, which can then be overridden in the child class to add support for joined contigs. In this case, a new private method `_genome_sequence_array()` takes care of this. This allows the public method `genome_sequence()` to remain in the parent class, and avoids any duplication of code or docstrings.

I reckon we should be able to follow a similar approach at least for the `genome_features()` method and probably others too.

WDYT?

sanjaynagi:
Hey @alimanfoo that sounds and looks great. 

Apologies, in hindsight I should have liaised with you more about the approach. When implementing I was very hesitant to create any new methods in case I overcomplicated things unnecessarily, but that in itself caused over-complication (and code duplication). 

I think it will be cleaner to close this PR and start fresh, and also split the iHS method into a separate PR. I'll just try and do one method at a time this time!

alimanfoo:
Good call, thanks @sanjaynagi 🙏

PR #346, opened 2023-03-01
Ag3.genome_sequence() - add support for joined arms

alimanfoo:
This PR adds support for joined arms to the `genome_sequence()` method. Support is added for virtual whole-chromosome contigs "2RL" and "3RL" formed by joining two chromosome arm contigs together. Based on ideas originally proposed in PR #345.

 The approach taken here is slightly different than #345 because it aims to:

* Have no code duplication between parent and child classes
* Have no duplication of method docstrings between parent and child classes

This is achieved by adding a new private method `_genome_sequence_array()` to the `AnophelesDataResource` base class, which returns an array for a complete contig. This is then overridden in the `Ag3` child class to add support for the joined arm virtual contigs.

This then means that the public `genome_sequence()` method doesn't need to be overridden, but works for all cases.


sanjaynagi:
This is cool @alimanfoo, I hadn't thought about the possibility of calling an `ag3` function from within a function in `AnophelesDataResource`, only the other way around. I know you say its not for merging, but shall we just use this PR for the`genome_sequence() implementation?

alimanfoo:
Thanks @sanjaynagi, yeah if you're happy with it maybe we merge this one.

alimanfoo:
CI failure looks like a data corruption issue. I'll try to relaunch.

alimanfoo:
I've tweaked CI to force a refresh of the GCS cache, in case there was some corrupt data in there causing CI errors.

I also noticed an error about poetry lock file being out of date, so have done a poetry update.

Let's see how CI does now...

alimanfoo:
(The github actions are configured to cache data read from GCS between runs, to speed things up. But it can be that if reading data gets interrupted, the cached data is corrupted.)

alimanfoo:
> I hadn't thought about the possibility of calling an `ag3` function from within a function in `AnophelesDataResource`, only the other way around.

Yeah, when you have an instance of the Ag3 class, it will have all the methods defined in the Ag3 class, and all the methods defined in the parent AnophelesDataResource class. So if you have a method `bar()` and internally that method calls `self.foo()`, Python will first look to see if there is a `foo()` method defined in the child class, and if not look in the parent class. So if `foo()` is defined in both parent and child, then the implementation in the child class will be used.

sanjaynagi:
Yeah, that makes sense! its a very neat implementation, i like it!!

alimanfoo:
Hmm, not sure what's happening here with CI.

alimanfoo:
I get this now, race condition caching a file.

alimanfoo:
Hi @sanjaynagi, merged here, over to you :)

Issue #332, opened 2023-02-03
Support joined-arm chromosomes (2RL and 3RL) in Ag3 


alimanfoo:
For some methods, particularly GWSS, it would be useful to run across a whole chromosome, across the centromere joining the two arms. For Ag3, the reference genome splits chromosome 2 into arms 2R and 2L, and similar for chromosome 3. However, it would be reasonable to support 2RL and 3RL virtually, by concatenating data from the two arms, and offsetting the positions in the left arm by adding the size of the right arm (remembering that the right arms goes first, then the left arm, by convention).

It would then be possible for users to do things like, e.g.:

```
ag3.plot_h12_gwss(region="2RL", ...)
```

...etc.

Note that in the funestus reference genome, arms are already joined, so this behaviour is only needed for Ag3, not Af1.

sanjaynagi:
Sounds good. May also be worth an option or a separate function to run for all chromosomes (2, 3, X) and produce a concatenated plot? Possibly non-bokeh, as exporting the images is a pain. I'm thinking for situations like Sophie's where she's likely to need to do this anyway for publication. Alternatively, we may want to write a little function for this which is not necessarily in malariagen_data.

PR #347, opened 2023-03-05
ag3.genome_features() - add support for joined arms

sanjaynagi:
In this PR I have added support for joined chromosomal arms in Ag3 for the `genome_features()` method, following the approach outlined in PR #346. This contributes towards addressing #332. 

- For the private method `_genome_features_array()`, I chose to follow the `_genome_sequence_array()` naming and keep the suffix '_array' even though technically the returned object is a pandas DataFrame. If preferred, we could change this to `_genome_features_frame()` or `_genome_features_df()`.

### DIFF
--- a/tests/test_ag3.py
+++ b/tests/test_ag3.py
@@ -291,9 +291,11 @@ def test_genome_features_joined_arms(chrom):
     max_r = ag3.genome_sequence(contig_r).shape[0]
     df_l = df_l.assign(start=lambda x: x.start + max_r, end=lambda x: x.end + max_r)
     df_concat = pd.concat([df_r, df_l], axis=0).reset_index(drop=True)
+    df_concat = df_concat.assign(contig=chrom)
     df = ag3.genome_features(region=chrom)
     assert isinstance(df, pd.DataFrame)
     assert df.shape[0] == df_r.shape[0] + df_l.shape[0]
+    assert all(df["contig"] == chrom)
     assert_frame_equal(df, df_concat)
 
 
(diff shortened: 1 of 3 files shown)


tests/test_ag3.py
```diff
@@ -281,6 +281,34 @@ def test_genome_sequence_joined_arms_region(region):
...
+@pytest.mark.parametrize(
+    "region", ["2RL:61,000,000-62,000,000", "3RL:53,000,000-54,000,000"]
+)
+def test_genome_features_joined_arms_region(region):
+    ag3 = setup_ag3()
+    contigs = (region[0] + region[1], region[0] + region[2])
+    df = ag3.genome_features(region=region)
+    assert isinstance(df, pd.DataFrame)
+    assert all(np.isin(df.contig.unique(), contigs))
+    assert all(np.isin(contigs, df.contig.unique()))
```
alimanfoo, line 309: Change to test that contig value in dataframe equals `region.split(":")[0]`.

alimanfoo (commented):
Hi @sanjaynagi, this looks good to me. 

Just one small suggested to change to modify the values in the "contig" column of the output dataframe.

sanjaynagi:
Awesome! will get on it. 

> > * For the private method `_genome_features_array()`, I chose to follow the `_genome_sequence_array()` naming and keep the suffix '_array' even though technically the returned object is a pandas DataFrame. If preferred, we could change this to `_genome_features_frame()` or `_genome_features_df()`.
> 
> I actually wonder if, for clarity, we should rename these private methods to something like `_genome_sequence_for_contig()` and `_genome_features_for_contig()` just to be super clear what their purpose is.

Yeah, I was thinking this, because also we already have some private functions for the other methods like `_open_snp_genotypes()`. Will do.

sanjaynagi:
Ok, have had a go at those :)

alimanfoo (approved):
LGTM :balloon:

alimanfoo:
Thanks @sanjaynagi, I've merged here.

Issue #344, opened 2023-02-28
Add quarter field to sample metadata


alimanfoo:
It would be convenient to add "quarter" field to the sample metadata, derived from month. Values would be an int giving the quarter within a year, -1 if month is missing.

leehart:
`_read_general_metadata()` in `anopheles.py` seems like a convenient place to add this.

leehart:
I think there are at least a couple of ways to do this:
1. Create a `date` column based off the `month` and `year`, so that we can use `pd.PeriodIndex(df['month'], freq='Q')`, which seems to require a `datetime`. (Then replace the blanks with -1.)
2. Manually create bins and labels, so that we can use `pd.cut(df['month'], bins=bins)` and `df['quarter'].map(labels)`.

I think I prefer the first way, in case a bone fide `date` column is useful for other derived values, and it seems more straightforward, and it might avoid human errors. So I'll prepare a PR with that, including tests.
I might have to cater for situations where the year is not available though, maybe by simply providing a default year, which I suppose would invalidate the `date` column for downstream use.

leehart:
Actually, I think I found a simpler way, which seems to work:
```
df['quarter'] = df['month'].replace(-1, pd.NaT).astype('datetime64[M]').dt.quarter
```

PR #349, opened 2023-03-06
Add quarter derived from month to sample metadata. Add tests.

leehart:
Re: https://github.com/malariagen/malariagen-data-python/issues/344


leehart:
To do:
- [ ] Fix tests that don't expect the new `quarter` column.

alimanfoo (approved):
LGTM, thanks @leehart :cherry_blossom:

PR #352, opened 2023-03-08
Ag3 snp_genotypes(), snp_sites(), snp_variants(), snp_calls() - support for joined arms 

sanjaynagi:
In this PR I have added support for joined chromosomal arms in Ag3 for the `snp_genotypes()` and `snp_sites()` methods, following the approaches in PR #346 and #347. This contributes towards addressing #332. 

For the snp_genotypes method - I have renamed and slightly modified the existing method `_snp_genotypes()` to `_snp_genotypes_for_contig()` in the `Anopheles` class and added an `ag3` equivalent method which deals with both arms. Previously, `_snp_genotypes()` took a region as input, but it was only ever called by `snp_genotypes()` and it was only ever passed a contig as input, so I've changed this behaviour so that it only takes a contig as input, just as with the other `for_contig()` methods. This also meant removing a bit of code which could subset data to a region, but which as far as I can see would never be used. 

snp_sites required a new method `_snp_sites_for_contig()` in both classes, as with the previous PRs.


malariagen_data/ag3.py
```diff
@@ -586,6 +586,62 @@ def _genome_features_for_contig(self, *, contig, attributes):
...
+            contig_r, contig_l = _chrom_to_contigs(contig)
+
+            field_r = super()._snp_sites_for_contig(
+                contig=contig_r, field=field, inline_array=inline_array, chunks=chunks
+            )
+            field_l = super()._snp_sites_for_contig(
+                contig=contig_l, field=field, inline_array=inline_array, chunks=chunks
+            )
+
+            # TODO check what other fields are?
```
alimanfoo, line 634: Other fields are just REF and ALT so no need to do anything special with those.

malariagen_data/anopheles.py
```diff
@@ -1479,14 +1486,21 @@ def _snp_sites(
...
+                pos = ret
             else:
-                pos = root[f"{region.contig}/variants/POS"][:]
+                pos = self._snp_sites_for_contig(
+                    contig=region.contig,
+                    field="POS",
+                    inline_array=inline_array,
+                    chunks=chunks,
+                )
             loc_region = locate_region(region, pos)
```
alimanfoo, line 1504: This looks good but just noting there is a subtle change here, because `pos` is now a dask array whereas previously it was a numpy array. This should be OK, because it will I believe get converted to a numpy array (i.e., computed) within the `locate_region` function (actually within `allel.SortedIndex`). We want this to happen, because performing any kind of binary search on a dask array is likely to have poor performance.

tests/test_ag3.py
```diff
@@ -251,23 +251,23 @@ def test_missing_species_calls(analysis):
...
+#     seq = ag3.genome_sequence(region=chrom)
+#     assert isinstance(seq, da.Array)
+#     assert seq.dtype == "S1"
+#     assert seq.shape[0] == seq_r.shape[0] + seq_l.shape[0]
+#     # N.B., we use a single-threaded computation here to avoid race conditions
+#     # when data are being cached locally from GCS (which manifests as blosc
+#     # decompression errors).
+#     assert da.all(seq == da.concatenate([seq_r, seq_l])).compute(
+#         scheduler="single-threaded"
+#     )
```
alimanfoo, line 270: Uncomment.

tests/test_ag3.py
```diff
@@ -358,6 +358,44 @@ def test_snp_sites(chunks, region):
...
+def test_snp_sites_for_contig(chrom):
+    ag3 = setup_ag3()
+    contig_r = chrom[0] + chrom[1]
+    contig_l = chrom[0] + chrom[2]
+    sites_r = ag3.snp_sites(region=contig_r, field="POS")
+    sites_l = ag3.snp_sites(region=contig_l, field="POS")
+    max_r = ag3.genome_sequence(region=contig_r).shape[0]
+    sites = da.concatenate([sites_r, sites_l + max_r])
+
+    sites_joined = ag3.snp_sites(region=chrom, field="POS")
```
alimanfoo, line 371: Very small nit, I'd be inclined to rename `sites` to `sites_expected`, and rename `sites_joined` to `sites_actual`. I find it often useful to use this "expected"/"actual" naming convention to help keep track of what's what in tests.

tests/test_ag3.py
```diff
@@ -358,6 +358,44 @@ def test_snp_sites(chunks, region):
         assert d.shape == d.compute().shape
 
 
+@pytest.mark.parametrize("chrom", ["2RL", "3RL"])
+def test_snp_sites_for_contig(chrom):
```
alimanfoo, line 362:
```suggestion
def test_snp_sites_joined_arms(chrom):
```

Nit, for naming consistency with previous tests.

tests/test_ag3.py
```diff
@@ -358,6 +358,44 @@ def test_snp_sites(chunks, region):
...
+    assert sites_joined.ndim == 1
+    assert sites_joined.dtype == "int32"
+    assert np.all(sites == sites_joined)
+
+
+@pytest.mark.parametrize(
+    "region", ["2RL:61,000,000-62,000,000", "3RL:53,000,000-54,000,000"]
+)
+@pytest.mark.parametrize("field", ["POS", "REF", "ALT"])
+def test_snp_sites_for_contig_region(region, field):
```
alimanfoo, line 383:
```suggestion
def test_snp_sites_joined_arms_region(region, field):
```

tests/test_ag3.py
```diff
@@ -358,6 +358,44 @@ def test_snp_sites(chunks, region):
...
+    ag3 = setup_ag3()
+    sites = ag3.snp_sites(region=region, field=field)
```
alimanfoo, line 385: Could check the shape here too. Don't know exactly what it will be, but could check that shape[0] is less than or equal to size of requested region.

tests/test_ag3.py
```diff
@@ -442,6 +480,37 @@ def test_snp_genotypes_chunks(sample_sets, region):
...
+    assert isinstance(gt, da.Array)
+    assert gt.ndim == 3
+    assert gt.dtype == "i1"
+    assert gt.shape == d.shape
+
+
+@pytest.mark.parametrize(
+    "region", ["2RL:61,000,000-62,000,000", "3RL:53,000,000-54,000,000"]
+)
+def test_snp_genotypes_for_contig_region(region):
```
alimanfoo, line 503:
```suggestion
def test_snp_genotypes_joined_arms_region(region):
```

alimanfoo (commented):
Looks good, just a few small suggestions.

Do you want to add `snp_calls()` in here too? There is already a `_snp_calls_dataset()` method which could be renamed to `snp_calls_for_contig()` and overridden to handled joined arms.

alimanfoo:
> For the snp_genotypes method - I have renamed and slightly modified the existing method `_snp_genotypes()` to `_snp_genotypes_for_contig()` in the `Anopheles` class and added an `ag3` equivalent method which deals with both arms. Previously, `_snp_genotypes()` took a region as input, but it was only ever called by `snp_genotypes()` and it was only ever passed a contig as input, so I've changed this behaviour so that it only takes a contig as input, just as with the other `for_contig()` methods. This also meant removing a bit of code which could subset data to a region, but which as far as I can see would never be used.

Good call on all of this :+1:

tests/test_ag3.py
sanjaynagi: although I still think my poetry/conda still interacting a bit so probably its that

malariagen_data/anopheles.py
sanjaynagi: Yeah I meant to highlight this in the PR and check with you if it was OK but forgot, perhaps it would be good for me to build the PR message as I'm doing the coding.

sanjaynagi:
Awesome, working on `snp_calls_dataset()` now, looks like it will also require a similar refactoring of `_snp_variants_dataset()`

sanjaynagi:
Cool, I've now renamed `snp_calls_dataset()` to `snp_calls_for_contig()`, and done the same for `snp_variants_dataset()` to `snp_variants_for_contig()`, and added equivalent methods in `ag3`. Have added some tests for both. 

I've also added a line of code which appends the virtual contigs as contig attributes to the xarray ds. Wasnt too sure what to do here, but it doesn't seem like those attributes are massively useful (correct if wrong). Behaviour is similar to current situation in that if you request only '2RL', all of the contigs and virtual contigs will appear as attributes.

alimanfoo:
Cool, thanks @sanjaynagi, I'll take a look tomorrow :relaxed:

sanjaynagi:
CI was failing as the contig attribute for snp_calls hadn't been set yet, I just removed the line, lets sort tomorrow

tests/test_ag3.py
alimanfoo: Ouch, segfault is not good! Let me know if it keeps happening.

malariagen_data/anopheles.py
alimanfoo: No worries at all, just thought I'd flag in case we notice any performance issues.

alimanfoo (approved):
All looks tidy to me, thanks @sanjaynagi :cherry_blossom:

alimanfoo:
OK to merge?

alimanfoo:
:tada:

PR #354, opened 2023-03-10
plot_haplotype_clustering() fix - append extra metadata to hover data if specified

sanjaynagi:
This PR fixes a bug where if you specified extra metadata with `add_extra_metadata()` and then tried to use those columns for colors or symbols during `plot_haplotype_clustering()` the extra metadata columns could not be found. 

This is because they needed to be added to a list of hover data columns which is used to subset the metadata.

### DIFF
--- a/tests/test_ag3.py
+++ b/tests/test_ag3.py
@@ -375,7 +375,7 @@ def test_snp_sites_for_joined_arms(chrom):
     assert isinstance(sites_actual, da.Array)
     assert sites_actual.ndim == 1
     assert sites_actual.dtype == "int32"
-    assert np.all(sites_expected == sites_actual)
+    assert da.all(sites_expected == sites_actual)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/test_ag3.py
```diff
@@ -375,7 +375,7 @@ def test_snp_sites_for_joined_arms(chrom):
     assert isinstance(sites_actual, da.Array)
     assert sites_actual.ndim == 1
     assert sites_actual.dtype == "int32"
-    assert np.all(sites_expected == sites_actual)
+    assert da.all(sites_expected == sites_actual)
```
alimanfoo, line 378: This won't actually test anything I don't think, because return value of da.all() is an uncomputed dask array. To test it you'd need to `assert da.all(sites_expected == sites_actual).compute()`

alimanfoo: If you're still getting blosc errors, can try:

```
assert da.all(sites_expected == sites_actual).compute(scheduler='single-threaded')
```

...IIRC we had to do something similar for testing the genome sequence method.

sanjaynagi: Thanks! I've now added this in the #355 PR

alimanfoo (approved):

PR #355, opened 2023-03-10
ag3.haplotypes() - add support for joined arms

sanjaynagi:
In this PR I have added support for joined arms with the `haplotypes()` function, following approaches in #352, #346, #347. Should be the final step in addressing #332. 

As with `snp_calls()` in PR #352 , I have changed `_haplotypes_dataset()` to `_haplotypes_for_contig()`, and implemented an equivalent method in `ag3`. 

Because some sample sets dont have haplotypes, when converting genome coordinates for the left arm of the xr.Dataset, I had to add some logic to say only do this if the left arm ds is not none.

### DIFF
--- a/tests/test_ag3.py
+++ b/tests/test_ag3.py
@@ -761,6 +761,37 @@ def test_snp_variants_for_joined_arms_region(region):
     assert sites.shape[0] == ds_vars["variant_position"].shape[0]
 
 
+@pytest.mark.parametrize("chrom", ["2RL", "3RL"])
+def test_haplotypes_for_joined_arms(chrom):
+    ag3 = setup_ag3()
+    contig_r = chrom[0] + chrom[1]
+    contig_l = chrom[0] + chrom[2]
+    ds_r = ag3.haplotypes(region=contig_r)
+    ds_l = ag3.haplotypes(region=contig_l)
+    ds_expected = xr.concat([ds_r, ds_l], dim="variants")
+
+    ds_actual = ag3.haplotypes(region=chrom)
+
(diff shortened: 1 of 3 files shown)


tests/test_ag3.py
```diff
@@ -761,6 +761,37 @@ def test_snp_variants_for_joined_arms_region(region):
...
+    "region", ["2RL:61,000,000-62,000,000", "3RL:53,000,000-54,000,000"]
+)
+def test_haplotypes_for_joined_arms_region(region):
+    ag3 = setup_ag3()
+    ds_snps = ag3.haplotypes(region=region)
+
+    assert isinstance(ds_snps, xr.Dataset)
+    assert len(ds_snps.dims) == 4
+    assert ds_snps["call_genotype"].dtype == "int8"
+    assert ds_snps["variant_position"].dtype == "int32"
```
alimanfoo, line 792:
```suggestion
    ds_haps = ag3.haplotypes(region=region)

    assert isinstance(ds_haps, xr.Dataset)
    assert len(ds_haps.dims) == 4
    assert ds_haps["call_genotype"].dtype == "int8"
    assert ds_haps["variant_position"].dtype == "int32"
```

Nit, suggest renaming for clarity.

alimanfoo (approved):
Looks good, just one small suggestion.

alimanfoo:
Looks like CI is failing on the same thing you're seeing in #354. Suggest applying fix described [here](https://github.com/malariagen/malariagen-data-python/pull/354#discussion_r1132615657).

PR #360, opened 2023-03-13
implement iHS GWSS 

### LINKED ISSUE #310: Implement IHS GWSS methods for Anopheles
Propose to add functions to run and plot IHS GWSS.

xref https://nbviewer.org/github/alimanfoo/shiny-train/blob/master/notebooks/gwss/ihs_scans.ipynb

sanjaynagi:
This PR adds new GWSS methods to compute and plot the integrated haplotype score - `ihs_gwss()` and `plot_ihs_gwss()`. 

The methods include an optional minor allele frequency filter with the argument `min_maf`. Currently, the default is 0, i.e no MAF filtering, I don't know whats preferable for iHS?

Users can also optionally apply a custom function `window_func` (default=np.nanmax) to summarise the iHS values in windows. This is activated if `window_size` is not None. 

Resolves #310.


malariagen_data/anopheles.py
```diff
@@ -7726,6 +7726,354 @@ def plot_h1x_gwss(
...
+    ):
+        """Run IHS GWSS.
+
+        Parameters
+        ----------
+        contig: str
+            Contig name (e.g., "2L" or "3RL")
+        window_size : int
+            The size of windows used to summarise iHS over.
+            If None, per-variant iHS values are returned.
```
alimanfoo, line 7749:
```suggestion
        window_size : int
            The size of window in number of SNPs used to summarise iHS over.
            If None, per-variant iHS values are returned.
```

alimanfoo (commented):
Small suggestion...

alimanfoo:
Thanks @sanjaynagi. Couple of things interesting to discuss here. I'll put in separate comments below to make it easy reply.

alimanfoo:
Question: Additional parameters? scikit-allel's [ihs](https://scikit-allel.readthedocs.io/en/stable/stats/selection.html#allel.ihs) function supports a number of additional tuning parameters, including `min_ehh`, `min_maf`, `include_edges`, `gap_scale`, `max_gap`, `is_accessible`, `use_threads`. 

For completeness we might want to expose all of these (except `is_accessible`, see below) in the malariagen_data API, although set default values we think are sensible. 

For Anopheles I previously used these defaults:

```
min_ehh=0.05
min_maf=0.05
max_gap=10_000
gap_scale=1_000
include_edges=True
```

These are slightly different from scikit-allel defaults, not sure why I deviated.

Regarding use of the `is_accessible` parameter, we could set that internally, although not sure if it's worth it. Certainly not worth exposing to the user.

alimanfoo:
Question: Standardisation?

Normally when computing IHS we standardise (centre and scale) the data, but do this in allele frequency bins. scikit-allel has a function [`standardize_by_allele_count()`](https://scikit-allel.readthedocs.io/en/stable/stats/selection.html#allel.standardize_by_allele_count) for this. 

I would suggest we do this by default, i.e., standardised the IHS scores before summarising in windows.

But we might want to make this optional, e.g., offer a `standardize=True` parameter which can be set to false to skip. We might also want to expose some of standardisation parameters like `standardization_bins=None`, `standardization_n_bins=None`, `standardization_diagnostics=False` with defaults set.

alimanfoo:
Question: How best to handle minor allele frequency threshold?

This is subtle because there are performance considerations and biological considerations.

The choice is this. You can either:

(1) Filter the haplotype data prior to remove any SNPs below the `min_maf` threshold prior to passing to scikit-allel's `ihs()` function for computing IHS.
(2) Do not filter the haplotype data, but pass a `min_maf` parameter to scikit-allele's `ihs()` function, which will cause a `nan` to be reported for any SNP below the `min_maf` threshold.
(3) Some combination of both (1) and (2).

The advantage of (1) is that the computation is faster, because the IHS algorithm has less data to scan. However, it will affect the influence of the window_size parameter, because the window_size refers to the number of SNPs in the data after applying this filter. Setting a good window size can become a little tricky. There can be a subtle effect here too, where because of a selective sweep in the haplotypes you're computing, there are less segregating SNPs, which effectively extends the physical window size in the neighbourhood of a sweep.

The advantage of (2) is that the window size depends only on the density of SNPs in the panel, not in the particular dataset that you're examining. This means that the window_size parameter will behave in a very similar way to the window_size passed to the H12 functions. I.e., you will get the same number of windows and the same window positions. It is also somewhat robust against variations in SNP density within the particular haplotypes you're computing on, because SNP density comes from the original panel that phasing was done at.

You can also do a bit of a hybrid approach. E.g., [previously](https://nbviewer.org/github/alimanfoo/shiny-train/blob/master/notebooks/gwss/ihs_scans.ipynb) I removed all singletons from the haplotype data before passing to scikit-allel, but then passed through `min_maf=0.05` to scikit-allel.

In general I would lean towards (2) for biological reasons. But this may be too slow, we'd need to see how long it takes to compute.

sanjaynagi:
Awesome, thanks @alimanfoo, all of this sounds great. I like the look of the multiple quartiles etc. I also would lean towards approach 2 for the maf filter. 

Will have a go at these soon but probably tomorrow :)

alimanfoo:
> Question: How best to handle minor allele frequency threshold?
> 
> This is subtle because there are performance considerations and biological considerations.
> 
> The choice is this. You can either:
> 
> (1) Filter the haplotype data to remove any SNPs below the `min_maf` threshold prior to passing to scikit-allel's `ihs()` function for computing IHS.
> 
> (2) Do not filter the haplotype data, but pass a `min_maf` parameter to scikit-allele's `ihs()` function, which will cause a `nan` to be reported for any SNP below the `min_maf` threshold.
> 
> (3) Some combination of both (1) and (2).

OK, scratch my previous discussion on this point, it's been a while and I was forgetting how this works.

With IHS a basic problem is that it is fragile, in the sense that it only takes one mismatch to break EHH between a pair of haplotypes. If that mismatch is a genotyping error or a mutation then EHH is broken erroneously, if otherwise EHH might have continued. So, you might want to filter out some SNPs **before** you compute IHS, particularly singletons, to reduce the chances of falsely breaking EHH. Let's call this `filter_min_maf` meaning the threshold below which you remove SNPs from the haplotype dataset before running IHS.

It's also generally not worth trying to compute IHS for SNPs with very low MAF, because there will only be a small number of haplotypes for the minor allele, and so statistics have high variance. Let's call this `compute_min_maf` meaning the threshold below which you won't bother trying to compute an IHS value for a SNP.

These two values can be different. 

E.g., if `filter_min_maf=0` and `compute_min_maf=0.05` then you won't filter any SNPs out before running IHS, and will only compute IHS at SNPs with 5% MAF or greater (other SNPs will get a nan). 

Or, e.g., if `filter_min_maf=0.05` and `compute_min_maf=0.05` then you will filter out any SNPs below 5% MAF before computing IHS. The `compute_min_maf` threshold in this case has no effect, because you only pass SNPs above this threshold into the IHS computation.

Which approach is correct is hard to say without experimenting in a particular dataset, and it might be worth supporting both these parameters.

A key point, however, is that you want to remove any SNPs with nan IHS values before performing the moving window computation, to ensure that each window has the same number of SNPs with IHS values. This ensures that every window has the same number of data points in it, and so you are less prone to outliers due to windows with only a small number of data points within them.

In other words, if you set a window size of 100 SNPs, then this would result in the same number of windows in both of the two cases above. Because it will be windows of 100 SNPs with IHS values. Which is what you want.

Hope that vaguely makes sense!

alimanfoo:
> E.g., if `filter_min_maf=0` and `compute_min_maf=0.05` then you won't filter any SNPs out before running IHS, and will only compute IHS at SNPs with 5% MAF or greater (other SNPs will get a nan).
> 
> Or, e.g., if `filter_min_maf=0.05` and `compute_min_maf=0.05` then you will filter out any SNPs below 5% MAF before computing IHS. The `compute_min_maf` threshold in this case has no effect, because you only pass SNPs above this threshold into the IHS computation.

To elaborate a bit, in both of these cases you will end up with the same number of SNPs with IHS values.

But the actual IHS values will be different, and in general will be lower for the first case, because you have more SNPs at which to observe mismatches and thus break EHH. In other words, in this first case there are SNPs in the dataset between 0 and 5% MAF where you don't compute IHS but which can contribute to breaking homozygosity between haplotypes.

sanjaynagi:
> Question: Do we want to support multiple summary statistics? E.g., not just max, but 50, 75, 95 percentile too?

How do you think it would be best to approach this @alimanfoo ? We could allow users to provide either a single function or a list of functions in `window_func`? Or if we expect this to be the most common usage, we could just allow the user to pass 'percentiles' or a list of percentiles and do it that way? any thoughts? :)

sanjaynagi:
And yeah, I think the updated description of the maf filter makes sense!

sanjaynagi:
Hey @alimanfoo. I've updated the PR to do various things:

- expose the aforementioned parameters to `allel.ihs()`. Please could you check and advise on the defaults?
- implement maf filter before iHS `filter_min_maf` and with iHS `compute_min_maf`
- add standardisation by allele counts
- remove any NaNs prior to window size analysis 
- replace `window_func` with `window_percentiles`, a list of percentiles to report in windows if iHS values are to be summarised. and accommodate plotting multiple values with a bokeh.palette 


Here is a [colab](https://colab.research.google.com/drive/1Rc9sKDvq3YGxFzCzIpXqDY4VGXfgcxX_#scrollTo=97fdc59f) notebook with an example iHS analysis. The plots I'm getting look a little weird though compared to yours, I don't know if this is to do with the default settings that have been selected, or if I've done something slightly wrong. In particular, the 50th and 75th percentile values seem to drop around selective sweeps, rather than increase?

malariagen_data/anopheles.py
```diff
@@ -7726,6 +7727,550 @@ def plot_h1x_gwss(
...
+        if window_size:
+            # remove any NaNs prior to windowed percentiles
+            na_mask = np.isnan(ihs)
+            ihs = ihs[~na_mask]
+            pos = pos[~na_mask]
+
+            ihs = allel.moving_statistic(
+                ihs, statistic=np.percentile, size=window_size, q=window_percentiles
+            )
+            pos = allel.moving_statistic(pos, statistic=np.mean, size=window_size)
```
alimanfoo, line 7928: Couple of things here.

In the common case of default parameters, we'll need the allele counts twice, once for the minor allele frequency filtering, and again for the standardisation. Would be better to compute allele counts once up front, then re-use for both.

Also, would probably be better to remove nans before standardisation. Not sure it matters much, but might be a little more efficient if nothing else.

```suggestion

        # compute allele counts up front
        ac = ht.count_alleles(max_allele=1)
        
        # filter haplotypes by minor allele frequency
        if filter_min_maf > 0:
            af = ac.to_frequencies()
            # compute minor allele frequency, N.B., could be either ref or alt allele
            maf = np.min(af, axis=1)
            maf_filter = maf > filter_min_maf
            ht = ht.compress(maf_filter, axis=0)
            pos = pos[maf_filter]
            ac = ac[maf_filter]

        # compute IHS
        ihs = allel.ihs(
            h=ht,
            pos=pos,
            min_maf=compute_min_maf,
            min_ehh=min_ehh,
            include_edges=include_edges,
            max_gap=max_gap,
            gap_scale=gap_scale,
            use_threads=use_threads,
        )

        # remove any NaN values
        na_mask = ~np.isnan(ihs)
        ihs = ihs[na_mask]
        pos = pos[na_mask]
        ac = ac[na_mask]
        
        if standardize:
            # standardize IHS values
            ihs, _ = allel.standardize_by_allele_count(
                score=ihs,
                aac=ac[:, 1],
                bins=standardization_bins,
                n_bins=standardization_n_bins,
                diagnostics=standardization_diagnostics,
            )

        if window_size:
            # summarize IHS values in windows
            ihs = allel.moving_statistic(
                ihs, statistic=np.percentile, size=window_size, q=window_percentiles
            )
            pos = allel.moving_statistic(pos, statistic=np.mean, size=window_size)
```

alimanfoo:
> The plots I'm getting look a little weird though compared to yours, I don't know if this is to do with the default settings that have been selected, or if I've done something slightly wrong. In particular, the 50th and 75th percentile values seem to drop around selective sweeps, rather than increase?

Hahaha, this is fun...

image

Sorry forgot to mention, we should take the absolute value. I.e., as the last step before applying windowed computation, do something like:

```
ihs = np.fabs(ihs)
```

malariagen_data/anopheles.py
alimanfoo: Also minor allele frequency calculation needs to account for minor allele being either ref or alt.

alimanfoo:
Btw, interesting how IHS gives clear signal at Ace1...

image

Not sure if you'd see the same with H12 in this specific case, but I seem to recall this is one locus where generally IHS seems to pick up a signal better than H12.

sanjaynagi:
> 
> Sorry forgot to mention, we should take the absolute value. I.e., as the last step before applying windowed computation, do something like:
>

Hahaha well it at least it makes sense now!

Thanks @alimanfoo Ive pushed all the suggested changes except the `min_cohort_size` and `max_cohort_size`, will do that later.

sanjaynagi:
@alimanfoo do you want to do `min_cohort_size` and `max_cohort_size` in here? happy to but wont be able to get to it today.

alimanfoo:
> @alimanfoo do you want to do `min_cohort_size` and `max_cohort_size` in here? happy to but wont be able to get to it today.

No worries, yeah let's do `min_cohort_size` and `max_cohort_size` for IHS here. Can deal with H12 etc. later.

sanjaynagi:
OK, `min_cohort_size` and `max_cohort_size`, now implemented, hopefully it makes sense!

Definitely agree that we need both parameters for these gwss functions.

malariagen_data/anopheles.py
```diff
@@ -7726,6 +7727,591 @@ def plot_h1x_gwss(
...
+            ihs, _ = allel.standardize_by_allele_count(
+                score=ihs,
+                aac=ac[:, 1],
+                bins=standardization_bins,
+                n_bins=standardization_n_bins,
+                diagnostics=standardization_diagnostics,
+            )
+
+        if window_size:
+            # remove any NaNs prior to windowed percentiles
```
alimanfoo, line 7953:
```suggestion
```

comment doesn't apply any more

malariagen_data/anopheles.py
```diff
@@ -7726,6 +7727,591 @@ def plot_h1x_gwss(
...
+            self.results_cache_set(name=name, params=params, results=results)
+
+        x = results["x"]
+        ihs = results["ihs"]
+
+        return x, ihs
+
+    def _ihs_gwss(
+        self,
+        contig,
```
alimanfoo, line 7863:
```suggestion
    def _ihs_gwss(
        self,
        *,
        contig,
```

force use of keyword arguments, prevents bugs due to args in wrong order

alimanfoo (commented):
Thanks @sanjaynagi. Couple of tiny suggestions.

Also, could you add a notebook in the `notebooks` directory with one or two calls to `plot_ihs_gwss()` just to cover testing of the plotting functions.

Btw usually I clear outputs from those notebooks before committing to github, to avoid large notebook files in the repo. This is fine as the notebook mainly serves as a test and will be executed by CI.

sanjaynagi:
notebooks added which made me realise I hadn't added a cache name for af1!

alimanfoo (approved):
Awesome!!!

alimanfoo:
Merged :partying_face:

PR #364, opened 2023-03-16
Raise error when _snp_df() transcript is not found

### LINKED ISSUE #339: API gives cryptic error when providing unrecognised transcript to snp_allele_frequencies()
Reported by @jorels1 

When an unrecognised `transcript` is provided to the `snp_allele_frequencies` function, the error raised is rather cryptic for the end user, i.e. `TypeError: The region parameter must be a string or Region object.`

For example:
```
from malariagen_data import Af1
af1 = Af1()

# This works
snp_allele_freqs_df = af1.snp_allele_frequencies(
    transcript="LOC125761549_t5",
    cohorts="cohort_admin1_year",
    sample_sets="1231-VO-MULTI-WONDJI-VMF00043",
    drop_invariant=True,
)

# This breaks with
# TypeError: The region parameter must be a string or Region object.
snp_allele_freqs_df = af1.snp_allele_frequencies(
    transcript="AFUN011616-RB",
    cohorts="cohort_admin1_year",
    sample_sets="1231-VO-MULTI-WONDJI-VMF00043",
    drop_invariant=True,
)
```

leehart:
* Resolves #339.

### DIFF
--- a/malariagen_data/anopheles.py
+++ b/malariagen_data/anopheles.py
@@ -1897,6 +1897,10 @@ class AnophelesDataResource(ABC):
         debug("get feature direct from genome_features")
         gs = self.genome_features()
         feature = gs[gs["ID"] == transcript].squeeze()
+        if feature.empty:
+            raise ValueError(
+                f"No genome feature ID found matching transcript {transcript}"
+            )
         contig = feature.contig
         region = Region(contig, feature.start, feature.end)
 
--- a/tests/test_anopheles.py
+++ b/tests/test_anopheles.py
@@ -608,6 +608,36 @@ def test_snp_allele_frequencies__dup_samples(
 
 
 @pytest.mark.parametrize(
+    "subclass, transcript, sample_sets",
+    [
+        (
+            Ag3,
+            "foobar",
+            "3.0",
+        ),
+        (
+            Af1,
+            "foobar",
+            "1.0",
+        ),
+    ],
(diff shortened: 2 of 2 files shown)


alimanfoo (commented):
Thanks @leehart. Maybe add a test that ValueError is raised, rather than TypeError?

alimanfoo:
Hi @leehart, tests are failing here with `ipinfo.exceptions.RequestQuotaExceededError`, this will be related to the use of `ipinfo` to check the location where the code is being run, which is then used to raise a warning in colab if the user's VM is outside the US.

You could work around this by setting `check_location=False` whenever you instantiate either the `Ag3` or `Af1` classes.

alimanfoo:
> Hi @leehart, tests are failing here with `ipinfo.exceptions.RequestQuotaExceededError`, this will be related to the use of `ipinfo` to check the location where the code is being run, which is then used to raise a warning in colab if the user's VM is outside the US.
> 
> You could work around this by setting `check_location=False` whenever you instantiate either the `Ag3` or `Af1` classes.

Thinking about this some more, we should probably only be bothering with the location check if we know we are in colab. Otherwise there's not much point to the location check. I might raise a separate issue for that.

alimanfoo (approved):

alimanfoo:
Thanks @leehart :cherry_blossom:

PR #369, opened 2023-03-23
Implement G123 gwss method

### LINKED ISSUE #305: Implement G123 genome-wide selection scan
Implement functions for performing G123 genome-wide selection scans, analogous to functions available for H12.

sanjaynagi:
This PR adds support for the G123 genome-wide selection scan statistic for unphased diplotype data.  I have added a test for each `ag3` and `af1`, and a notebook.

To support G123 calculation, I have made a function `_diplotype_frequencies()` analogous to `_haplotype_frequencies()`. As with the `ihs` functions, I've added support for both a `min` and `max_cohort_size`. 

Main question for @alimanfoo, should I implement a segregating site filter? I think its probably needed, because at the minute we are basically windowing by bases rather than SNPs. 

Resolves #305.

### DIFF
--- a/malariagen_data/anopheles.py
+++ b/malariagen_data/anopheles.py
@@ -6690,6 +6690,14 @@ class AnophelesDataResource(ABC):
             self._cache_haplotype_sites[analysis] = root
         return root
 
+    def _haplotype_sites_for_contig(
+        self, *, contig, analysis, field, inline_array, chunks
+    ):
+        sites = self.open_haplotype_sites(analysis=analysis)
+        arr = sites[f"{contig}/variants/{field}"]
+        arr = da_from_zarr(arr, inline_array=inline_array, chunks=chunks)
+        return arr
+
     def _haplotypes_for_contig(
         self, *, contig, sample_set, analysis, inline_array, chunks
     ):
@@ -6698,6 +6706,9 @@ class AnophelesDataResource(ABC):
         debug("open zarr")
         root = self.open_haplotypes(sample_set=sample_set, analysis=analysis)
         sites = self.open_haplotype_sites(analysis=analysis)
+
+        debug("variant_position")
+        pos = sites[f"{contig}/variants/POS"]
 
         # some sample sets have no data for a given analysis, handle this
         # TODO consider returning a dataset with 0 length samples dimension instead, would
@@ -6708,8 +6719,6 @@ class AnophelesDataResource(ABC):
         coords = dict()
         data_vars = dict()
 
-        debug("variant_position")
-        pos = sites[f"{contig}/variants/POS"]
         coords["variant_position"] = (
             [DIM_VARIANT],
             da_from_zarr(pos, inline_array=inline_array, chunks=chunks),
@@ -8343,6 +8352,7 @@ class AnophelesDataResource(ABC):
         self,
         contig,
         window_size,
+        sites=DEFAULT,
         site_mask=DEFAULT,
         sample_sets=None,
         sample_query=None,
@@ -8358,6 +8368,12 @@ class AnophelesDataResource(ABC):
             Contig name (e.g., "2L" or "3RL")
         window_size : int
             The size of windows used to calculate g123 over.
+        sites: str, {'all', 'segregating', 'gamb_colu', 'gamb_colu_arab', 'arab'}
+            How to filter sites. 'all' includes all sites that pass
+            site filters, 'segregating' includes only segregating sites for
+            the given cohort, or a phasing analysis identifier can be
+            provided to use sites from the phased haplotype data, which is an
+            approximation to finding segregating sites in the entire ag3/af1 cohort.
         site_mask : str
             Which site filters mask to apply. See the `site_mask_ids`
             property for available values.
@@ -8389,9 +8405,15 @@ class AnophelesDataResource(ABC):
         # invalidate any previously cached data
         name = self._g123_gwss_cache_name
 
+        assert sites in self.site_mask_ids or sites in [
+            "all",
+            "segregating",
+        ], "sites parameter must be one of 'all', 'segregating', 'gamb_colu', 'gamb_colu_arab', 'arab'"
+
         params = dict(
             contig=contig,
-            site_mask=self._prep_site_mask_param(site_mask=site_mask),
+            sites=sites,
+            site_mask=site_mask,
             window_size=window_size,
             sample_sets=self._prep_sample_sets_param(sample_sets=sample_sets),
             # N.B., do not be tempted to convert this sample query into integer
@@ -8418,6 +8440,7 @@ class AnophelesDataResource(ABC):
     def _g123_gwss(
         self,
         contig,
+        sites,
         site_mask,
         window_size,
         sample_sets,
@@ -8426,11 +8449,12 @@ class AnophelesDataResource(ABC):
         max_cohort_size,
         random_seed,
     ):
+        debug = self._log.debug
         ds_snps = self.snp_calls(
             region=contig,
-            site_mask=site_mask,
             sample_query=sample_query,
             sample_sets=sample_sets,
+            site_mask=site_mask,
             random_seed=random_seed,
         )
 
@@ -8451,12 +8475,30 @@ class AnophelesDataResource(ABC):
                 ds_snps = ds_snps.isel(samples=loc_downsample)
 
         gt = allel.GenotypeDaskArray(ds_snps["call_genotype"].data)
-        with self._dask_progress(desc="Load Genotypes"):
+        with self._dask_progress(desc="Load genotypes"):
             gt = gt.compute()
         pos = ds_snps["variant_position"].values
 
-        g123 = allel.moving_statistic(gt, statistic=self._garud_g123, size=window_size)
+        if sites in self.site_mask_ids:
+            debug("subsetting to haplotype positions")
+            haplotype_pos = self._haplotype_sites_for_contig(
+                contig=contig,
+                analysis=sites,
+                field="POS",
+                inline_array=True,
+                chunks="native",
+            )
+            hap_site_mask = np.in1d(pos, haplotype_pos.values)
+            pos = pos[hap_site_mask]
+            gt = gt.compress(hap_site_mask, axis=0)
+        elif sites == "segregating":
+            debug("subsetting to segregating sites")
+            ac = gt.count_alleles(max_allele=3)
+            seg = ac.is_segregating()
+            pos = pos[seg]
+            gt = gt.compress(seg, axis=0)
 
+        g123 = allel.moving_statistic(gt, statistic=self._garud_g123, size=window_size)
         x = allel.moving_statistic(pos, statistic=np.mean, size=window_size)
 
         results = dict(x=x, g123=g123)
@@ -8467,6 +8509,7 @@ class AnophelesDataResource(ABC):
         self,
         contig,
         window_size,
+        sites,
         site_mask=DEFAULT,
         sample_sets=None,
         sample_query=None,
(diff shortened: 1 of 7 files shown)


alimanfoo:
> Main question for @alimanfoo, should I implement a segregating site filter? I think its probably needed, because at the minute we are basically windowing by bases rather than SNPs

Ideally we'd want some way to calibrate a sensible window size for g123, analogous to h12.

If we window on bases (sites) then the windowing will be unaffected by any genomic variations in diversity within the cohort you're analysing, which is good. But you will get spikes around centromeres because of low diversity generally. Also computation may be slower because you're hashing much more data (lots of nonseg sites).

If we window on segregating sites within the cohort of interest, computation should be more efficient and less prone to peaks in centromeres. But also may reduce sensitivity to detect true selection signals, because low diversity around a signal will cause windows to get stretched. 

Ideally you want to ascertain segregating sites in some much larger cohort first, or in some other cohort where you think it's unlikely they're affected by recent selection. That's effectively what happens with h12, because it's using the haplotypes, and the haplotypes are phased at biallelic sites that we're segregating in Ag3.0.

Hm, tricky...

alimanfoo:
One option could be to subset the genotype calls to the same sites that we used for haplotype phasing. This would effectively be using Ag3.0 as a whole to ascertain segregating sites. If we did go down that route, we'd probably want to expose the fact that actually there are three sets of sites to choose from: gamb_colu_arab, gamb_colu and arab.

alimanfoo:
Possibly could do something like expose a `sites` parameter to allow different approaches. If value is "all" then use all sites passing site filters, regardless of whether there is any variation. If value is "segregating" then ascertain segregating sites within the cohort that pass site filters. If value is "gamb_colu_arab" or "gamb_colu" or "arab" then use the corresponding sites used for haplotype phasing.

sanjaynagi:
I like that idea!

alimanfoo:
> I like the idea of using the haplotype sites! and yes will be nice to have the different approaches!

Cool, let's go with that then. Give me a shout if need any help subsetting genotypes to the haplotype positions.

Maybe we need to add some window size calibration functions too, analogous to H12.

sanjaynagi:
Ok, Ive now added support for a `sites` parameter which can be either 'all', 'segregating', or one of the phasing analyses. To support the haplotype site retrieval, I've added a method in `anopheles` `_haplotype_sites_for_contig()` which retrieves haplotype sites for a given contig, with a method of the same name in `ag3` which handles joined arms. 

Ive also added a method for G123 calibration which is analogous to H12, and added some tests for that and added it to the notebook. 

A couple of questions:
- There is some code duplication in the gwss and calibration functions, for the code which subsets to either the haplotype positions or finds segregating sites. Is this OK or should I be making a new function which does this?
- The 'all' parameter does not actually do anything (apart from the fact I've added an assert statement into `g123_gwss` to ensure the `sites` parameter is something we expect. Sometimes I run into situations like this, where one of the parameter options is kind of useless, ie, if it wasn't for the assert statement, we could put `sites='gobbledegook'` and it would still run just like I put 'all'. It always feels weird, is there a good approach for handling situations like this? thanks!

or at least, this is what I want to push to the branch, but git is giving me 
`! [remote rejected] g123-gwss-23-03-23 -> g123-gwss-23-03-23 (failure)
error: failed to push some refs to 'git@github.com:sanjaynagi/malariagen-data-python.git'`, possibly due to the github ssh keys changing, trying to figure out.

malariagen_data/anopheles.py
```diff
@@ -8337,35 +8351,52 @@ def plot_ihs_gwss(
 
         bkplt.show(fig)
 
-    def plot_haplotype_clustering(
+    def _garud_g123(self, gt):
+        """Compute Garud's G1, G12, G123, G2/G1."""
+        f = _diplotype_frequencies(gt)
+        # convert dict to array of sorted frequencies
+        f = np.sort(np.fromiter(f.values(), dtype=float))[::-1]
```
alimanfoo, line 8358: This seems overly complex...?

malariagen_data/anopheles.py
```diff
@@ -8337,35 +8351,52 @@ def plot_ihs_gwss(
...
+        contig: str
+            Contig name (e.g., "2L" or "3RL")
+        window_size : int
+            The size of windows used to calculate g123 over.
+        sites: str, {'all', 'segregating', 'gamb_colu', 'gamb_colu_arab', 'arab'}
+            How to filter sites. 'all' includes all sites that pass
+            site filters, 'segregating' includes only segregating sites for
+            the given cohort, or a phasing analysis identifier can be
+            provided to use sites from the phased haplotype data, which is an
+            approximation to finding segregating sites in the entire ag3/af1 cohort.
```
alimanfoo, line 8397:
```suggestion
        sites: str
            How to filter sites. 'all' includes all sites that pass
            site filters, 'segregating' includes only segregating sites for
            the given cohort, or a phasing analysis identifier can be
            provided to use sites from the haplotype data, which is an
            approximation to finding segregating sites in the entire Ag3.0 
            (gambiae complex) or Af1.0 (funestus) cohort.
```

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
-        from .plotly_dendrogram import create_dendrogram
+        # change this name if you ever change the behaviour of this function, to
+        # invalidate any previously cached data
+        name = self._g123_gwss_cache_name
 
-        debug = self._log.debug
+        assert sites in self.site_mask_ids or sites in [
+            "all",
+            "segregating",
+        ], "sites parameter must be one of 'all', 'segregating', 'gamb_colu', 'gamb_colu_arab', 'arab'"
```
alimanfoo, line 8432:
```suggestion
        if sites == DEFAULT:
            sites = self._default_phasing_analysis
        valid_sites = self.phasing_analysis_ids + ("all", "segregating")
        if sites not in valid_sites:
            raise ValueError(f"Invalid value for `sites` parameter, must be one of {valid_sites}.")
```

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+                )
+        if max_cohort_size is not None:
+            n_samples = ds_snps.dims["samples"]
+            if n_samples > max_cohort_size:
+                rng = np.random.default_rng(seed=random_seed)
+                loc_downsample = rng.choice(
+                    n_samples, size=max_cohort_size, replace=False
+                )
+                loc_downsample.sort()
+                ds_snps = ds_snps.isel(samples=loc_downsample)
```
alimanfoo, line 8496: Suggest to push this logic up into the snp_calls() function (which already has support for the cohort_size parameters).

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
-        fig.update_xaxes(mirror=False, showgrid=True, showticklabels=False, ticks="")
-        fig.update_yaxes(
-            mirror=False, showgrid=True, showline=True, range=[-2, max_dist + 1]
-        )
+        gt = allel.GenotypeDaskArray(ds_snps["call_genotype"].data)
+        with self._dask_progress(desc="Load genotypes"):
+            gt = gt.compute()
+        pos = ds_snps["variant_position"].values
+
+        if sites in self.site_mask_ids:
```
alimanfoo, line 8503:
```suggestion
        if sites in self.phasing_analysis_ids:
```

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+            debug("subsetting to haplotype positions")
+            haplotype_pos = self._haplotype_sites_for_contig(
+                contig=contig,
+                analysis=sites,
+                field="POS",
+                inline_array=True,
+                chunks="native",
+            ).compute()
+            hap_site_mask = np.in1d(pos, haplotype_pos)
```
alimanfoo, line 8512:
```suggestion
            hap_site_mask = np.in1d(pos, haplotype_pos, assume_unique=True)
```

Apparently can speed up the computation.

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+        contig: str
+            Contig name (e.g., "2L" or "3RL")
+        window_size : int
+            The size of windows used to calculate g123 over.
+        sites: str, {'all', 'segregating', 'gamb_colu', 'gamb_colu_arab', 'arab'}
+            How to filter sites. 'all' includes all sites that pass
+            site filters, 'segregating' includes only segregating sites for
+            the given cohort, or a phasing analysis identifier can be
+            provided to use sites from the phased haplotype data, which is an
+            approximation to finding segregating sites in the entire ag3/af1 cohort.
```
alimanfoo, line 8560: See comment above.

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
```
alimanfoo, line 8687: See comment above.

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+            merge_tools=True,
+            sizing_mode=sizing_mode,
+        )
+
+        bkplt.show(fig)
+
+    def g123_calibration(
+        self,
+        contig,
+        sites,
```
alimanfoo, line 8770:
```suggestion
        sites=DEFAULT,
```

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+        Parameters
+        ----------
+        contig: str
+            Contig name (e.g., "2L" or "3RL")
+        sites: str, {'all', 'segregating', 'gamb_colu', 'gamb_colu_arab', 'arab'}
+            How to filter sites. 'all' includes all sites that pass
+            site filters, 'segregating' includes only segregating sites for
+            the given cohort, or a phasing analysis identifier can be
+            provided to use sites from the phased haplotype data, which is an
+            approximation to finding segregating sites in the entire ag3/af1 cohort.
```
alimanfoo, line 8790: See comment above.

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+        site_mask : str
+            Which site_mask to use. See the `phasing_site_mask_ids`
```
alimanfoo, line 8792:
```suggestion
            Which site_mask to use. See the `site_mask_ids`
```

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+        -------
+        calibration runs : list of numpy.ndarray
+            A list of g123 calibration run arrays for each window size, containing
+            values and percentiles.
+
+        """
+
+        # change this name if you ever change the behaviour of this function, to
+        # invalidate any previously cached data
+        name = self._g123_calibration_cache_name
```
alimanfoo, line 8823: Is this defined?

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+                )
+        if max_cohort_size is not None:
+            n_samples = ds_snps.dims["samples"]
+            if n_samples > max_cohort_size:
+                rng = np.random.default_rng(seed=random_seed)
+                loc_downsample = rng.choice(
+                    n_samples, size=max_cohort_size, replace=False
+                )
+                loc_downsample.sort()
+                ds_snps = ds_snps.isel(samples=loc_downsample)
```
alimanfoo, line 8884: If push this up into snp_calls() can avoid the duplication here.

malariagen_data/anopheles.py
```diff
@@ -8374,190 +8405,852 @@ def plot_haplotype_clustering(
...
+                inline_array=True,
+                chunks="native",
+            ).compute()
+            hap_site_mask = np.in1d(pos, haplotype_pos)
+            gt = gt.compress(hap_site_mask, axis=0)
+        elif sites == "segregating":
+            debug("subsetting to segregating sites")
+            ac = gt.count_alleles(max_allele=3)
+            seg = ac.is_segregating()
+            gt = gt.compress(seg, axis=0)
```
alimanfoo, line 8906: This is annoying to have to duplicate. Maybe factor out into a common private method, e.g., `_load_gt_for_g123(contig, sites, site_mask)`.

alimanfoo (commented):
Awesome :cherry_blossom:  

A few small suggestions...

malariagen_data/anopheles.py
alimanfoo: Oh maybe not, forgot the frequencies are coming from a Counter.

alimanfoo (approved):
Looks good, thanks @sanjaynagi! :bow:

PR #377, opened 2023-03-29
Drop support for Python 3.7

### LINKED ISSUE #365: Drop support for PY37
Almost all our dependencies have dropped support for PY37 for quite a while. The only thing holding us back following suit was colab, but colab has also upgraded (now on PY39). 

So let's drop support for PY37.

alimanfoo:
* Resolves #365
* Upgrades poetry version to 1.4.1 in CI


alimanfoo:
Merging as hopefully uncontroversial, but happy to address comments or suggestions in a follow-up PR.

Issue #366, opened 2023-03-18
Anopheles refactor


alimanfoo:
The `AnophelesDataResource` class is becoming a [God class](https://medium.com/@carlos.ariel.mamani/the-god-object-or-god-class-anti-pattern-bfb8c15eb513), there are a large number of methods covering different areas of functionality, and it is becoming unwieldy and hard to navigate and maintain.

Propose to refactor this somehow, compartmentalising different areas of functionality into separate classes.

alimanfoo:
We could do this using something along the lines of the mixin pattern (e.g., [here](https://dev.to/bikramjeetsingh/write-composable-reusable-python-classes-using-mixins-6lj)). 

E.g., we could have classes like `AnophelesBase`, `AnophelesMetadata`, `AnophelesSnpData`, `AnophelesCnvData`, `AnophelesHaplotypeData`, `AnophelesPca`, `AnophelesH12`, etc., each in their own module. The `AnophelesDataResource` class could then end up with the same API as now, but would be constructed like:

```python
class AnophelesDataResource(
    AnophelesPca,
    AnophelesH12,
    AnophelesHaplotypeData,
    AnophelesSnpData,
    AnophelesCnvData,
    AnophelesMetadata,
    AnophelesBase):
    ....
```

Potential complexity here is that some of these mixin-style classes would depend on others. E.g., the `AnophelesH12` class would depend on the `AnophelesHaplotypeData` class. But should be able to handle that if careful about order of superclasses.

alimanfoo:
Current slowest tests, to aid prioritisation:

![image](https://github.com/malariagen/malariagen-data-python/assets/703554/be7978d6-db08-4598-b9b6-d84f7f5ac4dd)

AswaniSahoo:
Hi @jonbrenas, I'd like to take on the AnophelesHapNetAnalysis extraction.

Plan:
- Create `anoph/hapnet.py` with AnophelesHapNetAnalysis class
- Move `plot_haplotype_network()` and its internal helpers from `anopheles.py`
- Keep the existing `hapnet_params` module as-is (it already exists)
- Add simulated-data unit tests in `tests/anoph/test_hapnet.py`
- Follow the same pattern as the existing `hapclust.py` and `heterozygosity.py` 
  extractions

This function uses Dash/Cytoscape for network visualization, so the test 
strategy will focus on the data pipeline (MJN computation, node/edge 
construction) rather than the Dash rendering layer.

[...]

PR #380, opened 2023-03-31
Anopheles refactor part 1 - AnophelesBase

### LINKED ISSUE #368: Only check client location if we're in colab
Currently we check the client location by default, using the `ipinfo` library. However, we only make use of this information if the client is running on colab, where we raise a warning if the client is outside the US. 

It would probably be better to check first whether we're running on colab, and only then make a location check.

alimanfoo:
This PR begins the work of breaking up the `AnophelesDataResource` class into a number of smaller classes with related and self-contained functionality, as proposed in #366.

Here a new class `AnophelesBase` is created, which assumes responsibility for loading release configuration and accessing release manifests (sample sets).

Note also that some new tests are added which specifically test this new class. These new tests use locally-created data which mimics the organisation and format of data in Ag3 and Af1 but is local and smaller to allow for faster test runs.

Also resolves #368 along the way.

### DIFF
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -10,32 +10,28 @@ jobs:
     notebooks:
         strategy:
             fail-fast: true
-            matrix:
-                python-version: ['3.10']
-                poetry-version: ['1.4.1']
-                os: [ubuntu-latest]
-        runs-on: ${{ matrix.os }}
+        runs-on: ubuntu-latest
         steps:
             - name: Checkout source
               uses: actions/checkout@v3
+
+            - name: Setup python
+              uses: actions/setup-python@v4
+              with:
+                  python-version: '3.10'
+                  cache: 'poetry'
+
+            - name: Install poetry
+              run: pipx install poetry==1.4.1
+
+            - name: Install dependencies
+              run: poetry install
 
             - name: Restore GCS cache
               uses: actions/cache/restore@v3
               with:
                   path: gcs_cache
                   key: gcs_cache_notebooks
-
-            - name: Install poetry
-              run: pipx install poetry==${{ matrix.poetry-version }}
-
-            - name: Setup python
-              uses: actions/setup-python@v4
-              with:
-                  python-version: ${{ matrix.python-version }}
-                  cache: 'poetry'
-
-            - name: Install dependencies
-              run: poetry install
 
             - name: Run notebooks
               run: poetry run jupyter nbconvert --execute notebooks/* --inplace
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -15,14 +15,14 @@ jobs:
             - name: Checkout source
               uses: actions/checkout@v3
 
+            - name: Install poetry
+              run: pipx install poetry==1.4.1
+
             - name: Setup python
               uses: actions/setup-python@v4
               with:
                   python-version: '3.10'
                   cache: 'poetry'
-
-            - name: Install poetry
-              run: pipx install poetry==1.4.1
 
             - name: Install dependencies
               run: poetry install
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,12 +19,6 @@ jobs:
             - name: Checkout source
               uses: actions/checkout@v3
 
-            - name: Restore GCS cache
-              uses: actions/cache/restore@v3
-              with:
-                  path: gcs_cache
-                  key: gcs_cache_tests
-
             - name: Install poetry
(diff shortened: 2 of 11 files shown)


malariagen_data/af1.py
```diff
@@ -134,7 +132,7 @@ def __init__(
         cohorts_analysis=None,
         site_filters_analysis=None,
         pre=False,
-        **kwargs,  # used by simplecache, init_filesystem(url, **kwargs)
+        **storage_kwargs,  # used by fsspec via init_filesystem(url, **kwargs)
```
alimanfoo, line 135: Rename for clarity.

malariagen_data/af1.py
```diff
@@ -148,7 +146,10 @@ def __init__(
             show_progress=show_progress,
             check_location=check_location,
             pre=pre,
-            **kwargs,  # used by simplecache, init_filesystem(url, **kwargs)
+            gcs_url=GCS_URL,
+            major_version_number=MAJOR_VERSION_INT,
+            major_version_path=MAJOR_VERSION_GCS_STR,
```
alimanfoo, line 151: Making these parameters to the super-class constructor makes testing much easier.

tests/test_anopheles.py
```diff
@@ -58,9 +58,9 @@ def test_sample_sets(subclass, url, release, sample_sets_count):
     assert len(df_sample_sets) == sample_sets_count
     assert tuple(df_sample_sets.columns) == ("sample_set", "sample_count", "release")
 
-    # test duplicates not allowed
-    with pytest.raises(ValueError):
-        anoph.sample_sets(release=[release, release])
+    # test duplicates are handled
+    df_dup = anoph.sample_sets(release=[release, release])
+    assert_frame_equal(df_sample_sets, df_dup)
```
alimanfoo, line 63: Slight change of behaviour here, seemed more friendly just to handle and dups.

.github/workflows/notebooks.yml
```diff
@@ -10,33 +10,29 @@ jobs:
     notebooks:
         strategy:
             fail-fast: true
-            matrix:
-                python-version: ['3.10']
-                poetry-version: ['1.4.1']
-                os: [ubuntu-latest]
-        runs-on: ${{ matrix.os }}
```
alimanfoo, line 17: Taking the opportunity to simplify the action, we don't actually need a matrix here.

.github/workflows/tests.yml
```diff
@@ -12,21 +12,13 @@ jobs:
             fail-fast: true
             matrix:
                 python-version: ['3.8', '3.9', '3.10']
-                poetry-version: ['1.4.1']
-                os: [ubuntu-latest]
-        runs-on: ${{ matrix.os }}
```
alimanfoo, line 17: Taking the opportunity to simplify the action, keeping only the parameters which are needed in the matrix.

README.md
```diff
@@ -45,31 +45,31 @@ Fork and clone this repo:
...
 ```
 
 Create development environment:
 
 ```bash
 cd malariagen-data-python
-poetry use 3.10
+poetry use 3.8
 poetry install
 ```
```
alimanfoo, line 74: Recommending Python 3.8 for local development, because otherwise if we use a later Python we can accidentally try to use features not supported in Python 3.8, which doesn't then crop up until CI fails. Better to fail fast locally.

.github/workflows/tests.yml
```diff
@@ -37,8 +29,18 @@ jobs:
             - name: Install dependencies
               run: poetry install
 
-            - name: Run tests
-              run: poetry run pytest -v
+            # Run a subset of tests first which run quickly to fail fast.
+            - name: Run fast unit tests
+              run: poetry run pytest -v tests/anoph
```
alimanfoo, line 34: Pulling out a subset of tests here which are guaranteed to run fast, so we fail faster if there are any obvious bugs.

alimanfoo:
Keen to push ahead here so planning to merge soon if no objections, but please shout if you'd like more time to review.

alimanfoo:
Thanks @leehart :pray:

sanjaynagi:
Sent from my iPhoneOn 31 Mar 2023, at 17:06, Alistair Miles ***@***.***> wrote:﻿
Keen to push ahead here so planning to merge soon if no objections, but please shout if you'd like more time to review.

sanjaynagi:
Sorry was at the pub Sent from my iPhoneOn 31 Mar 2023, at 17:06, Alistair Miles ***@***.***> wrote:﻿
Keen to push ahead here so planning to merge soon if no objections, but please shout if you'd like more time to review.

PR #385, opened 2023-04-14
Add quarter cohorts in `read_cohort_metadata()` exceptions

### LINKED ISSUE #384: Update exceptions for cohorts in `read_cohort_metadata()`
When reading the sample metadata during pre-release checks on the API (with `ag3.sample_metadata()`, we compare the sample metadata columns in the pre-release against the ones that already have been released. At this stage, we haven't performed cohort analysis, so all columns for cohorts are empty, and manually specified with an exception in `read_cohort_metadata()`. 

During ag3.7, two new cohorts were added `cohort_admin1_quarter` and `cohort_admin2_quarter`; which need to be included in the exception to avoid assertion errors during the pre-release checks in ag3.8 (and going forward) (see [error](https://github.com/malariagen/vector-ops/blob/171e724f081a334685c7980d59db1608f8b72691/tracking/release/v3.8/check_staged_API.ipynb)).

ahernank:
Resolves #384

Added the two new cohorts to the exceptions in `read_cohort_metadata()`. 

As this is in `anopheles.py`, will have to keep in mind that we would expect cohort analysis for quarters in af1 too.

### DIFF
--- a/tests/anoph/conftest.py
+++ b/tests/anoph/conftest.py
@@ -153,10 +153,10 @@ class Gff3Simulator:
                 gene_start = cur_fwd + inter_size
             else:
                 gene_start = cur_rev + inter_size
-            if gene_start >= contig_size:
+            gene_end = gene_start + gene_size
+            if gene_end >= contig_size:
                 # Bail out, no more space left on the contig.
                 return
-            gene_end = min(gene_start + gene_size, contig_size)
             assert gene_end > gene_start
             gene_attrs = f"ID={gene_id}"
             for attr in self.attrs:
@@ -206,7 +206,6 @@ class Gff3Simulator:
         # the same coordinates as the parent gene, which is not strictly
         # accurate in real data.
 
-        gene_size = gene_end - gene_start
         for transcript_ix in range(
             random.randint(self.n_transcripts_low, self.n_transcripts_high)
         ):
@@ -231,7 +230,6 @@ class Gff3Simulator:
                 contig=contig,
                 strand=strand,
                 gene_ix=gene_ix,
-                gene_size=gene_size,
                 transcript_ix=transcript_ix,
                 transcript_id=transcript_id,
                 transcript_start=transcript_start,
@@ -244,7 +242,6 @@ class Gff3Simulator:
         contig,
         strand,
         gene_ix,
-        gene_size,
         transcript_ix,
         transcript_id,
         transcript_start,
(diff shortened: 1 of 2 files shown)


ahernank:
cc @leehart

alimanfoo:
Hi @ahernank, we'll need to figure out how to handle different analysis versions here, as some will include the additional columns and some won't currently. 

The Ag3 20230223 cohorts analysis (current default) includes these columns, but previous Ag3 cohorts analyses do not, and the Af1 20221129 cohorts analysis (current default) does not either.

Lots of ways we could hack this, but not sure what to recommend, thinking...

alimanfoo:
OK, this isn't perfect but here's a suggestion for how to hack this for the time being.

We could deal with this based on date, assuming that all cohorts analysis prior to 20230223 do not include the quarter columns, and all cohorts analyses from this date onwards do (or will) include these columns. This would then work for both Ag3 and Af1, assuming that we add these columns into the next Af1 cohorts analysis.

If we take this approach, then we could implement this check using a comparison with the analysis version, which is a date string in "YYYYMMDD" format and so less than and greater than comparisons will work correctly.

I.e., the code could be something like:

```python
            except Exception:
                if self._cohorts_analysis < "20230223":
                    # cohort_cols without quarter
                    ...
                else:
                    # cohort_cols with quarter
                    ...
```

alimanfoo:
Btw it would ultimately be better if the cohorts analyses were self-describing, in the sense that each analysis version somehow declared what columns it uses, e.g., via some kind of configuration file or metadata catalog published alongside the data. Then the API wouldn't need to have any baked-in knowledge of what columns to expect, and each cohorts analysis would have the flexibility to add or remove columns. But that's probably a bridge too far for this PR, something to consider for later.

alimanfoo:
Not sure why tests are failing here, unrelated to changes in this PR, investigating...

tests/anoph/conftest.py
```diff
@@ -153,10 +153,10 @@ def simulate_genes(self, *, contig, contig_size):
                 gene_start = cur_fwd + inter_size
             else:
                 gene_start = cur_rev + inter_size
-            if gene_start >= contig_size:
+            gene_end = gene_start + gene_size
+            if gene_end >= contig_size:
```
alimanfoo, line 157: This change ensures simulated genes will always be within the requested size range, not truncated by the end of the contig. This in turn fixes the intermittent issues arising with simulating exons.

alimanfoo (approved):

alimanfoo:
Hi @ahernank, this LGTM, are you happy for me to merge if CI passes?

ahernank:
Thanks, @alimanfoo! Yes, please.

PR #386, opened 2023-04-19
Anopheles refactor part 4 - AnophelesSampleMetadata

### LINKED ISSUE #341: Asynchronously load sample metadata
It doesn't take too long to load sample metadata, but it is a bit annoying, especially as the number of sample sets grows. The time is all network latency, because lots of relatively small CSV files are being retrieved. It would be possible to load these asynchronously, which would probably speed things up by an order of magnitude.

alimanfoo:
Here we continue towards #366 by pulling out functions for accessing sample metadata into a new class `AnophelesSampleMetadata`.

Also we resolve #341 along the way, by making use of the ability of `fsspec.cat()` to read multiple files concurrently. Because reading metadata is now faster, we change the caching strategy a little, and we do not report a progress bar while loading metadata.

Also we take this opportunity to remove support for older legacy AIM and species analyses, and to rename some of the supporting functions for clarity. This should not affect user code, as the main user-facing functions retain the same names and parameters.


malariagen_data/anoph/base.py
```diff
@@ -223,10 +239,38 @@ def __init__(
...
+        # from GCS can be slow if files are not read concurrently. Hence we
+        # want to make use of cat() here and provide paths for all files to
+        # read concurrently. For more information see:
+        # https://filesystem-spec.readthedocs.io/en/latest/async.html
+        full_files = self._fs.cat(full_paths, on_error=on_error)
+
+        # Strip off the prefix.
+        files = {k.split(prefix, 1)[1]: v for k, v in full_files.items()}
+
+        return files
```
alimanfoo, line 272: This new function allows us to read files concurrently.

alimanfoo:
There is probably more test refactoring that could be done here, moving across some testing logic which currently runs against the real data, and running it on the locally simulated data instead. But I might leave that for a future PR.

alimanfoo:
A note about the testing strategy. Here we continue with the strategy of simulating a smaller data resource with local files, to accelerate the unit tests. Rather than simulate sample metadata from scratch, I have pulled in a small selection of real sample metadata CSV files, and then we sub-sample from those to generate the simulated data resource.

PR #393, opened 2023-05-03
Anopheles refactor part 5 - AnophelesSnpData

### LINKED ISSUE #358: Consider validating function arguments with pydantic
This looks useful...

https://docs.pydantic.dev/usage/validation_decorator/

### LINKED ISSUE #396: Speed up install on Python 3.10
Currently install in Python 3.10 is slow because scikit-allel doesn't have wheels for this Python version. This now affects colab which has upgraded to Python 3.10.

alimanfoo:
Here we continue working towards #366 by pulling out functions for accessing SNP data into a new `AnophelesSnpData` composable class.

* [x] open_snp_sites()
* [x] open_snp_genotypes()
* [x] open_site_filters()
* [x] site_filters()
* [x] snp_sites()
* [x] snp_genotypes()
* [x] snp_variants()
* [x] snp_calls()
* [x] open_site_annotations()
* [x] site_annotations()
* [x] is_accessible()
* [x] snp_allele_counts()
* [x] plot_snps_track()
* [x] plot_snps()
* [x] fast unit tests on simulated data, coverage up

While we are here, we also introduce the use of the `typeguard` package. A `check_types` decorator is added to public methods to provide runtime checking of parameter types, which then removes the need for us to write any type checking code. This resolves #358 (preferring typeguard over pydantic because the type checking behaviour is simpler and there is no coersion of types).

Along the way, this PR finds and fixes a number of typing errors.

Also resolves #396 by upgrading scikit-allel.


malariagen_data/ag3.py
```diff
@@ -13,13 +14,17 @@
...
+from .anoph.base import base_params
 from .anopheles import AnophelesDataResource, gplt_params
 from .util import (
     DIM_SAMPLE,
     DIM_VARIANT,
     Region,
     da_from_zarr,
     init_zarr_store,
+    parse_multi_region,
+    parse_single_region,
```
alimanfoo, line 26: The functions for parsing `region` parameter have been reworked to make sure it's clear when a function will accept multiple regions in a list, versus when only a single region is accepted.

malariagen_data/anoph/base.py
```diff
@@ -33,56 +47,97 @@ class base_params:
...
+
+    sample_indices: TypeAlias = Annotated[
+        List[int],
+        """
+        Advanced usage parameter. A list of indices of samples to select,
+        corresponding to the order in which the samples are found within the
+        sample metadata. Either provide this parameter or sample_query, not
+        both.
+        """,
+    ]
```
alimanfoo, line 104: This parameter has been introduced to make typing easier when handling the case where samples are being selected via a list of indices rather than a pandas query.

malariagen_data/anoph/base.py
```diff
@@ -489,11 +580,6 @@ def _prep_sample_sets_param(
                 # Make sure we end up with a flat list of sample sets.
                 prepped_sample_sets.extend(sp)
 
-        else:
-            raise TypeError(
-                f"Invalid type for sample_sets parameter; expected str, list or tuple; found: {sample_sets!r}"
-            )
-
```
alimanfoo, line 496: Handling type errors isn't required any more, because the `@check_types` decorator ensures that function parameters will conform to the type annotations.

malariagen_data/anoph/sample_metadata.py
```diff
@@ -507,18 +521,19 @@ def sample_metadata(
...
-            else:
-                # Assume it is an indexer.
-                df_samples = df_samples.iloc[sample_query]
+            # Assume a pandas query string.
+            df_samples = df_samples.query(sample_query)
+            df_samples = df_samples.reset_index(drop=True)
+        elif sample_indices is not None:
+            # Assume it is an indexer.
+            df_samples = df_samples.iloc[sample_indices]
             df_samples = df_samples.reset_index(drop=True)
```
alimanfoo, line 532: Example of how we now use two parameters `sample_query` and `sample_indices` to more clearly handle different types of sample selection.

malariagen_data/util.py
```diff
@@ -839,3 +884,43 @@ def check_colab_location(*, gcs_url: str, url: str) -> Optional[ipinfo.details.D
             pass
 
     return details
+
+
+def check_types(f):
```
alimanfoo, line 889: N.B., we roll our own simple decorator here for performing runtime type checks on parameter values. See the docstring for explanation of why we don't use typeguard's `@typechecked` decorator.

notebooks/repr.ipynb
alimanfoo, line 1: Clear outputs.

tests/test_af1.py
alimanfoo, line 1: All test functions have been migrated to use simulated data where possible, to speed up the test suite.

tests/test_ag3.py
alimanfoo, line 1: All test functions have been migrated to use simulated data where possible, to speed up the test suite.

tests/test_anopheles.py
alimanfoo, line 1: All test functions have been migrated to use simulated data where possible, to speed up the test suite.

alimanfoo (commented):
A few notes on the changes in this PR...

alimanfoo:
Going to merge if CI passes, but happy to follow up any comments. Need to cut a release as the xarray concatenation problem is causing big slowdowns for trying to run large PCAs, we need a release with the fix in #395.

alimanfoo:
> Going to merge if CI passes, but happy to follow up any comments. Need to cut a release as the xarray concatenation problem is causing big slowdowns for trying to run large PCAs, we need a release with the fix in #395.

Doh, actually forgot we already had a release with #395 in it. Ah well, probably good to get these changes in sooner.

PR #401, opened 2023-05-20
Anopheles refactor part 6 - haplotype data

alimanfoo:
Continue working towards #366, refactoring haplotype data access functions.

Also some other changes:

* All parameter definitions have been moved to separate dedicated modules, to make it easier to find them. E.g., the base_params are now available via `from malariagen_data.anoph import base_params`.
* Haplotype data access tests have been refactored to use simulated data rather than real data for speed.
* Some other general improvements to the fast unit tests for speed and simplicity.


alimanfoo:
Hi @leehart, FYI more spring cleaning coming in here.

PR #406, opened 2023-06-01
Use webgl for bokeh genome plots

alimanfoo:
An issue we are experiencing in the selection atlas project is that multiple GWSS plots can cause problems in the browser, like long page load times and unresponsiveness (xref https://github.com/anopheles-genomic-surveillance/selection-atlas/issues/51).

Here we switch all bokeh genome plots to use "webgl" instead of the default "canvas" for the `output_backend` parameter. Webgl in general renders plots faster and can handle plots with more points than canvas. See the [bokeh docs on webgl acceleration](https://docs.bokeh.org/en/latest/docs/user_guide/output/webgl.html#webgl-acceleration) for more info.

In addition to faster and more responsive plots, this also removes the downsampling of glyphs that otherwise happens with the canvas backend when zooming in or out, which is nice.

Because the webgl plots have a very slightly different visual appearance, this PR also includes some tweaks to the visual style of plots to make things clearer and more uniform between backends. In particular, it seems that glyphs can appear a little blurry with webgl if the `line_width` parameter is a fractional value, so we set it to 0 or 1 here throughout.

There are also a couple of tweaks to the genes track and RoH track plots to make them more compact.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -1741,6 +1741,7 @@ class Ag3(AnophelesDataResource):
         line_kwargs=None,
         show=True,
         x_range=None,
+        output_backend="webgl",
     ):
         """Plot CNV HMM data for a single sample, using bokeh.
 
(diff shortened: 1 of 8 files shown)


alimanfoo:
Hi @sanjaynagi, FYI, trying this to see if it at least partially solves our problem with lots of GWSS plots on the same page.

sanjaynagi (approved):
Very nice, that's ideal to have it in malariagen_data (although guess we may not have had a choice actually). Looks good.

alimanfoo:
Thanks @sanjaynagi, yeah had to plumb it in up here in malariagen_data so we can try it out in selection atlas.

PR #408, opened 2023-06-01
Relax ipyleaflet pinning

alimanfoo:
Previously I had somehow got the idea that ipyleaflet 0.17.0 didn't work on colab. I therefore pinned the dependency to `>0.17.0`. However, perhaps that was a transient thing, because now 0.17.0 seems to work fine on colab. Also, 0.17.0 is actually better than more recent versions because it doesn't have a bug where maps are blank after conversion to html (https://github.com/jupyter-widgets/ipyleaflet/issues/1090). Hence we should allow malariagen_data to be installed with ipyleaflet 0.17.0.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -142,24 +142,25 @@ test = ["pytest", "pytest-cov"]
 
 [[package]]
 name = "anyio"
-version = "3.6.2"
+version = "3.7.0"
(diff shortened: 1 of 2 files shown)


alimanfoo:
cc @ahernank, doing this to make it easier to install malariagen_data together with ipyleaflet 0.17.0.

Issue #363, opened 2023-03-15
Better error message when accessing CNV data and some data is missing


alimanfoo:
When trying to call functions like `ag3.gene_cnv('AGAP001356')`, if there are any releases where CNV data is not yet available, this will generate a cryptic error message:

```
FileNotFoundError: '.zmetadata'
```

Could be useful to either give a clearer error message, or just return data from the sample_sets that do have them.

alimanfoo:
Yep this error will come and go, depending on the state of the data. If at any point there is an incomplete data release, where sample metadata and SNP data are present but CNV data are not present yet, you will hit this.

Also IIRC we may have a specific problem with Ag3.7 because some sample sets will never have CNV calls because they are outside the species we run CNV calling on.

leehart:
@alimanfoo As mentioned in #555, it looks like if we skip samples without CNV HMM during `cnv_hmm()`, which is used by `_gene_cnv()` and indirectly by `gene_cnv_frequencies()`, then this will avoid the error messages, but it's not clear to me if there are any unintended implications, e.g. statistically or in plots, which might somehow be misleading.

```python
            lx = []
            for r in regions:
                ly = []
                for s in sample_sets:
                    y = self._cnv_hmm_dataset(
                        contig=r.contig,
                        sample_set=s,
                        inline_array=inline_array,
                        chunks=chunks,
                    )

                    # If no CNV HMM dataset was found then skip
                    if y is None:
                        continue

                    ly.append(y)

                debug("concatenate data from multiple sample sets")
                x = simple_xarray_concat(ly, dim=DIM_SAMPLE)

                debug("handle region, do this only once - optimisation")
                if r.start is not None or r.end is not None:
                    start = x["variant_position"].values
                    end = x["variant_end"].values
                    index = pd.IntervalIndex.from_arrays(start, end, closed="both")
                    # noinspection PyArgumentList
                    other = pd.Interval(r.start, r.end, closed="both")
                    loc_region = index.overlaps(other)  # type: ignore
                    x = x.isel(variants=loc_region)

                lx.append(x)
```

leehart:
I believe this is resolved now. If there are still problems with the handling of missing CNV data, please open a new issue with a description.

PR #409, opened 2023-06-02
Add check for missing CNV HMM data

leehart:
Re: issue #363 

This is just a small amend to catch missing CNV HMM data, identical to the existing approach for `open_cnv_coverage_calls()`.

However, I expect it would be preferable to just return data from the sample_sets that do have CNV data. Alas, it's not clear to me how best to achieve that yet.


alimanfoo:
Hi @leehart, we might need to give this a little thought.

In the case of the CNV HMM data, it can happen that some sample sets have no data yet because it hasn't been released. As you suggest, in this case it might be better to just load data for the sample sets that have data, rather than raise an exception. Often user's hit this error because they are loading data for a different sample set completely, and so the missing data doesn't actually impact them at all.

In the case of the coverage calls data, it is slightly different, because some sample sets will not ever have data for a certain analyses, because they don't contain mosquitoes of certain species. We might want to treat that case differently, not sure.

hmm...

alimanfoo:
Hi @leehart, just a reminder to ourselves, we should revisit this one and see if we can figure out what to do.

leehart:
Closing because we probably won't solve the issue this way.

Issue #404, opened 2023-05-26
Anopheles refactor CNV data access functions


alimanfoo:
As part of work towards #366 we'd like to refactor the functions for accessing CNV data out into a separate module and class, following a similar approach to that taken for SNP data and haplotype data.

We'd also like to make these functions available as part of the Af1 API, as the CNV data will be available soon.

Raising this issue to plan out the work involved.

alimanfoo:
Here is a proposed implementation plan:

## Part 1 - setup

1. Create a new module `malariagen_data.anoph.cnv_data`. Inside that module, declare a new class `AnophelesCnvData`. This class will inherit from parents `AnophelesSampleMetadata`, `AnophelesGenomeFeaturesData`, `AnophelesGenomeSequenceData`. Constructor for this class probably doesn't need to take any specific arguments, should just be able to take `**kwargs` and pass through to `super().__init__(**kwargs)`.
2. Change the definition of the `AnophelesDataResource` class to add `AnophelesCnvData` as a parent class. Make sure this occurs as the first parent in the list of parent classes.

At this point, you should be ready to start moving CNV data access functions from the `Ag3` class up to the `AnophelesCnvData` class. As you're doing so, the existing tests in the `test_ag3` module should still continue to run and pass. The `plot_cnv` notebook should also continue to work.

## Part 2 - code refactoring

Here's a suggested order for moving code from `Ag3` to `AnophelesCnvData`:

3. Move the declaration of the cache variables in the constructor function `_cache_cnv_hmm`, `_cache_cnv_coverage_calls`, `_cache_cnv_discordant_read_calls`.
4. Move the methods `open_cnv_hmm`, `_cnv_hmm_dataset`, `cnv_hmm`.
5. Move the methods `open_cnv_coverage_calls`, `_cnv_coverage_calls_dataset`, `cnv_coverage_calls`.
6. Move the methods `open_cnv_discordant_read_calls`, `_cnv_discordant_read_calls_dataset`, `cnv_discordant_read_calls`.
7. Move the plotting methods `plot_cnv_hmm_coverage_track`, `plot_cnv_hmm_coverage`, `plot_cnv_hmm_heatmap_track`, `plot_cnv_hmm_heatmap`.

Note that I suggest leaving the methods relating to CNV frequencies alone for now (`gene_cnv_frequencies` etc.) and dealing with those in a later PR.

## Part 3 - type hints and documentation

[...]

PR #411, opened 2023-06-08
Refactor CNV data functions for Anopheles (inc. Af1)

leehart:
Re: #404

- [x] Part 1: setup
- [x] Part 2: code refactoring
- [x] Part 3: type hints and documentation
- [x] Part 4: test refactoring


leehart:
Perhaps `lookup_sample()` belongs in `AnophelesSampleMetadata` instead.

leehart:
Need to resolve the `contig` param of `cnv_discordant_read_calls()` in `AnophelesCnvData`.
Can't use `base_params.contig` because it needs to support a `list` as well as a `str`.
There is also a note that says:
```
# N.B., we cannot support region instead of contig here, because some
# CNV alleles have unknown start or end coordinates.
```
🤔

leehart:
In order to cater for `cnv_discordant_read_calls()`, I wonder if we need to switch to using something similar to the `region_param_type` for contigs, e.g.
```
single_contig_param_type: TypeAlias = str

contig_param_type: TypeAlias = Union[
    single_contig_param_type,
    List[single_contig_param_type],
    Tuple[single_contig_param_type, ...],
]
```

leehart:
Another approach or convention, which would be a breaking change, could be to use plural param names when either a string or a list or a tuple are accepted, e.g. `contigs`, and use singular when only a string, representing the singular, are accepted, e.g. `contig`.

alimanfoo:
> In order to cater for `cnv_discordant_read_calls()`, I wonder if we need to switch to using something similar to the `region_param_type` for contigs, e.g.
> 
> ```
> single_contig_param_type: TypeAlias = str
> 
> contig_param_type: TypeAlias = Union[
>     single_contig_param_type,
>     List[single_contig_param_type],
>     Tuple[single_contig_param_type, ...],
> ]
> ```

This seems reasonable to me.

malariagen_data/anoph/het_params.py
```diff
@@ -58,6 +58,11 @@
     "Passed through to bokeh circle() function.",
 ]
 
+line_kwargs: TypeAlias = Annotated[
+    Mapping,
+    "Passed through to bokeh line() function.",
+]
+
```
alimanfoo, line 65: Out of scope for this PR?

malariagen_data/anoph/sample_metadata.py
```diff
@@ -7,7 +7,7 @@
 from numpydoc_decorator import doc
 
 from ..util import check_types
-from . import base_params, map_params
+from . import base_params, het_params, map_params
```
alimanfoo, line 10: I don't think we should be importing `het_params` here. The `het_params` should go with the (yet to be created) `het_analysis` module which should be downstream of sample metadata. If we need the `single_sample` param here then I'd suggest moving that parameter up to `base_params`.

leehart:
Need to resolve `plot_cnv_hmm_coverage_track()` use of the string `"auto"` for param `y_max`, which is usually a float.

alimanfoo:
> Perhaps `lookup_sample()` belongs in `AnophelesSampleMetadata` instead.

SGTM.

malariagen_data/anoph/cnv_data.py
```diff
@@ -0,0 +1,885 @@
...
+        self._cache_cnv_hmm = dict()
+        self._cache_cnv_coverage_calls = dict()
+        self._cache_cnv_discordant_read_calls = dict()
+
+    @check_types
+    @doc(
+        summary="Open CNV HMM zarr.",
+        returns="Zarr hierarchy.",
+    )
+    def open_cnv_hmm(self, sample_set: base_params.sample_set):
```
alimanfoo, line 49:
```suggestion
    def open_cnv_hmm(self, sample_set: base_params.sample_set) -> zarr.hierarchy.Group:
```

malariagen_data/anoph/cnv_data.py
```diff
@@ -0,0 +1,885 @@
...
+    )
+    def cnv_hmm(
+        self,
+        region: base_params.region,
+        sample_sets: Optional[base_params.sample_sets] = None,
+        sample_query: Optional[base_params.sample_query] = None,
+        max_coverage_variance: cnv_params.max_coverage_variance = cnv_params.max_coverage_variance_default,
+        inline_array: base_params.inline_array = base_params.inline_array_default,
+        chunks: base_params.chunks = base_params.chunks_default,
+    ):
```
alimanfoo, line 145:
```suggestion
    ) -> xr.Dataset:
```

malariagen_data/anoph/cnv_data.py
```diff
@@ -0,0 +1,885 @@
...
+        return ds
+
+    @check_types
+    @doc(
+        summary="Open CNV coverage calls zarr.",
+        returns="Zarr hierarchy.",
+    )
+    def open_cnv_coverage_calls(
+        self, sample_set: base_params.sample_set, analysis: hap_params.analysis
+    ):
```
alimanfoo, line 217:
```suggestion
    ) -> zarr.hierarchy.Group:
```

malariagen_data/anoph/cnv_data.py
```diff
@@ -0,0 +1,885 @@
...
+
+        return ds
+
+    @check_types
+    @doc(
+        summary="Open CNV coverage calls zarr.",
+        returns="Zarr hierarchy.",
+    )
+    def open_cnv_coverage_calls(
+        self, sample_set: base_params.sample_set, analysis: hap_params.analysis
```
alimanfoo, line 216: I feel like we shouldn't be using the `hap_params.analysis` type here, but should define a new type like `cnv_params.coverage_calls_analysis`. This is because the possible values will be slightly different, e.g., there is no "gamb_colu_arab" CNV analysis in Ag3 IIRC.

malariagen_data/anoph/cnv_data.py
```diff
@@ -0,0 +1,885 @@
...
+            return self._cache_cnv_coverage_calls[key]
+        except KeyError:
+            release = self.lookup_release(sample_set=sample_set)
+            release_path = self._release_to_path(release)
+            path = f"{self._base_path}/{release_path}/cnv/{sample_set}/coverage_calls/{analysis}/zarr"
+            # N.B., not all sample_set/analysis combinations exist, need to check
+            marker = path + "/.zmetadata"
+            if not self._fs.exists(marker):
+                raise ValueError(
+                    f"analysis f{analysis!r} not implemented for sample set {sample_set!r}"
```
alimanfoo, line 229:
```suggestion
                    f"CNV coverage calls analysis f{analysis!r} not implemented for sample set {sample_set!r}."
```

leehart:
It was easier to create a new base_param named `contigs` than it was to change all cases that use `contig` to use `single_contig` (there is only one case that supports a list of contigs), but this leads to a naming inconsistency with `region`. I am tempted to rename and refactor all cases of `region` to use `regions` instead. This would make the difference between singular and plural types more obvious, and consistent.
```
contig: TypeAlias = Annotated[
    single_contig_param_type,
    """
    Reference genome contig name. See the `contigs` property for valid contig
    names.
    """,
]

contigs: TypeAlias = Annotated[
    contig_param_type,
    """
    Reference genome contig name. See the `contigs` property for valid contig
    names. Can also be a sequence (e.g., list) of contigs.
    """,
]

single_region: TypeAlias = Annotated[
    single_region_param_type,
    """
    Region of the reference genome. Can be a contig name, region string
    (formatted like "{contig}:{start}-{end}"), or identifier of a genome
    feature such as a gene or transcript.
    """,
]

region: TypeAlias = Annotated[
    region_param_type,
    """
    Region of the reference genome. Can be a contig name, region string
    (formatted like "{contig}:{start}-{end}"), or identifier of a genome
    feature such as a gene or transcript. Can also be a sequence (e.g., list)
    of regions.
    """,
]
```

malariagen_data/anoph/cnv_data.py
```diff
@@ -0,0 +1,885 @@
...
+        ds = simple_xarray_concat(lx, dim=DIM_VARIANT)
+
+        return ds
+
+    @check_types
+    @doc(
+        summary="Open CNV discordant read calls zarr.",
+        returns="Zarr hierarchy.",
+    )
+    def open_cnv_discordant_read_calls(self, sample_set: base_params.sample_set):
```
alimanfoo, line 378:
```suggestion
    def open_cnv_discordant_read_calls(self, sample_set: base_params.sample_set) -> zarr.hierarchy.Group:
```

alimanfoo (commented):
Looking great so far :)

A few minor comments...

malariagen_data/anoph/het_params.py
leehart: Used in methods like `plot_cnv_hmm_coverage_track()`.

alimanfoo:
> I am tempted to rename and refactor all cases of `region` to use `regions` instead. This would make the difference between singular and plural types more obvious, and consistent.

No objection. Would also then have consistency in type name with the `sample_set` and `sample_sets` types too.

malariagen_data/anoph/sample_metadata.py
leehart: Ha, OK. I think that's what I did the first time, and then changed my mind for some reason! I wasn't sure what `het_params` was for. Thanks. I'll see if I can.

malariagen_data/anoph/cnv_data.py
leehart: Thanks. I missed all of these!

malariagen_data/anoph/cnv_data.py
leehart: I did wonder about that, thanks. Will do.

leehart:
I'm now wondering if it makes sense for cnv_data to be using het_params, or whether it would be better to promote these to base_params (or perhaps gplt_params), like single_sample was.
```
    def plot_cnv_hmm_coverage_track(
        self,
        sample: het_params.sample,
        region: base_params.region,
        sample_set: Optional[base_params.sample_set] = None,
        y_max: cnv_params.y_max = cnv_params.y_max_default,
        sizing_mode: gplt_params.sizing_mode = gplt_params.sizing_mode_default,
        width: gplt_params.width = gplt_params.width_default,
        height: gplt_params.height = 200,
        circle_kwargs: Optional[het_params.circle_kwargs] = None,
        line_kwargs: Optional[het_params.line_kwargs] = None,
        show: gplt_params.show = True,
        x_range: Optional[gplt_params.x_range] = None,
        output_backend: gplt_params.output_backend = gplt_params.output_backend_default,
    ) -> gplt_params.figure:
```
I am also slightly concerned about cnv_params now having a mixed-type y_max and a different y_max_default to het_params, although this avoids changing the API.

leehart:
As suspected, having a mixed-type for y_max causes type-safety / inconstencey flags, e.g.
```
fig.yaxis.ticker = list(range(y_max + 1))
```
which would error if a string other that "auto" was used as param
```
        if y_max == "auto":
            y_max = data["call_CN"].max() + 2
```
🤔 

It also looks like there is a problem re-using the het_params circle_kwargs for cnv_data, e.g.
```
circle_kwargs.setdefault("size", 3)
```
Raising `no attribute "setdefault"`.

🤔 

There is also an "incompatible types" issue in cnv_data's treatment of base_param `sample_sets`, i.e.
```
        if sample_sets is not None:
            if isinstance(sample_sets, (list, tuple)):
                sample_sets_text = ", ".join(sample_sets)
            else:
                sample_sets_text = sample_sets
            title += f" - {sample_sets_text}"
```
Somehow clashing with `Union[Sequence[str], str],`, I gather `Sequence` isn't actually an instance of a list or tuple, so gets treated as if it were a string in this logic.

alimanfoo:
Hi @leehart, just a heads-up that I've merged the AIM refactoring PR and resolved a couple of subsequent conflicts here, so please pull before continuing.

leehart:
I suspect this might be overly cautious, and perhaps more trust should be put into the type checking, but I keep changing my mind about it:
```
        title = "CNV HMM"
        if sample_sets is not None:
            if isinstance(sample_sets, Sequence):
                if all(isinstance(s, str) for s in sample_sets):
                    sample_sets_text = ", ".join(sample_sets)
                else:
                    raise TypeError("All elements in 'sample_sets' must be type 'str'")
            elif isinstance(sample_sets, str):
                sample_sets_text = sample_sets
            else:
                raise TypeError("'sample_sets' is an unhandled type")
            title += f" - {sample_sets_text}"
```

I suppose a simpler approach would be something like:
```
        title = "CNV HMM"
        if sample_sets is not None:
            if isinstance(sample_sets, Sequence):
                sample_sets_text = ", ".join(sample_sets)
            else:
                sample_sets_text = sample_sets
            title += f" - {sample_sets_text}"
```
Actually, seeing it in the flesh convinces me it's OK to rely on type checking and errors raised from unhandled/uncaught/unexpected types. One of the boons is to reduce the need to manually raise TypeErrors all over the shop, so I reckon this is in spirit.

leehart:
I gather I need to improve test coverage on `cnv_data.py` from 65% to 97% to satisfy these CI checks.
- https://github.com/malariagen/malariagen-data-python/pull/411/checks?check_run_id=14772791285

leehart:
Nearly there! `96.32% of diff hit (target 96.95%)`

malariagen_data/anoph/gplt_params.py
```diff
@@ -36,6 +36,13 @@
     "Plot height in pixels (px).",
 ]
 
+row_height: TypeAlias = Annotated[
+    int,
+    "Plot height per row (sample) in pixels (px).",
+]
+
+row_height_default: row_height = 7
```
alimanfoo, line 44: This feels quite specific to the CNV heatmap plotting function, which is the only place where it's used I think. I reckon better not to define here, rather can just define within the function signature.

alimanfoo (commented):
Partial review, one small thing I noticed...

malariagen_data/anoph/gplt_params.py
leehart: OK, will do. 
`7` was the default `row_height` for both `plot_cnv_hmm_heatmap` and `plot_cnv_hmm_heatmap_track`.

`plot_frequencies_heatmap` has a default `row_height` of `20`.

`plot_aim_heatmap` has a default `row_height` of `4`.

tests/anoph/conftest.py
```diff
@@ -584,6 +584,318 @@ def simulate_aim_variants(path, contigs, snp_sites, n_sites_low, n_sites_high):
...
+        contig_grp = root.require_group(contig)
+
+        # Create the calldata group for this contig.
+        calldata_grp = contig_grp.require_group("calldata")
+
+        # Get the SNP positions for this contig.
+        snp_pos = snp_sites[f"{contig}/variants/POS"][:]
+
+        # Get the number of sites (SNP positions) for this contig.
+        n_sites = snp_pos.shape[0]
```
alimanfoo, line 643: The CNV HMM data is calculated in 300 bp non-overlapping windows across the genome, and is completely independent of the SNP data. I.e., each "variant" is a window. 

The POS array then gives the window start positions, and the END array should be the window end positions (start positions + 299 IIRC). 

So it would be better here to use the length of the contig to derive the number of windows and window start positions.

tests/anoph/conftest.py
```diff
@@ -584,6 +584,318 @@ def simulate_aim_variants(path, contigs, snp_sites, n_sites_low, n_sites_high):
...
+        variants_grp = contig_grp.require_group("variants")
+
+        # Simulate POS under variants.
+        # Note: choosing a random subset would cause dimension size conflicts.
+        variants_grp.create_dataset(name="POS", data=snp_pos)
+
+        # Simulate END under variants.
+        random_multipliers = np.random.randint(2, 12, size=len(snp_pos))
+        end = random_multipliers * snp_pos
+        variants_grp.create_dataset(name="END", data=end)
```
alimanfoo, line 666: See comment above, POS should be window starts (1, 301, 601, ...) and END should be window ENDs (300, 600, 900, ...).

tests/anoph/conftest.py
```diff
@@ -584,6 +584,318 @@ def simulate_aim_variants(path, contigs, snp_sites, n_sites_low, n_sites_high):
...
+        contig_grp = root.require_group(contig)
+
+        # Create the calldata group for this contig.
+        calldata = contig_grp.require_group("calldata")
+
+        # Get the SNP positions for this contig.
+        snp_pos = snp_sites[f"{contig}/variants/POS"][:]
+
+        # Get the number of sites (SNP positions) for this contig.
+        n_sites = snp_pos.shape[0]
```
alimanfoo, line 722: The CNV coverage calls are also completely independent of the SNP data, and so would be good to simulate them as such.

For these data, each variant is a CNV allele, which can start and end at an arbitrary position in the genome.

To simulate this, you could first randomly decide how many CNV variants to simulate, then simulate their start positions as random integers less than the contig size, then simulate their lengths as random integers in some range. You would also have to sort them so the CNVs appear in ascending order of start position.

POS should give the CNV start coordinate, and END should give the end coordinate.

tests/anoph/conftest.py
```diff
@@ -584,6 +584,318 @@ def simulate_aim_variants(path, contigs, snp_sites, n_sites_low, n_sites_high):
...
+        variants_grp.create_dataset(name="FILTER_qMerge", data=filter_qMerge)
+
+        # Simulate POS under variants.
+        # Note: choosing a random subset would cause dimension size conflicts.
+        variants_grp.create_dataset(name="POS", data=snp_pos)
+
+        # Simulate END under variants.
+        random_multipliers = np.random.randint(2, 12, size=len(snp_pos))
+        end = random_multipliers * snp_pos
+        variants_grp.create_dataset(name="END", data=end)
```
alimanfoo, line 774: See comment above regarding POS and END for coverage calls.

tests/anoph/conftest.py
```diff
@@ -584,6 +584,318 @@ def simulate_aim_variants(path, contigs, snp_sites, n_sites_low, n_sites_high):
...
+        contig_grp = root.require_group(contig)
+
+        # Create the calldata group for this contig.
+        calldata_grp = contig_grp.require_group("calldata")
+
+        # Get the SNP positions for this contig.
+        snp_pos = snp_sites[f"{contig}/variants/POS"][:]
+
+        # Get the number of sites (SNP positions) for this contig.
+        n_sites = snp_pos.shape[0]
```
alimanfoo, line 840: Again these data are independent of SNPs. They are similar to the coverage calls, in that each CNV variant has a start coordinate and an end coordinate. So you could take a similar approach to simulating these.

Of course there are much fewer variants in the discordant reads data than in the coverage calls.

alimanfoo (commented):
There is a sheep ton of work in here, thanks so much @leehart 🙏🏻 

It looks great, I just picked up one thing regarding how the CNV stard and end coordinates are simulated. There should be no dependency between the CNV data and the SNP data, so it would be good to simulate the CNV start and end coordinates independently, according to the different types of CNV calling. Happy to talk through if would help.

leehart:
Ah, I suspected some of this and should have asked! Many thanks @alimanfoo . I'll see what I can do and might need to clarify some things next week.

tests/anoph/conftest.py
leehart: Many thanks again. I wonder what a good maximum number of CNV variants to simulate would be, i.e. to randomly simulate between 1 and N variants?

alimanfoo: Maybe somewhere between 1000 and 10000?

alimanfoo (approved):
Looks good, feel free to merge after making the suggested changes to simulation of CNV coordinates.

tests/anoph/conftest.py
leehart: Thanks. Have done.

tests/anoph/conftest.py
leehart: Thanks. Have done.

tests/anoph/conftest.py
leehart: Thanks. Have done.
The catch in this case is that the number of variants need to be consistent for each contig across all sample sets, because the data get concatenated that way.

tests/anoph/conftest.py
leehart: Thanks. Have done.

tests/anoph/conftest.py
leehart: Thanks. Have done.

PR #413, opened 2023-06-23
Implement plot_samples_bar()

### LINKED ISSUE #390: Sample count bar charts
It would be very useful to have some convenient functions for making bar charts of sample counts, like these...

image

image

alimanfoo:
* Resolves #390


alimanfoo:
Couple of examples:

image

image

PR #416, opened 2023-07-07
Add tools to bokeh plots 

### LINKED ISSUE #414: Add tools to bokeh plots
It would be useful to add the save tool to all bokeh plots, so users have an easy way to download a static version of the plot.

Might also be worth adding the crosshair tool, at least to some plots (e.g., GWSS) to make it easier to determine the genomic coordinates for a feature of interest.

harbi811:
* Resolves #414 

Save and crosshair tools are added to bokeh plots.


alimanfoo:
Thanks @harbi811, looks good. Could you check that these additional tools also appear when you call functions like `plot_h12_gwss()` which include both the H12 track and the genes track?

harbi811:
Thank you @alimanfoo . The added tools appear on all plots with or without the genes track. Here are a couple of screenshots from `plot_fst_gwss()` and `plot_fst_gwss_track()`
Screenshot 2023-07-07 172632

Screenshot 2023-07-07 172644

alimanfoo:
Thanks @harbi811, I guess this must be handled by the merging of tools from different subplots (`merge_tools` parameter) which is great.

alimanfoo (approved):

alimanfoo:
Merged, thanks @harbi811 🎆

harbi811:
Thank you @alimanfoo 😊

PR #418, opened 2023-08-09
Update cohorts version & expected frq col

### LINKED ISSUE #417: Test failures due to year mismatch
I reckon this is due to a recent change to sample collection year (from `2012` to `2006`) in the metadata for sample set AG1000G-GM-B.

Re: https://github.com/malariagen/vector-ops/issues/1996

~I expect it will require a simple update to the relevant test expectations.~ [Looks more complicated.]

```
=================================== FAILURES ===================================
______________ test_allele_frequencies_advanced__sample_sets[3.0] ______________

sample_sets = '3.0'

    @pytest.mark.parametrize(
        "sample_sets", ["AG1000G-BF-A", ["AG1000G-BF-A", "AG1000G-ML-A"], "3.0"]
    )
    def test_allele_frequencies_advanced__sample_sets(sample_sets):
>       _check_snp_allele_frequencies_advanced(
            sample_sets=sample_sets,
        )

tests/test_ag3.py:1327: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

transcript = 'AGAP004707-RD', area_by = 'admin1_iso', period_by = 'year'
sample_sets = '3.0', sample_query = None, min_cohort_size = 10
nobs_mode = 'called', variant_query = 'max_af > 0.02'

    def _check_snp_allele_frequencies_advanced(
        transcript="AGAP004707-RD",
        area_by="admin1_iso",
        period_by="year",
        sample_sets=["AG1000G-BF-A", "AG1000G-ML-A", "AG1000G-UG"],
        sample_query=None,
        min_cohort_size=10,
        nobs_mode="called",
        variant_query="max_af > 0.02",
    ):
        ag3 = setup_ag3(cohorts_analysis="20211101")
    
        ds = ag3.snp_allele_frequencies_advanced(
            transcript=transcript,
            area_by=area_by,
            period_by=period_by,
            sample_sets=sample_sets,
            sample_query=sample_query,
            nobs_mode=nobs_mode,
            variant_query=variant_query,
     'ML-2_gamb_2004',
     'ML-2_gamb_2014',
     'ML-3_colu_2012',
     'ML-3_gamb_2012',
     'ML-4_colu_2004',
     'MW-S_arab_2015',
     'MZ-I_gamb_2004',
     'TZ-05_arab_2015',
     'TZ-05_gamb_2015',
     'TZ-13_arab_2012',
     'TZ-25_gamb_2013',
     'TZ-25_gcx3_2013',
     'TZ-26_arab_2012',
     'UG-E_arab_2012',
     'UG-E_gamb_2012',
     'UG-W_gamb_2012',
    ]
FAILED tests/test_ag3.py::test_gene_cnv_frequencies_advanced__max_coverage_variance[None] - AssertionError: assert ['GM-L_gcx2_2...-N_gcx1_2011'] == ['GM-L_gcx2_2...-N_gcx1_2011']
  At index 0 diff: 'GM-L_gcx2_2006' != 'GM-L_gcx2_2012'
  Full diff:
  - ['GM-L_gcx2_2012', 'GM-M_gcx2_2012', 'GM-N_gcx1_2011']
  ?               ^^
  + ['GM-L_gcx2_2006', 'GM-M_gcx2_2012', 'GM-N_gcx1_2011']
  ?               ^^
=========== 2 failed, 529 passed, 27 warnings in 1794.48s (0:29:54) ============
Error: Process completed with exit code 1.
##[debug]Finishing: Run full test suite
```

ahernank:
* Resolves #417 

To solve the assertion errors, I had to update to the latest cohort version to ensure the cohort names match (i.e. accounting for the single mismatch due to the year change) & update the hard-coded expected_frq_columns on `test_gene_cnv_frequencies__max_coverage_variance()`. 

After this, we also need to update the expected number of positions returned from `test_snp_allele_frequencies__str_cohorts__effects()`, as the number of rows in the AF df is higher than the expected one. This is due to the change in cohorts versions.

Not sure whether there was a reason for using the `20211101` version here (e.g. is this a fixed cohorts release for 3.0?) but it seems like a good chance to update the cohorts version for these tests, given that we have `20230516` as the default in the package. 

Alternatively, if we want to keep the older version instead, it would be a case of manually changing the cohorts file for the GM-B sample set so the column values will properly match.

### DIFF
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -31,7 +31,7 @@ jobs:
         uses: actions/cache/restore@v3
         with:
           path: gcs_cache
-          key: gcs_cache_tests_20230811
+          key: gcs_cache_notebooks_20230811
 
       - name: Run notebooks
         run: poetry run jupyter nbconvert --execute notebooks/* --inplace
@@ -41,4 +41,4 @@ jobs:
         if: always()
         with:
           path: gcs_cache
-          key: gcs_cache_tests_20230811
+          key: gcs_cache_notebooks_20230811
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -1,44 +1,44 @@
 name: notebooks
 on:
-    push:
-        branches:
-            - master
-    pull_request:
-        branches:
-            - master
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 jobs:
-    notebooks:
-        strategy:
-            fail-fast: true
-        runs-on: ubuntu-latest
-        steps:
-            - name: Checkout source
-              uses: actions/checkout@v3
+  notebooks:
+    strategy:
+      fail-fast: true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
 
-            - name: Install poetry
-              run: pipx install poetry==1.4.2
+      - name: Install poetry
+        run: pipx install poetry==1.4.2
 
-            - name: Setup python
-              uses: actions/setup-python@v4
-              with:
-                  python-version: '3.11'
-                  cache: 'poetry'
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: "poetry"
 
-            - name: Install dependencies
-              run: poetry install
+      - name: Install dependencies
+        run: poetry install
 
-            - name: Restore GCS cache
-              uses: actions/cache/restore@v3
-              with:
-                  path: gcs_cache
-                  key: gcs_cache_notebooks_20230425
+      - name: Restore GCS cache
+        uses: actions/cache/restore@v3
+        with:
+          path: gcs_cache
+          key: gcs_cache_tests_20230811
 
-            - name: Run notebooks
-              run: poetry run jupyter nbconvert --execute notebooks/* --inplace
+      - name: Run notebooks
+        run: poetry run jupyter nbconvert --execute notebooks/* --inplace
 
-            - name: Save GCS cache
-              uses: actions/cache/save@v3
-              if: always()
-              with:
-                  path: gcs_cache
-                  key: gcs_cache_notebooks_20230425
+      - name: Save GCS cache
+        uses: actions/cache/save@v3
+        if: always()
+        with:
+          path: gcs_cache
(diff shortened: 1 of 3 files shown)


leehart:
Thanks for looking into this @ahernank 👍 

I don't know why `20211101` is still being used here either,  but using the package's default instead seems like a good move forwards to me. It might be that some of this simply needs updating. I would personally be reluctant to manually change the older cohorts files, in case it breaks other expectations by rewriting history.

ahernank:
Thanks @leehart, annoyingly all tests are passing locally but will have to figure out what is happening with the failures here..

leehart:
Annoying when that happens!
I wonder if local cache, e.g. `gcs_cache`, needs refreshing in order to reproduce test failure.

ahernank:
I did have to clear-up my local cache when I started working on this one, as I couldn't reproduce the error initially. I would have assumed that the cache gets cleaned every time the workflow is run here, but it might be grabbing a specific version (`gcs_cache_tests_20230420`?). 

Almost certainly missing something obvious on my side, will keep thinking..

leehart:
Ah, right. Yeah, I wonder if the key needs updating in `tests.yml` in order to refresh the CI cache. I'm not sure how this works.
```
            - name: Restore GCS cache
              uses: actions/cache/restore@v3
              with:
                  path: gcs_cache
                  key: gcs_cache_tests_20230420

            - name: Run full test suite
              run: poetry run pytest -v tests

            - name: Save GCS cache
              uses: actions/cache/save@v3
              if: always()
              with:
                  path: gcs_cache
                  key: gcs_cache_tests_20230420
```

leehart:
🤞 
By the way, I expect for `notebooks.yml`, the key might need to remain distinct form the `tests.yml`, i.e. `gcs_cache_notebooks_20230811` rather than `gcs_cache_tests_20230811`, judging from the diffs.

leehart:
Any idea what the `@v3` is for, in `restore@v3` and `save@v3`?
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

ahernank:
Yes, I think this is related to GH actions, and relates the node runtime version used (e.g. same as in `checkout@v3`).

leehart (approved):

Issue #387, opened 2023-04-20
Add functions for pairwise FST comparisons between cohorts


KellyLBennett:
Add functions to calculate FST between a pair of cohorts, a set of cohorts and to plot these values as a heat map. 
Draft functions are as follows:

```
def average_fst(
    region,
    admin_level,
    cohort1,
    cohort2,
    blen=10000,
    site_mask=None,
    site_class=None,
    cohort_size=10,
    random_seed=None,
    ):
    cohort1_counts=ag3.snp_allele_counts(region=region, sample_sets=None, sample_query=f"{admin_level} == '{cohort1}'", site_mask=site_mask, site_class=site_class, cohort_size=cohort_size, random_seed=random_seed)
    cohort2_counts=ag3.snp_allele_counts(region=region, sample_sets=None, sample_query=f"{admin_level} == '{cohort2}'", site_mask=site_mask, site_class=site_class, cohort_size=cohort_size, random_seed=random_seed)
    fst_hudson, se_hudson, vb_hudson, _ = allel.blockwise_hudson_fst(cohort1_counts, cohort2_counts, blen=blen)
    return fst_hudson, se_hudson
```

KellyLBennett:
def pairwise_average_fst(
    region,
    admin_level,
    cohort_list,
    blen=10000,
    site_mask=None,
    site_class=None,
    cohort_size=10,
    random_seed=None,
    diag=False,
    annotate_se=False
    ):
    #if wish to sort list by species, will place sample location samples together before year.
    #cohort_list.sort(key = lambda x: x.split("_")[1])
    fst_dict = {key:[] for key in cohort_list}
    se_dict = {key:[] for key in cohort_list}
    for cohort1 in cohort_list:
        for cohort2 in cohort_list:
            fst_hudson, se_hudson = average_fst(region=region, admin_level=admin_level, cohort1=cohort1, cohort2=cohort2, blen=blen, site_mask=site_mask, site_class=site_class, cohort_size=cohort_size, random_seed=random_seed)
            #convert minus numbers to 0
            if fst_hudson < 0:
                fst_hudson = 0
            # put values in dict for df.
            fst_dict[cohort1].append(round(fst_hudson,3))
            se_dict[cohort1].append(round(se_hudson,3))
            print ("calculating FST between " + cohort1, cohort2)
    #put dictionary into pd df and add index.
    fst_df = pd.DataFrame.from_dict(fst_dict)
    fst_df.index = cohort_list
[...]

PR #419, opened 2023-08-14
Add pairwise fst functions

KellyLBennett:
Tagging issue #387 .


leehart:
First I will try to resolve the branch conflicts in the `anopheles.py` file, if that's OK @KellyLBennett ?
(You'll need to pull this branch after I've fixed it, if you want the fixed version.)

leehart:
@KellyLBennett I think I resolved the branch conflict and automated tests are now running.

Was there a plan to include automated tests for this new function? I guess you've been checking that it works in a local notebook or something?

KellyLBennett:
Hi @leehart thanks for resolving the conflict. I have added/been modifying the function in pycharm and running the tests to check everything passes. I am not sure if I should be adding anything else to test this function. I didn't see any specific tests for some (presumably recent) functions and so wasn't sure if its standard to add something in but happy to take advise on this.
    On Monday, 14 August 2023 at 10:23:38 BST, Lee ***@***.***> wrote:  
 
 


@KellyLBennett I think I resolved the branch conflict and automated tests are now running.

Was there a plan to include automated tests for this new function? I guess you've been checking that it works in a local notebook or something?

leehart:
Thanks @KellyLBennett . Do you have any example code that I can try to run that should demonstrate the function working properly?

E.g.
```
ag3 = malariagen_data.Ag3()
ag3.average_fst(
        region="2L",
        cohort1_query="foo",
        cohort2_query="bar",
        cohort_size"baz"
        n_jack = 300,
        site_class = "qux",
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )
```

KellyLBennett:
Sure, you could try...

ag3.average_fst (

        region="2L",

        cohort1_query="cohort_admin1_year == 'TZ-05_arab_2015'",

        cohort2_query="cohort_admin1_year == 'TZ-26_arab_2012'",

        n_jack = 300,

        site_class = "arab",

        min_cohort_size = 10,

        max_cohort_size = 100,

        random_seed = 41
    )
    On Monday, 14 August 2023 at 10:43:25 BST, Lee ***@***.***> wrote:  
 
 


Thanks @KellyLBennett . Do you have any example code that I can try to run that should demonstrate the function working properly?

E.g.
ag3 = malariagen_data.Ag3()
ag3.average_fst(
        region="2L",
        cohort1_query="foo",
        cohort2_query="bar",
        cohort_size"baz"
        n_jack = 300,
        site_class = "qux",
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )

leehart:
Perhaps a silly question: What's the difference between `cohort_size` and `min_cohort_size`?

For example, when I run the function with the following parameters (which might be nonsensical), I get the error:
```
ValueError: not enough samples (27) for minimum cohort size (50)
```

```
ag3.average_fst(
        region="2L",
        cohort1_query="cohort_admin2_year == 'ML-2_Kati_colu_2014'",
        cohort2_query="cohort_admin2_year == 'ML-2_Kati_gamb_2014'",
        cohort_size=50,
        n_jack = 300,
        site_class = "UTR_5PRIME",
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )
```

I see that the cohort has only 27 samples, and I've specified a cohort_size of `50`, so it makes sense that there would be an error, but it's not clear to me what `min_cohort_size` does if `cohort_size` is effectively the minimum cohort size. I suspect that `cohort_size` might be a downsampling target, e.g. so that a cohort of 100 samples would downsample to only 50 samples in this case, but a cohort of 27 samples cannot downsample to 50.

leehart:
Thanks @KellyLBennett .
When I try to run the function with these params, I get the error:
```
TypeError: average_fst() missing 1 required positional argument: 'cohort_size'
```
```
ag3.average_fst (
        region="2L",
        cohort1_query="cohort_admin1_year == 'TZ-05_arab_2015'",
        cohort2_query="cohort_admin1_year == 'TZ-26_arab_2012'",
        n_jack=300,
        site_class = "arab",
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )
```

I'm starting to wonder whether `min_cohort_size` and `max_cohort_size` were meant to replace the `cohort_size` argument?

KellyLBennett:
Yes sorry @leehart its either that you can specify cohort_size which hopefully has a default of None or you can add min_cohort size and max_cohort size. The cohort size does indeed mean to use that as the threshold, similar to min_cohort_size. I thought you should be able to not include cohort_size and it would run with the default and hopefully vice versa, at least that was the intent, but there could very well be a mistake.
    On Monday, 14 August 2023 at 11:30:29 BST, Lee ***@***.***> wrote:  
 
 


Thanks @KellyLBennett .
When I try to run the function with these params, I get the error:
TypeError: average_fst() missing 1 required positional argument: 'cohort_size'
ag3.average_fst (
        region="2L",
        cohort1_query="cohort_admin1_year == 'TZ-05_arab_2015'",
        cohort2_query="cohort_admin1_year == 'TZ-26_arab_2012'",
        n_jack=300,
        site_class = "arab",
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )

I'm starting to wonder whether min_cohort_size and max_cohort_size were meant to replace the cohort_size argument?

leehart:
We might need to do something similar to the `fst_gwss()` function, if you want to support all three parameters. Here the parameters (inc. `cohort_size`) are marked as optional and have a default:

```
    def fst_gwss(
        self,
        contig: base_params.contig,
        window_size: fst_params.window_size,
        cohort1_query: base_params.sample_query,
        cohort2_query: base_params.sample_query,
        site_mask: base_params.site_mask = DEFAULT,
        cohort_size: Optional[base_params.cohort_size] = fst_params.cohort_size_default,
        min_cohort_size: Optional[
            base_params.min_cohort_size
        ] = fst_params.min_cohort_size_default,
        max_cohort_size: Optional[
            base_params.max_cohort_size
        ] = fst_params.max_cohort_size_default,
    ) -> Tuple[np.ndarray, np.ndarray]:
```

leehart:
@KellyLBennett While I have the box open, shall I try to fix this by updating `average_fst()`'s `cohort_size` param to:

```
cohort_size: Optional[base_params.cohort_size] = fst_params.cohort_size_default,
```

KellyLBennett:
Yes please @leehart
    On Monday, 14 August 2023 at 11:48:26 BST, Lee ***@***.***> wrote:  
 
 


@KellyLBennett While I have the box open, shall I try to fix this by updating average_fst()'s cohort_size param to:
cohort_size: Optional[base_params.cohort_size] = fst_params.cohort_size_default,

leehart:
Just to note the doc for the `site_class` param, doesn't take "arab" as an arg.
```
site_class: TypeAlias = Annotated[
    str,
    """
    Select sites belonging to one of the following classes: CDS_DEG_4,
    (4-fold degenerate coding sites), CDS_DEG_2_SIMPLE (2-fold simple
    degenerate coding sites), CDS_DEG_0 (non-degenerate coding sites),
    INTRON_SHORT (introns shorter than 100 bp), INTRON_LONG (introns
    longer than 200 bp), INTRON_SPLICE_5PRIME (intron within 2 bp of
    5' splice site), INTRON_SPLICE_3PRIME (intron within 2 bp of 3'
    splice site), UTR_5PRIME (5' untranslated region), UTR_3PRIME (3'
    untranslated region), INTERGENIC (intergenic, more than 10 kbp from
    a gene).
    """,
]
```

leehart:
@KellyLBennett Do we have any way to check that the results returned by this function are correct, just to be sure that it is working correctly, e.g. any set of arguments/values for the parameters of this function that we know the outputs for?

The function seems to be functioning now, but I don't know if the values it returns are as they should be. For example:
```
ag3.average_fst(
        region="2L",
        cohort1_query="cohort_admin2_year == 'ML-2_Kati_colu_2014'",
        cohort2_query="cohort_admin2_year == 'ML-2_Kati_gamb_2014'",
        cohort_size=27,
        n_jack=300,
        site_class = "UTR_5PRIME",
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )
```
Returns `(0.045100720849134884, 0.002425472105329579)`.

```
ag3.average_fst (
        region="2L",
        cohort1_query="cohort_admin1_year == 'TZ-05_arab_2015'",
        cohort2_query="cohort_admin1_year == 'TZ-26_arab_2012'",
        n_jack=300,
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )
```
Returns: `(0.00445733333528514, 0.0004737521443549097)`.

It would be reassuring to have some way of validating that the returned values are exactly what we expect them to be. (I'm sorry, I don't know much about this.)

KellyLBennett:
I am unclear of a way to do this at the moment @leehart. I can only think of adding some test cohorts with known values, but the value is likely to change slightly on iteration and so I'm unsure of how to do this. 
    On Monday, 14 August 2023 at 13:57:16 BST, Lee ***@***.***> wrote:  
 
 


@KellyLBennett Do we have any way to check that the results returned by this function are correct, just to be sure that it is working correctly, e.g. any set of arguments/values for the parameters of this function that we know the outputs for?

The function seems to be functioning now, but I don't know if the values it returns are as they should be. For example:
ag3.average_fst(
        region="2L",
        cohort1_query="cohort_admin2_year == 'ML-2_Kati_colu_2014'",
        cohort2_query="cohort_admin2_year == 'ML-2_Kati_gamb_2014'",
        cohort_size=27,
        n_jack=300,
        site_class = "UTR_5PRIME",
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )

Returns (0.045100720849134884, 0.002425472105329579).
ag3.average_fst (
        region="2L",
        cohort1_query="cohort_admin1_year == 'TZ-05_arab_2015'",
        cohort2_query="cohort_admin1_year == 'TZ-26_arab_2012'",
        n_jack=300,
        min_cohort_size = 10,
        max_cohort_size = 100,
        random_seed = 41
    )

Returns: (0.00445733333528514, 0.0004737521443549097).

It would be reassuring to have some way of validating that the returned values are exactly what we expect them to be. (I'm sorry, I don't know much about this.)

leehart:
> I am unclear of a way to do this at the moment @leehart. I can only think of adding some test cohorts with known values, but the value is likely to change slightly on iteration and so I'm unsure of how to do this.

No worries, @KellyLBennett . I suppose we can add some tests in later, if we can figure out a good way to check. It's just difficult to review and approve the code if I can't verify that it's working properly. The code looks relatively simple in this case, so I'm not too worried.

Before I approve, is this PR just for the `average_fst()` function? Do you plan to create additional PRs once this has been merged in (for `pairwise_average_fst()` and `plot_pairwise_average_fst()`)? I'm just thinking of changing the title of this PR, if that's the case.

KellyLBennett:
@leehart yes I was intending to add further functions, but if you would like I could just continue working on this pull request with the additional functions before merge, then we can think further on testing the function. 
    On Monday, 14 August 2023 at 15:30:06 BST, Lee ***@***.***> wrote:  
 
 



I am unclear of a way to do this at the moment @leehart. I can only think of adding some test cohorts with known values, but the value is likely to change slightly on iteration and so I'm unsure of how to do this.


No worries, @KellyLBennett . I suppose we can add some tests in later, if we can figure out a good way to check. It's just difficult to review and approve the code if I can't verify that it's working properly. The code looks relatively simple in this case, so I'm not too worried.

Before I approve, is this PR just for the average_fst() function? Do you plan to create additional PRs once this has been merged in (for pairwise_average_fst() and plot_pairwise_average_fst())? I'm just thinking of changing the title of this PR, if that's the case.

leehart:
Just to note that the notebooks CI checks are failing here due to a separate issue regarding cohorts data changes, #424

KellyLBennett:
Thanks @leehart it would be good to get comments on the function and tests that are in the PR, but I am tempted to add all I have on this PR for review which means adding two more functions.

KellyLBennett:
Tests that are failing in pycharm are those in test_af1 (test_h1x_gwss and test_fst_gwss) and in test_ag3 (test_gene_cnv_freq) but these should not be modified by the addition of the Fst functions.

leehart:
Just to note that the CI linting failure reports
```
malariagen_data/anopheles.py:2327: error: Unsupported operand types for < ("int" and "None")  [operator]
[12](https://github.com/malariagen/malariagen-data-python/actions/runs/6508825929/job/17678958681?pr=419#step:6:13)
malariagen_data/anopheles.py:2327: note: Right operand is of type "int | None"
```

Which refers to line 2327 in `anopheles.py`, part of the `diversity_stats()` function:
```python
        if sample_query is not None:
            cohort_queries = {
                cohort_label: f"({cohort_query}) and ({sample_query})"
                for cohort_label, cohort_query in cohort_queries.items()
            }
```

Which hasn't changed in this PR, as far as I can see.


This seems related to some other weird issues we've been having with tests failing on code that hasn't been changed and apparently passed previous tests, which seems to suggest that something might have moved the goal posts on the testing environment front.

The `pyproject.toml` and `poetry.lock` haven't changed for 5 months, so I guess it's not that. Same goes for the `linting.yml` and `coverage.yml` workflows. We updated the `gcs_cache` keys for `tests.yml` and `notebooks.yml` a couple of months ago, but I can't see how that would cause this effect. Why would `linting.yml` change its outcome? 🤔 

The linting tests that are failing in CI sound different to the tests that @KellyLBennett reports as failing in PyCharm. I gather there are differences in the PyCharm environment compared to the CI environment.

malariagen_data/anopheles.py
```diff
@@ -2363,6 +2426,210 @@ def diversity_stats(
...
+        random_seed: base_params.random_seed = 42,
+    ):
+        # set up cohort queries
+        cohort_queries_checked = self._setup_cohorts(
+            cohorts,
+            sample_sets=sample_sets,
+            sample_query=sample_query,
+            cohort_size=cohort_size,
+            min_cohort_size=min_cohort_size,
+        )
```
alimanfoo, line 2525:
```suggestion
        cohorts_checked = self._setup_cohorts(
            cohorts,
            sample_sets=sample_sets,
            sample_query=sample_query,
            cohort_size=cohort_size,
            min_cohort_size=min_cohort_size,
        )
```

malariagen_data/anopheles.py
```diff
@@ -2363,6 +2426,210 @@ def diversity_stats(
...
+
+        cohort1_query_list = []
+        cohort2_query_list = []
```
alimanfoo, line 2528:
```suggestion
        cohort_ids = list(cohorts_checked.keys())
        cohort_queries = list(cohorts_checked.values())
        cohort1_ids = []
        cohort2_ids = []
```

malariagen_data/anopheles.py
```diff
@@ -2363,6 +2426,210 @@ def diversity_stats(
...
+
+        for i in range(len(cohort_queries_checked)):
+            for j in range(i + 1):
+                (
+                    fst_hudson,
+                    se_hudson,
+                ) = self.average_fst(
+                    region=region,
+                    cohort1_query=list(cohort_queries_checked.values())[i],
+                    cohort2_query=list(cohort_queries_checked.values())[j],
```
alimanfoo, line 2540:
```suggestion
                    cohort1_query=cohort_queries[i],
                    cohort2_query=cohort_queries[j],
```

malariagen_data/anopheles.py
```diff
@@ -2363,6 +2426,210 @@ def diversity_stats(
...
+                    site_mask=site_mask,
+                    site_class=site_class,
+                    random_seed=random_seed,
+                )
+                # convert minus numbers to 0
+                if fst_hudson < 0:
+                    fst_hudson = 0
+                # add values to lists
+                cohort1_query_list.append(list(cohort_queries_checked.keys())[i])
+                cohort2_query_list.append(list(cohort_queries_checked.keys())[j])
```
alimanfoo, line 2554:
```suggestion
                cohort1_ids.append(cohort_ids[i])
                cohort2_ids.append(cohort_ids[j])
```

alimanfoo (commented):
Hi @KellyLBennett, revised implementation for pairwise_average_fst() looks good, just some suggestions to simplify the code a little and make it clear when you're handling cohort identifiers versus cohort queries. 

Also by convention I'm trying to keep dataframe column names always all lower case, so you never have to remember if anything is upper case or not.

tests/test_ag3.py
```diff
@@ -1829,6 +1829,102 @@ def test_h1x_gwss():
...
+        n_jack=n_jack,
+        site_mask=site_mask,
+    )
+
+    # check data
+    assert isinstance(fst_hudson, np.float64)
+    assert isinstance(se_hudson, np.float64)
+
+    # check dimensions
+    assert fst_hudson.ndim == se_hudson.ndim == 0
```
alimanfoo, line 1853:
```suggestion
    assert np.isscalar(fst_hudson)
    assert np.isscalar(se_hudson)
```

A slightly more obvious way to show we expect scalar values here.

alimanfoo (commented):
Hi @KellyLBennett, a few small suggestions on the unit tests.

tests/test_ag3.py
```diff
@@ -1829,6 +1829,102 @@ def test_h1x_gwss():
     assert np.all(h1x >= 0)
 
 
+def test_average_fst():
+    ag3 = setup_ag3()
+    region = "3L"
```
alimanfoo, line 1834:
```suggestion
    region = "3L:15,000,000-16,000,000"
```

For testing purposes, should be fine to run on a smaller genome region, which will mean tests run faster.

tests/test_ag3.py
```diff
@@ -1829,6 +1829,102 @@ def test_h1x_gwss():
...
+    # check some values
+    assert_allclose(fst_hudson, 0.039983, rtol=1e5), fst_hudson
+    assert_allclose(se_hudson, 0.003327, rtol=1e5), se_hudson
+    assert np.all(fst_hudson <= 1)
+    assert np.all(fst_hudson >= -0.05)
+
+
+def test_pairwise_average_fst():
+    ag3 = setup_ag3()
+    region = "3L"
```
alimanfoo, line 1864:
```suggestion
    region = "3L:15,000,000-16,000,000"
```

For testing purposes, should be fine to run on a smaller genome region, which will mean tests run faster.

alimanfoo (commented):
Couple more suggestions to avoid tests running too slowly.

malariagen_data/anopheles.py
```diff
@@ -2363,6 +2426,212 @@ def diversity_stats(
...
+
+        cohort_ids = list(cohorts_checked.keys())
+        cohort_queries = list(cohorts_checked.values())
+        cohort1_ids = []
+        cohort2_ids = []
+        fst_stats = []
+        se_stats = []
+
+        for i in range(len(cohorts_checked)):
+            for j in range(i + 1):
```
alimanfoo, line 2535:
```suggestion
        n_cohorts = len(cohorts_checked)
        for i in range(n_cohorts):
            for j in range(i + 1, n_cohorts):
```

Bug fix - in the inner loop, need to iterate starting from i+1, up to n_cohorts.

alimanfoo (commented):
Noticed a bug here, will mean the test dataframe also needs to be changed to have the correct expected cohort pairs.

alimanfoo:
Hi @KellyLBennett, it would be great if you could apply suggestions from reviews above then fix the unit tests so they're all passing.

Once you've done that, could you add a notebook at file path `notebooks/fst_average.ipynb` which demonstrates the `plot_pairwise_average_fst()` function, to show that's working. Would be good to include examples where `annotate_se=True` and `annotate_se=False` to show how that parameter works. 

🙏🏻

malariagen_data/anopheles.py
alimanfoo: ...in order to get the computation working on the correct pairs of cohorts, i.e., all pairs run once and avoiding self comparisons.

KellyLBennett:
Hi @alimanfoo I have updated the tests and added the example notebook. All runs fine in the notebook but tests are failing in pycharm. This seems to be something to do with accessing the sample metadata which I have not modified, and it seems many other functions fail their tests for the same reason. Tests were failing for other CI reasons previously,  do you know what is happening on this occasion?

alimanfoo:
Hi @KellyLBennett, it may be that some data has changed on GCS, and you need to delete the local cache of files from GCS. if you delete the "gcs_cache" folder within your local clone of the repo, then rerun tests locally, does that fix?

leehart:
It looks like the CI checks are failing here for a couple of reasons, seen by clicking on `Details`:

1. `fst_hudson` is not defined in the `average_pairwise_fst.ipynb` notebook in the line
```
assert isinstance(fst_hudson, np.float64)
```

2. Linting issue on line 2327 of `anopheles.py` "Unsupported operand types for < ("int" and "None")"
```
if n_samples < cohort_size:
```
...presumably because one of those variables is actually `None` at runtime, rather than the expected comparison of two integers. I guess it might be because `cohort_size` gets set (or defaults) to `None`, meaning we might have to handle that scenario.

malariagen_data/anopheles.py
```diff
@@ -2269,6 +2273,65 @@ def cohort_diversity_stats(
...
 
+    # @staticmethod?
+    def _setup_cohorts(
+        self,
+        cohorts: base_params.cohorts,
+        sample_sets: Optional[base_params.sample_sets] = None,
+        sample_query: Optional[base_params.sample_query] = None,
+        cohort_size: Optional[base_params.cohort_size] = 15,
+        min_cohort_size: Optional[base_params.min_cohort_size] = None,
+    ):
```
alimanfoo, line 2284:
```suggestion
    def _setup_cohorts(
        self,
        cohorts: base_params.cohorts,
        sample_sets: Optional[base_params.sample_sets],
        sample_query: Optional[base_params.sample_query],
        cohort_size: Optional[base_params.cohort_size],
        min_cohort_size: Optional[base_params.min_cohort_size],
    ):
```

Because this is a private method, default values aren't needed here. Default values will be provided in the public method which calls this internally.

tests/test_ag3.py
```diff
@@ -1829,6 +1829,85 @@ def test_h1x_gwss():
...
+        site_mask=site_mask,
+    )
+
+    # check data
+    assert isinstance(fst_df, pd.core.frame.DataFrame)
+
+    # check some values
+    pd.testing.assert_frame_equal(fst_df, test_df, rtol=1e5)
+    assert np.all(fst_df["fst"] <= 1)
+    assert np.all(fst_df["fst"] >= -0.05)
```
alimanfoo, line 1908:
```suggestion
    assert np.all(fst_df["fst"] >= 0)
```

Because negative values are being clipped to zero, we could check that here.

alimanfoo (commented):
Thanks @KellyLBennett.

* In the notebook, if you do a `!pip install malariagen_data` then you will be obliterating the local development version of malariagen_data installed in the poetry environment. Rather, what you want to do in the notebook is run the local development version of malariagen_data, including the new functions. This is why you won't see `pip install malariagen_data` in any of the other notebooks. To run this notebook with your local development version of malariagen_data, do `poetry run jupyter lab` - this will run a jupyter server using the poetry environment where your development version of malariagen_data is installed. This will also mean you don't need to duplicate the implementation of the fst functions within the notebook, you can just call them as methods on the Ag3 class. This is what the notebooks are for, they are a way to exercise the new code you've added within malariagen_data. In other words, there should be no need to duplicate any code in the notebook which is already present within malariagen_data. 

* In the linting CI check, mypy is failing with this message:

```
malariagen_data/anopheles.py:2327: error: Unsupported operand types for < ("int" and "None")  [operator]
malariagen_data/anopheles.py:2327: note: Right operand is of type "int | None"
```

...this is because at line 2327 in anopheles.py the `cohort_size` variable could be None. So we need to check first if `cohort_size is not None` before we can check if `n_samples < cohort_size`. This can be probably be done by changing the line to `if cohort_size is not None and n_samples < cohort_size:`.

* Couple of other comments below.

alimanfoo:
> 1. `fst_hudson` is not defined in the `average_pairwise_fst.ipynb` notebook in the line
> 
> ```
> assert isinstance(fst_hudson, np.float64)
> ```

Thanks @leehart, this is because the notebook is doing a `!pip install malariagen_data` but should rather just `import malariagen_data` as the correct (development) version of malariagen_data is already available in the poetry environment.

Also this assertion isn't necessary in the notebook anyway, it's already present in the unit tests.

KellyLBennett:
I have made the suggested changes @alimanfoo and have updated the notebook. Please let me know if anything else is needed.

alimanfoo:
Thanks @KellyLBennett, I'll take a look ASAP.

alimanfoo:
Hi @KellyLBennett,

Looking at this plot from the example notebook...

image

...this is supposed to be an analysis of pairwise Fst between three populations. So there are three pairs of cohorts and thus three Fst values to show. But this plot only shows one Fst value.

I think what we want here is each axis should have all three cohorts on it, then we'll have squares on the heatmap corresponding to all pairs. I.e., the heatmap should be all cohorts against all cohorts. Then you have a lower triangle to show Fst values, and an upper triangle to show SE values (if desired). A bit like this...

image

alimanfoo (commented):
Hi @KellyLBennett, Fst plotting is looking good, a few minor suggestions.

KellyLBennett:
I also removed the following since this code prevented the use of max = None.
for i in range(min(fig_df.shape)):
        #fig_df.iloc[i, i] = ""

Because NaNs are now present in the matrix this generates an error (invalid value encountered in reduce). I have added a warning suppression to prevent this. I am not sure if there is a better way to prevent the warning.

with np.errstate(invalid="ignore"):

alimanfoo:
> I also removed the following since this code prevented the use of max = None. for i in range(min(fig_df.shape)): #fig_df.iloc[i, i] = ""
> 
> Because NaNs are now present in the matrix this generates an error (invalid value encountered in reduce). I have added a warning suppression to prevent this. I am not sure if there is a better way to prevent the warning.
> 
> with np.errstate(invalid="ignore"):

Thanks Kelly, this seems reasonable to me.

alimanfoo (approved):
Hi @KellyLBennett, this LGTM, thank you for sticking with it!

I pushed some commits with a few more bits and pieces, including:

* Added types and docstring for the plot_pairwise_fst() function.
* Added the new Fst functions to the Ag3 and Af1 API docs.
* Added the sample_sets parameter to the average_fst() function.
* Added a couple more examples to the plot_pairwise_average_fst notebook to exercise some parameters.
* Ensured the sample_sets parameter is always provided in all notebooks to avoid any changes in behaviour when more data releases are added in future.
* Cleared output from all notebooks (generally do this for this repo to avoid having large notebook files in github).

Will merge here if CI is happy.

alimanfoo:
I've merged, thanks again Kelly, great to get this new functionality in! 🎉

Issue #424, opened 2023-10-06
CI failure, notebooks, h12 plot


leehart:
CI failing on notebooks, specifically `notebooks/plot_h12_h1x.ipynb` use of function `af1.plot_h12_gwss()`

> ValueError: No samples found for phasing analysis 'funestus' and query "cohort_admin2_year == 'GH-NP_Tolon-Kumbungu_fune_2017'"

leehart:
Closed by merged PR #426

PR #425, opened 2023-10-06
Update plot_h12_h1x.ipynb

leehart:
Re: issue #424

### DIFF
--- a/notebooks/plot_h12_h1x.ipynb
+++ b/notebooks/plot_h12_h1x.ipynb
@@ -197,7 +197,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


ahernank:
Yup. You can, of course, see those on the API but manually cross-checking the list on the cohort assignment NB for [the cohorts sync with Af1.0](https://github.com/malariagen/vobs-funestus/blob/master/tracking/release/v1.0/cohorts_20221129/assign_cohorts.ipynb) and finding the closest one to the ones we did as part of [Af1.1](https://github.com/malariagen/vobs-funestus/blob/master/tracking/release/cohorts_20230823/assign-cohorts.ipynb), might the quickest way to update these.

PR #426, opened 2023-10-06
Update Af1 cohort refs in notebooks, tests

leehart:
Re: issue #424


ahernank (approved):
Looks good to me, thanks @leehart!

Let’s add some notes to the docs in vector ops about running tests here while the cohorts data is in staging.

leehart:
Thanks @ahernank

Issue #427, opened 2023-10-09
Pin cohorts version in tests to avoid failures


leehart:
Re: PR #426

PR #428, opened 2023-10-09
Pin cohorts_analysis in tests, notebooks

leehart:
Re: issue #427


ahernank (approved):
Thanks @leehart, looks good to me. 

I was thinking whether the cohorts needed to be pinned anywhere else, and I could only come-up with the cases when we look at frequencies in a per-cohort basis (e.g. below from the freqs heatmap example [NB](https://github.com/malariagen/malariagen-data-python/blob/master/notebooks/plot_frequencies_heatmap.ipynb)); but maybe we want to tackle these ones another time, when/if needed.  

```
ag3.plot_frequencies_heatmap(
    df=cyp6aap_cnv_freqs_df[["label", "frq_BF-09_colu_2014"]],
)
```

leehart:
Thanks @ahernank . I've added that in here too, which was the only one I could find that wasn't pinned. Did you see any others?

ahernank:
Thanks @leehart, no, I didn't spot any others. I've taken a quick look, and agree, all the occurrences outside this NB seem to be pinned.

PR #429, opened 2023-10-10
Update README.md re tagged docs

leehart:


ahernank (approved):
Thanks @leehart, very exciting to see the new documentation! 

This looks great, I am just wondering if you are thinking of also updating the links on the README that currently point to the vector-data docs, in favour of the ones with your latest deployed docs on here.

leehart:
Thanks @ahernank . I think there might still be some more work to do on the docs before we make the switch.

PR #430, opened 2023-10-11
Add `study_id` & `study_url` to general metadata

### LINKED ISSUE #343: Add study to sample Ag3 metadata
It would be useful to have a "study" field with the MalariaGEN study ID in the sample metadata.

ahernank:
* Resolves #343 

Adding `study_id` & `study_url` to the general metadata df. 

I've also had to update some tests and the fixtures that use the results from sample_sets() as the manifest file also includes the study metadata. Also included a new function lookup_study(sample_set) for easiness.

### DIFF
--- a/tests/anoph/test_hap_data.py
+++ b/tests/anoph/test_hap_data.py
@@ -496,3 +496,24 @@ def test_haplotypes_with_max_cohort_size_param(
             sample_query=None,
             max_cohort_size=max_cohort_size,
         )
+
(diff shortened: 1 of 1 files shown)


leehart (approved):
Looks good to me. Thanks @ahernank

alimanfoo (approved):
Also looks good to me. I've updated this PR from master just to check if CI still runs OK.

PR #431, opened 2023-10-12
Delete init character on bash examples - `README.md` 

ahernank:
Deleted an init char in some of the bash commands (as I keep copying it!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ The `malariagen_data` Python package is available from the Python
 package index (PyPI) and can be installed via `pip`, e.g.:
 
 ```bash
-$ pip install malariagen-data
+pip install malariagen-data
 ```
 
 ## Documentation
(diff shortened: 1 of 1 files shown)


leehart:
The only remedy I know of is to satisfy the code coverage complaints by adding in more tests (specifically for lines `130` and `227` of `anoph/hap_data.py`, it seems) either here or in a another PR. But I don't know why or how this is happening. 😕

ahernank:
Yes, I spent some time struggling with a similar coverage failure on a ValueError expression on my other open PR. 
I thought it was a bit weird as other functions raising a KeyError, didn't need coverage on those lines. Ended-up adding the required lines to the test in 8614475456e03f90a2ae41870a0fe373b8707e51, which seemed to work. 

But, nevertheless, that kind of made sense if it was somehow related to some codecov background thing. I do not understand how this same issue here is related to hap_data when this one hasn't changed, and why this function specifically, as there are other functions which have uncovered lines with ValueErrors..

leehart (approved):
Thanks @ahernank

PR #433, opened 2023-10-16
Add test for sample sets with no phasing data

### LINKED ISSUE #432: Coverage of tests failing for `hap_data`
During some of the CI runs, the test coverage decreases for some of the lines (e.g. [here](https://app.codecov.io/gh/malariagen/malariagen-data-python/pull/431/indirect-changes)).

Alistair has pointed out this might be due to the selection of sample sets with no phasing data. As the runs pick-up random sample sets in some of the tests, this would explain why we don't see this issue consistently.

ahernank:
* Resolves #432


leehart (approved):
Thanks @ahernank

Issue #435, opened 2023-10-24
Add functions to access cohort group metadata


ahernank:
We have recently started generating metadata files about the cohort themselves, which include cohort size, country code, taxon, administrative units name, ISO code, geoBoundaries shape ID and representative latitude and longitude points & we want to be able to access these ones from the API (for each cohort category/name, i.e. "admin1_month", "admin1_quarter", "admin1_year","admin2_month", "admin2_quarter", "admin2_year")

Additionally, we also produce geojson files for each of the cohort categories (including geometries for each cohort), and we might want to consider accessing these too.

leehart:
@ahernank check status of geometries on each cohort.

kebarr:
All associated PRs merged, closing.

PR #436, opened 2023-10-24
Access cohort group metadata

ahernank:
* Partially addresses #435

Here we create a new class to host the cohort group metadata functions. The reasoning is to keep this apart from the usual sample_metadata (and the cohorts-related functions there) as those are all/can be accessed on a sample level; whereas these new functions are only accessible on a cohort level. 

I have also updated the default cohort analysis versions (and test files, params affected) as I think it is a good idea to keep a single source of cohort_analysis as of now, rather than allow the user to change the analysis on these functions, which could cause confusion with the rest of the functions. 

Here, we only tackle the csv metadata files. I attempted to also include the geojson but this is causing some issues related to large data files, and the geoJSON parsers required to access the file properly; so re-think/tackle in another PR.


tests/anopheles_test_data/test_missing_cohorts/ag/v3-config.json
```diff
@@ -5,5 +5,5 @@
     "GENESET_GFF3_PATH": "reference/genome/agamp4/Anopheles-gambiae-PEST_BASEFEATURES_AgamP4.12.gff3.gz",
     "DEFAULT_SPECIES_ANALYSIS": "aim_20220528",
     "DEFAULT_SITE_FILTERS_ANALYSIS": "dt_20200416",
-    "DEFAULT_COHORTS_ANALYSIS": "20220608"
+    "DEFAULT_COHORTS_ANALYSIS": "20230223"
```
alimanfoo, line 8: Sorry if I've forgotten, but why using 20230223 here rather than 20230516 as used elsewhere in this PR?

alimanfoo (commented):
They say the hardest thing in computer science is naming things :)

This looks good, I added some suggestions all based around simplifying terminology a little. 

Basically, we could reserve the word "metadata" to only be used in connection with sample metadata. So this means we could talk in this PR more simply about "cohort data", rather than "cohort group metadata". Suggestions for simplifying naming of classes and functions follow from this idea.

alimanfoo:
Hi @ahernank, codecov is complaining because [one line](https://app.codecov.io/gh/malariagen/malariagen-data-python/pull/436?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=malariagen#d9a5b70181668b9c5ce618b90587d63b-R29) of the code you've added in this PR isn't covered by any tests. Wouldn't hurt to test explicitly providing a cohorts analysis version. Probably easiest way would be to modify either the `ag3_sim_api` or `af1_sim_api` function to include a `cohorts_analysis` parameter.

tests/anopheles_test_data/test_missing_cohorts/ag/v3-config.json
ahernank: This one was a mistake, sorry. All updated.

malariagen_data/anoph/sample_metadata.py
```diff
@@ -164,37 +164,21 @@ def _cohorts_metadata_paths(self, *, sample_sets: List[str]) -> Dict[str, str]:
...
-                    "admin1_iso",
-                    "admin2_name",
-                    "taxon",
-                    "cohort_admin1_year",
-                    "cohort_admin1_month",
-                    "cohort_admin1_quarter",
-                    "cohort_admin2_year",
-                    "cohort_admin2_month",
-                    "cohort_admin2_quarter",
-                )
```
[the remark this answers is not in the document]
ahernank: I attempted to simplify this as we only have cohort (set) data for cohorts versions post 20230223. So, any version before this will not have data available. We could, of course, retrospectively generate the data for all cohorts we have but not sure how useful would this be. 

Agree this shouldn't mean that we should stop supporting the pre 20230223 cohorts! Will have to downgrade the analysis version in one of the fixture to ensure this gets tested, as I think the coverage is failing because I have updated all cohorts analysis across the API to post versions.

ahernank:
To solve the issues related to cohorts with quarters not having cohort data and ensure coverage is correct, I’ve added cohort data tests just using the Ag3 sim fixture. I’ve used cohorts without quarters for the Af1 simulator. This way we can avoid a new simulator.

alimanfoo:
Also remembering would be good to add the new `cohorts()` method to the API docs.

leehart (approved):

Issue #443, opened 2023-11-07
KeyError: 'Stamen'


alimanfoo:
As of [release 2023.10.1](https://github.com/geopandas/xyzservices/releases/tag/2023.10.1) xyzservices has removed stamen tiles.

This causes `KeyError: 'Stamen'` when trying to import malariagen_data.

Temporary workaround:

```
pip install malariagen_data "xyzservices < 2023.10.1"
```

alimanfoo:
Hi @leehart, any chance you could find a fix for this and make a micro release?

leehart:
> Hi @leehart, any chance you could find a fix for this and make a micro release?

Hi @alimanfoo . Sure, I'll do it first thing tomorrow.

alimanfoo:
Awesome, thanks @leehart. Sorry forgot you were away today.

leehart:
I had assumed from GeoPandas' statement `Use Stamen styles of Stadia provider instead` that these tiles would be available via Stadia instead, e.g.
```python
    "terrain": ipyleaflet.basemaps.Stadia.StamenTerrain,
    "watercolor": ipyleaflet.basemaps.Stadia.StamenWatercolor,
```
but judging from Stamen's FAQs and this test, it looks like we might have to drop our support for `watercolor` and `terrain` maps from our `map_params.py`, sadly
<img width="921" alt="Screenshot 2023-11-09 at 11 58 14" src="https://github.com/malariagen/malariagen-data-python/assets/4256466/a82e43a1-5945-4ce7-b8ee-efe9396c270a">

leehart:
From https://stadiamaps.com/stamen/
> I thought Stamen Map Tiles were free?
> For many years, Stamen offered a completely free map tile service. Unfortunately, mapping services are expensive to maintain, and it wasn’t possible to continue supporting an entirely free service. We partnered with Stamen to continue offering these iconic styles as part of the Stadia Maps’ platform.

leehart:
A new patch release has now been made (v7.14.1), but we probably want to guard against this fragility, somehow.

PR #445, opened 2023-11-09
Remove Stamen basemaps

leehart:
Re: issue #443


ahernank (approved):
Thanks @leehart! I couldn't immediately see any places where have been using these params, so seems sensible to drop support.

ahernank:
There might be other "terrain" sources if we wanted to regain this one in the future (e.g. https://github.com/geopandas/xyzservices/blob/5e9798aba5deafade06d11be573f5b1c1e993415/provider_sources/leaflet-providers-parsed.json#L1993) but looking through all of those and deciding which one is the closest to the Stamen one seems well-outside of this fix..

leehart:
Thanks @ahernank . Yes, there might be some other notebooks somewhere using these particular base maps tile sets, which will turn yucky due to Stamen withdrawing their free service, which is also affecting other middleware, e.g.: https://ipyleaflet.readthedocs.io/en/latest/map_and_basemaps/basemaps.html

...but the urgent problem is that it's preventing this package from being imported. We should probably look at avoiding this vulnerability, e.g. in case Esri also decide to withdraw their free service.
I will merge and make a patch release shortly to re-enable package import.

PR #441, opened 2023-11-03
Add option to use cohorts as colour or symbol in plot_haplotype_clustering

jonbrenas:
This seems to work on tests. Samples that do not belong to any cohort are shown on the plot (with a different colour or symbol, depending on the option selected) but do not appear in the legend. I don't know if that's the behaviour we want.

PR #450, opened 2023-11-27
Improve performance of haplotype clustering

### LINKED ISSUE #449: Speed up plot_haplotype_clustering()
The plot_haplotype_clustering() function is going pretty slow for larger numbers of samples. Because it involves a pairwise distance calculation it will scale roughly with the square of the number of samples, so some performance issues for larger numbers of samples is unavoidable. However, we might be able to improve the situation to some extent with a more efficient implementaiton.

Some thoughts:

* Noticing in particular that it looks like the pairwise distance calculation may be performed twice unnecessarily, once to find the max and min distances, and again to then compute the dendrogram. Can we avoid this?

* Any way to use a faster pairwise distance implementation? scikit-learn has an implementation which has thread parallelism.

* Would it help to copy haplotype data into fortran order first? Memory access can make a decent difference for pairwise distance calculations.

* Caching of the pairwise distance calculation would allow then faster replotting with different parameters like color, symbol or linkage_method.

* Find distinct haplotypes first, then only compute pairwise distance (and possibly clustering) with distinct haplotypes.

alimanfoo:
Resolves #449 via:

* [x] Only compute haplotype pairwise distances once.
* [x] Copy haplotype data to fortran memory layout.
* [x] Cache results from haplotype pairwise distances calculation.
* [x] Remove non-segregating sites before pairwise distance calculation.
* [x] Try optimising hamming pairwise distance with numba.

I also tried using scikit-learn's implementation of pairwise distances, but for hamming distance it reverts to using scipy anyway so performance is no better.

Also partially addresses #451 by adding `render_mode` parameter to the `plot_haplotype_clustering()` function, but I'll leave a full implementation of that for other plotting functions for another PR.

Also manually merges in changes from #441 (add option to use cohorts as colour or symbol) because things have been moved around quite a bit here within the `plot_haplotypes_clustering()` function.


alimanfoo:
For haplotype clustering over all samples in Ag3.0 at the Vgsc locus, this implementation [runs in 48s](https://colab.research.google.com/drive/1S422xMM2KwM1YiyhLLhFPStJEhscP4c_?usp=sharing) whereas the previous implementation [runs in 3m 14s](https://colab.research.google.com/drive/1CZbRzR-poEbqx0nr4lWIvSiq266AlPRd?usp=sharing).

alimanfoo:
Down to 28s with a optimised implementation of hamming pairwise distance.

alimanfoo:
For future reference, there is another possible optimisation which is to first find distinct haplotypes, then compute pairwise distances only between distinct haplotypes. I.e., it would be possible to skip the calculation for pairs of haplotypes that are identical. But that would introduce a fair amount of complexity, so exercise for the reader :)

alimanfoo:
Will merge here later if CI passes.

sanjaynagi:
Love it

PR #462, opened 2023-12-04
Add some spinners; tweak progress

### LINKED ISSUE #454: Add spinners for long-running computations where we can't show progress
Wherever possible we'd like to give some visual feedback to users when longer-running computations are occurring, so they can be reassured that computations are still running, understand a bit better what's happening behind the scenes, and begin to build some intuitions for how long things take.

We currently use tqdm to show progress wherever we are able to monitor progress of a longer-running computation. However, in some cases we cannot easily get information on the progress of a computation. Examples include computation of the SVD step within PCA, and computation of pairwise distances. In these cases, propose to show a spinner instead.

The [yaspin](https://github.com/pavdmyt/yaspin) package looks like a good candidate for this. 

We'd need to make sure it plays well with tqdm in notebooks, i.e., if a function has multiple tqdm progress bars and yaspin spinners for different parts of the internal computation, these would need to appear in the correct order.

### LINKED ISSUE #455: Do not leave tqdm progress bars
Currently tqdm progress bars remain visible after the progress of the computation has completed. However, this leaves output in the notebook, which can look ugly especially when a notebook is reopened later. 

The tqdm functions support a `leave` parameter which if set to `False` will remove the progress bar after it's completed. Suggest setting this to `False` so progress bars are removed after finished.

alimanfoo:
* Resolves #454 
* Resolves #455 

I'm sure there will be some other places where spinners could usefully be added to give more feedback on what is being computed, but this is a start.

### DIFF
--- a/docs/source/_static/switcher.json
+++ b/docs/source/_static/switcher.json
@@ -1,9 +1,14 @@
 [
     {
+        "name": "7.15.0",
+        "version": "v7.15.0",
+        "url": "https:///malariagen.github.io/malariagen-data-python/v7.15.0/",
+        "preferred": true
+    },
+    {
         "name": "7.14.1",
         "version": "v7.14.1",
-        "url": "https:///malariagen.github.io/malariagen-data-python/v7.14.1/",
-        "preferred": true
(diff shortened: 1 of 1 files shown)


leehart (approved):
🆒

PR #464, opened 2023-12-04
Improve performance applying site filters to SNP calls

### LINKED ISSUE #463: Slow performance calling snp_calls() with site_mask parameter
On colab, calling this:

```
ag3.snp_calls(region="3L", site_mask="gamb_colu_arab")
```

...spends more than 1m 30s in applying site filters.

Suspect this is because the site mask data is being read once for each variable in the dataset. 

Could be optimised to read only once.

alimanfoo:
* resolves #463

### DIFF
--- a/malariagen_data/util.py
+++ b/malariagen_data/util.py
@@ -215,6 +215,10 @@ def dask_compress_dataset(ds, indexer, dim):
     assert indexer.dtype == bool
     assert indexer.shape[0] == ds.dims[dim]
 
+    # compute the indexer once, to avoid multiple reads from the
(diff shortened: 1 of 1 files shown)


alimanfoo:
Down to ~10s - https://colab.research.google.com/drive/1RuJHkFTYXeDkrNuH60MfPynC1PBM8GEH?usp=sharing

leehart (approved):
👍

alimanfoo:
Advantage of spinners, when you start using them you start finding unexpected parts of the code that are running slower than you thought :)

PR #465, opened 2023-12-05
Improve performance of dendrogram plotting and saving results to the cache

### LINKED ISSUE #449: Speed up plot_haplotype_clustering()

### LINKED ISSUE #456: Improve performance of plotly dendrogram implementation
For large haplotype clustering problems, the current plotly dendrogram implementation doesn't perform particularly well. E.g., for ~10,000 haplotypes it can take more than a minute just to build the plot. I think this is because each set of three lines joining two clusters is added to the plot as a separate trace, so plotly spends a long time in the `update_traces` function. 

Instead it would be possible to plot all lines as a single trace. Some [rough benchmarking](https://colab.research.google.com/drive/1kUhp_XTUTrG_Kh2fo96N0zJtl2F_Td0p?usp=sharing) suggests this can execute in well under 1s.

### LINKED ISSUE #466: Slow performance saving results to cache
In benchmarking haplotype clustering with larger numbers of haplotypes I've found situations where it can take up to a minute to save the results of pairwise distance computation to the results cache. This seems to be entirely due to slow performance of numpy `savez_compressed()`. Using zarr `save()` instead, which is a drop-in replacement, runs much faster at around 1s.

alimanfoo:
This PR swaps out the implementation of plotly dendrogram which was borrowed from the web with a simpler and more performant implementation.

Also switches to using zarr to save data to the results cache to improve performance.

Also fix missing taxon colors for Ag3.

* Resolves #456
* Resolves #466

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -193,16 +193,20 @@ class Ag3(AnophelesDataResource):
 
         if plot_kwargs is None:
             plot_kwargs = dict()
-        taxon_palette = px.colors.qualitative.Plotly
+        taxon_palette = px.colors.qualitative.Vivid
         taxon_color_map = {
-            "gambiae": taxon_palette[0],
-            "coluzzii": taxon_palette[1],
+            "gambiae": taxon_palette[1],
+            "coluzzii": taxon_palette[0],
             "arabiensis": taxon_palette[2],
-            "gcx1": taxon_palette[3],
-            "gcx2": taxon_palette[4],
-            "gcx3": taxon_palette[5],
-            "intermediate_gambiae_coluzzii": taxon_palette[6],
-            "intermediate_arabiensis_gambiae": taxon_palette[7],
+            "merus": taxon_palette[3],
+            "melas": taxon_palette[4],
+            "quadriannulatus": taxon_palette[5],
+            "fontenillei": taxon_palette[6],
+            "gcx1": taxon_palette[7],
+            "gcx2": taxon_palette[8],
+            "gcx3": taxon_palette[9],
+            "gcx4": taxon_palette[10],
+            "unassigned": "black",
         }
         plot_kwargs.setdefault("color_discrete_map", taxon_color_map)
         plot_kwargs.setdefault(
--- a/malariagen_data/anopheles.py
+++ b/malariagen_data/anopheles.py
@@ -6428,8 +6428,9 @@ class AnophelesDataResource(
             Hierarchically cluster haplotypes in region and produce an interactive plot.
         """,
         parameters=dict(
-            scatter_y="Y coordinate at which to plot the scatter markers.",
-            kwargs="Passed through to `px.scatter()`.",
+            leaf_y="Y coordinate at which to plot the leaf markers.",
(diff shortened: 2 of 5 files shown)


malariagen_data/ag3.py
```diff
@@ -193,16 +193,20 @@ def _setup_taxon_colors(plot_kwargs=None):
 
         if plot_kwargs is None:
             plot_kwargs = dict()
-        taxon_palette = px.colors.qualitative.Plotly
+        taxon_palette = px.colors.qualitative.Vivid
```
alimanfoo, line 196: Need to switch to a slightly larger palette because we have more taxa now.

malariagen_data/anopheles.py
```diff
@@ -6428,7 +6428,9 @@ def _haplotype_pairwise_distances(
             Hierarchically cluster haplotypes in region and produce an interactive plot.
         """,
         parameters=dict(
-            kwargs="Passed through to `px.scatter()`.",
+            leaf_y="Y coordinate at which to plot the leaf markers.",
```
alimanfoo, line 6431: I decided to make a small visual change to the dendrogram to plot the leaves at Y coordinate zero by default, whereas previously it was -1. I did this because it creates a more visual connection between the dendrogram lines and the leaves. However, I made this a parameter so users can put the leaves back to -1 if they prefer.

malariagen_data/anopheles.py
```diff
@@ -6444,34 +6446,33 @@ def plot_haplotype_clustering(
...
         count_sort: hapclust_params.count_sort = True,
         distance_sort: hapclust_params.distance_sort = False,
-        width: plotly_params.width = 1000,
+        width: plotly_params.width = None,
         height: plotly_params.height = 500,
         show: plotly_params.show = True,
         renderer: plotly_params.renderer = None,
-        render_mode: plotly_params.render_mode = plotly_params.render_mode_default,
-        **kwargs,
+        render_mode: plotly_params.render_mode = "svg",
```
alimanfoo, line 6453: I set the default render mode as svg here rather than auto because I tried webgl and it doesn't seem to behave any better in the browser, in fact it seems worse, and also the hover labels for the leaves don't work.

alimanfoo (commented):
A couple of implementation notes...

PR #468, opened 2023-12-06
Biallelic SNP calls and diplotypes

### LINKED ISSUE #448: Biallelic SNP calls
Several popgen analyses either require or are easier to implement with biallelic SNPs only. E.g., our current pca() function locates biallelic SNPs first before running the PCA. Anticipating several other functions which will want to locate biallelic SNPs in the same way, I propose to pull out a function to do this. 

Proposed signature:

```
def snp_calls_biallelic(
    region,
    sample_sets,
    sample_query,
    site_mask,
    max_missing_an,
    min_minor_ac,
    n_snps,
    thin_offset,
) -> xr.Dataset:
```

The `max_missing_an`, `min_minor_ac`, `n_snps` and `thin_offset` parameters would have the same semantics as parameters in the current pca() function.

The return value would be an xarray Dataset with the following variables:

* sample_id
* variant_contig
* variant_position
* variant_allele
* call_genotype

The data would be transformed to provide the simplest representation of biallelic SNPs, so the variant_allele array would have 2 columns, and the maximum value in the call_genotype array would be 1.

alimanfoo:
* Resolves #448 

Adds a new method `biallelic_snp_calls()` which builds a dataset of SNP calls at sites which are biallelic within the selected samples. Note these do not necessarily have to include the reference allele, i.e., sites are included if there are only two alleles observed, regardles of which alleles.

Adds a new method `biallelic_diplotypes()` which computes alternate allele counts per genotype call, generating a 2-dimensional array suitable for use with PCA, neighbour-joining trees, admixture, etc.

Refactors `pca()` to use `biallelic_diplotypes()` internally.

N.B., because the set of input SNPs to PCA will be slightly different as a result of this change, I have bumped the results cache version number for the pca() function. Also the next release of malariagen_data should be a major version bump I would suggest - the API has not changed so code will still be backwards compatible, but analysis results could change slightly and so it would probably help to communicate that.

### DIFF
--- a/malariagen_data/anoph/base.py
+++ b/malariagen_data/anoph/base.py
@@ -1,4 +1,5 @@
 import json
+from contextlib import nullcontext
 from pathlib import Path
 from typing import (
     IO,
@@ -81,7 +82,7 @@ class AnophelesBase:
             bokeh.io.output_notebook(hide_banner=True)
 
         # Check colab location is in the US.
-        if check_location and self._gcs_url is not None:
+        if check_location and self._gcs_url is not None:  # pragma: no cover
             self._client_details = check_colab_location(
                 gcs_url=self._gcs_url, url=self._url
             )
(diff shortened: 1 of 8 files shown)


alimanfoo:
Going to merge but happy to follow up if any questions or suggestions.

PR #483, opened 2023-12-16
Refactor Anopheles IGV

alimanfoo:
Continue towards #366 with IGV functions.


alimanfoo:
Test failure because wgs data catalog not always sorted the same as general metadata.

PR #490, opened 2023-12-20
Refactor SNP frequencies

alimanfoo:
Continuing on the journey towards #366 here we refactor functions related to SNP allele frequencies analysis.

Also fix a bug in `aa_allele_frequencies()` where frequencies would be reported as 0 even where all genotypes are missing, should be nan instead.


alimanfoo:
Alright, think I'm just about done here. Planning to merge if CI looks good.

PR #493, opened 2024-01-05
Improve G123 and H12 performance

alimanfoo:
Here we speed up G123 and H12 computations by using a faster approach to hashing.


alimanfoo:
N.b. G123 is still relatively slow as most of the time is in retrieval of genotype data, but this does still make a difference.

PR #500, opened 2024-01-11
Add the correct nomenclature for genes in anopheles.py to enable CNVs for Af

### LINKED ISSUE #499: Different nomenclature in genome_features between Ag and Af
Genome features for Ag come from VectorBase and the ones for Af come from VEuPathDB and the two databases use slightly different nomenclatures for the genome features. Inn particular, VectorBase has a feature called 'gene' while VEuPathDB doesn't: it uses 'protein_coding_gene' instead. This is significative because (among other things), the function _gene_cnv (which is part of anopheles.py and thus shared between Ag and Af) looks for 'gene' features and can't find any in the genome feature data frame. Hence, it fails completely.

jonbrenas:
Will resolve #499.

I don't know yet if there are other instances where this problem occurs. I will add a test_gene_cnv for Af mirroring the one for Ag.


alimanfoo:
Thanks @jonbrenas. I did a search through the code base and think this the only place where this occurs, so I think you can say that this PR will resolve #499.

alimanfoo (commented):
Hi @jonbrenas, looking good.

Very small nit, would you mind calling it "gff_gene_name_attribute" instead of "gff_gene_name", just to help clarify.

jonbrenas:
Thanks @alimanfoo , hopefully, I made the change.

alimanfoo (approved):
Thanks @jonbrenas, LGTM

PR #501, opened 2024-01-11
Reduce memory usage when accessing SNP calls with site_class parameter

### LINKED ISSUE #498: RAM use for Af diversity_stats
For some reason, running af.diversity_stats uses an enormous amount of RAM (at least on Colab and Datalab). It crashes even my Pro sessions with 12GB of RAM. I tried:
af1.diversity_stats(
    cohorts="cohort_admin2_year",
    cohort_size=10,
    region="3RL:1,000,000-2,000,000",
    sample_query="country == 'Gabon'",
    site_mask='funestus',
    site_class="CDS_DEG_4"
)
and the same for Ghana (resp. 40 samples in one cohort and 31+29 in 2 cohorts) that fail in the same way. 

Trying the same thing with Ag works fine (though, it also crashes due to RAM use if 'site_mask' is dropped). I tested on Gabon with 5 cohorts including 3 that have more than 60 samples.

alimanfoo:
* Resolves #498


alimanfoo:
There are two separate optimisations here.

The first change is within the _snp_calls() method and reduces memory usage when requesting a relatively small genome region:

image

The second change is within the _locate_site_class() method and reduces memory usage even when accessing a large genome region:

image

malariagen_data/anoph/snp_data.py
```diff
@@ -834,6 +846,18 @@ def _locate_site_class(
...
+            # therefore subset to SNP positions.
+            pos = self.snp_sites(
+                region=region,
+                field="POS",
+                site_mask=site_mask,
+                inline_array=inline_array,
+                chunks=chunks,
+            )
+            idx = (pos - 1).compute()
+            loc_ann = da.take(loc_ann, idx, axis=0)
```
alimanfoo, line 859: This is a key move, we delay the indexing of the site annotations data to SNP positions until later on. This avoids the indexing array (idx) getting copied multiple times.

PR #506, opened 2024-02-18
Add new discordant read calls version

ahernank:
Updating Ag3 to use the new analysis version (`20230911`) of the discordant read calls. 

Similar update for Af1  for consistency with an empty version (although we currently do not have any released discordant read calls for Af). 

Note I'm suggesting to specify the version directly on the class (rather than using the config file) as, this version supersedes the previous one in Ag3 (and includes the previous sites) and to avoid including a default parameter on the Af1 config (as there are no discordant read calls yet).


alimanfoo (commented):
Hi @ahernank, just took another look, think this needs a few changes. Basically there is a confusion which I have introduced around what is an "analysis". There are in fact two different types of parameter, both of which are named "something_analysis" in the API but which are handled differently. Probably easier to explain in person :)

ahernank:
Thanks @alimanfoo! Will put this one on hold until we catch-up tomorrow during the data meeting.

alimanfoo (commented):
Hi @ahernank, the structure of the code looks right now, matches the other analysis version parameters. But there is a tricky problem here, because there is still no way to override and force the analysis version back to the original (previous) version of the data. The reason for this is that if a user passes either None or an empty string as the value of the discordant_read_calls_analysis parameter, then the value will evaluate to False when tested within the `_discordant_read_calls_analysis()` property getter method, causing the value from the config to be returned.

Will have a think about how to deal with this...

malariagen_data/anoph/cnv_data.py
```diff
@@ -45,6 +50,15 @@ def __init__(
         self._cache_cnv_coverage_calls: Dict = dict()
         self._cache_cnv_discordant_read_calls: Dict = dict()
 
+    @property
+    def _discordant_read_calls_analysis(self) -> Optional[str]:
+        if self._discordant_read_calls_analysis_override:
```
alimanfoo, line 55:
```suggestion
        if isinstance(self._discordant_read_calls_analysis_override, str):
```

Suggest we use `None` to indicate that the analysis version from the config should be used, and the empty string `""` to indicate an override that the original analysis version should be used. 

So here we check to see if the user provided a string, and if so, return it.

malariagen_data/anoph/cnv_data.py
```diff
@@ -399,7 +413,13 @@ def open_cnv_discordant_read_calls(
         except KeyError:
             release = self.lookup_release(sample_set=sample_set)
             release_path = self._release_to_path(release)
-            path = f"{self._base_path}/{release_path}/cnv/{sample_set}/discordant_read_calls/zarr"
+            analysis = self._discordant_read_calls_analysis
+            if analysis:
+                calls_version = f"discordant_read_calls_{analysis}"
+            else:
+                calls_version = "discordant_read_calls"
```
alimanfoo, line 420: If we use the empty string `""` to indicate the original analysis version, then this should work as desired. I.e., the empty string will evaluate to False.

alimanfoo (commented):
Here's a suggested way to enable the user to override and force the analysis version back to the original version (which has no version identifier).

Suggest we use `None` to indicate that the analysis version from the config should be used, and the empty string `""` to indicate an override that the original analysis version should be used.

ahernank:
Thanks @alimanfoo! We spotted this issue with `test_allele_frequencies_with_region` before in different regions in X.

```
tests/anoph/test_snp_frq.py::test_allele_frequencies_with_region[af1_sim] - ValueError: The genomic region 'X:117,109-117,609' is invalid for contig 'X' with length 117228.
```

I am trying to debug why this is happening, but it isn't consistent so far (i.e. doesn't seem to be related to specific regions or contigs/not failing consistently). 

We can re-run this test to pick another region but trying to see if I can pick the mistake , @leehart, do you happen to have any ideas?

leehart:
@ahernank The error message seems to be saying that contig X is only 117,228 positions long, so it can't give a region that goes up to position 117,609.

I don't know which is right: Does X have 117,228 positions or 117,609 positions? Somewhere there seems to be a disagreement, and I suspect the tests need to be modified to align with reality. Maybe the number of positions available has changed since the test was written, or maybe it's based on some other variable, I'm not sure.

ahernank:
Thanks @leehart. This was my same thought but the (full) X contig is much longer (24,393,108). This test doesn't seem to use that value, but rather a simulated one. 

Which I don't quite understand is why this simulated value changes every time (not in line with the region per se), so sometimes the same test on the same region fails or passes.

leehart:
Hi @ahernank . I just had a chat with Alistair and it looks like we just need to fix the `random_region_str` function so that it takes the `contig_size` into account. Basically, it's coming up with a random 500 position region and not being smart enough to make sure that doesn't go over the end of the synthetic contig.

https://github.com/malariagen/malariagen-data-python/blob/18bf2f354fdc22982ec234927d363d7fe66fba48/tests/anoph/conftest.py#L1014

Do you want me to try to fix this on this branch, or another branch, or do you want to try?

leehart:
Plan to fix random_region_str on a different PR.

Checks now pass due to random nature of tests, i.e. random test region falls within synthetic contig.

leehart (approved):

PR #511, opened 2024-03-23
Authenticate the user if accessing GCS

### LINKED ISSUE #510: Support authenticated access to GCS
It's possible we may have to move to GCS buckets that require authenticated access, to address challenges with some users causing excessive download (network) costs. Add support in the API for authenticated access to GCS.

alimanfoo:
* Resolves #510


alimanfoo:
Closing in favour of #513 because we can't authenticate github actions from a fork.

PR #513, opened 2024-03-24
Authenticate the user if accessing GCS

### LINKED ISSUE #510: Support authenticated access to GCS

alimanfoo:
* Resolves #510

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,9 +13,13 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
     runs-on: ubuntu-latest
+    # Add "id-token" with the intended permissions.
+    permissions:
+      contents: 'read'
+      id-token: 'write'
     steps:
       - name: Checkout source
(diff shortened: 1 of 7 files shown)


alimanfoo:
Hi @leehart, @ahernank, @cclarkson, this PR is ready for review. Happy to talk through in more detail. Basically once this is merged and included in a new release of malariagen_data, then anyone using malariagen_data inside google colab will be automatically authenticated. Anyone using it outside of google colab will need to run an authentication step. 

This is something we can implement **before** making any changes to access policies on the actual storage. I.e., we can get users used to logging in first, and make sure the machinery is in place to do it. This means that we can then make changes to access policies on storage if needed.

leehart (approved):

ahernank (approved):

alimanfoo:
Would be good to add some documentation about authentication, both for users and for developers.

PR #515, opened 2024-03-26
Add support for writing Anopheles SNP data to the plink binary file format

### LINKED ISSUE #248: Write SNP calls to plink binary file
Add support for writing a selection of SNP call data to a plink binary format file. Useful to export data for analysis with tools like ADMIXTURE.

Prototype implementation [here](https://colab.research.google.com/drive/1E95Dn-epsYfm3Krg6Okl7gpPfAN_crf7?usp=sharing).

tristanpwdennis:
This PR adds a new function `biallelic_snps_to_plink()` to the `Ag3` and `Af1` APIs which allows for selection of a set of biallelic SNPs for export to the [plink binary file format](https://www.cog-genomics.org/plink/1.9/formats#bed). This then allows for data to be imported into tools like ADMIXTURE which can read the plink format.

Resolves #248.


sanjaynagi:
Hey Tristan. Nice work! 

Ill save comments for now but FYI - when you add notebooks to malariagen_data, make sure you have cleared all outputs, otherwise they can become quite hefty in size and then the repo balloons in size over time (all of it is stored in git history).

tristanpwdennis:
I've found the source of the AssertionError (also see issue #516) - something to do with how ```dask.array.map_blocks``` computes ```variant_allele``` at line 1629 of ```snp_data.py```. 

I haven't managed to get to the bottom of it yet but in this PR there's a temporary fix that just applies ```apply_allele_mapping``` to an in-memory np array of ```variant_allele```, and I've now added ```biallelic_snp_calls``` to ```to_plink.py``` instead of calling ```snp_calls``` and thinning them manually.

tristanpwdennis:
I've (I hope!) made a fix to the above error (issue #516), and described it in more detail there

jonbrenas (changes requested):
Hi @tristanpwdennis. Before we merge this PR, could I ask you to do some clean-up? 
`test-1.ipynb` needs to be removed and you made some changes to `.gitignore` and `test_snp_data.py`. There are also quite a few `print` commands that need to be removed or switched to debug mode.
Could you also add a test that checks (at least) that the file is created?

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,182 @@
...
+            random_seed=random_seed,
+            inline_array=inline_array,
+            chunks=chunks,
+            n_snps=n_snps,
+        )
+
+        # Filter SNPs for segregating sites only
+        with self._spinner("Subsetting to segregating sites"):
+            gt = ds_snps["call_genotype"].data.compute()
+            print("count alleles")
```
jonbrenas, line 84: Can you remove this `print` statement?

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,182 @@
...
+            with ProgressBar():
+                ac = allel.GenotypeArray(gt).count_alleles(max_allele=3)
+                print("ascertain segregating  sites")
```
jonbrenas, line 87: Same thing here.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,182 @@
...
+                an_called = ac.sum(axis=1)
+                an_missing = n_chroms - an_called
+                min_ref_ac = min_minor_ac
+                max_ref_ac = n_chroms - min_minor_ac
+                loc_sites = (
+                    (ac[:, 0] >= min_ref_ac)
+                    & (ac[:, 0] <= max_ref_ac)
+                    & (an_missing <= max_missing_an)
+                )
+                print(f"ascertained {np.count_nonzero(loc_sites):,} sites")
```
jonbrenas, line 98: Same thing here.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Set up dataset with required vars for plink conversion
+        print("Set up dataset")
```
jonbrenas, line 101: Same thing here.

tests/anoph/test_snp_data.py
```diff
@@ -88,6 +88,7 @@ def test_open_snp_sites(fixture, api: AnophelesSnpData):
         assert "variants" in contig_grp
         variants = contig_grp["variants"]
         assert "POS" in variants
+        assert False
```
jonbrenas, line 91: Not sure what this is here for.

tristanpwdennis:
Hi Jon, 
Thanks! Will tidy further and add the test. Cheers :)

tristanpwdennis:
Hi @jonbrenas, I had a tidy and removed some redundant code from to_plink.py. I also added a test (test_plink_converter.py) to make sure the files are created. Let me know if this is sufficient or if you want me to add any more. Hope this works ok!
Thanks
-t

jonbrenas:
Great job @tristanpwdennis !

I may be misunderstanding the way the PLINK format work but shouldn't it work with any biallelic site and not only with those where the alleles are ref and 1st alt. Is there a reason why, you select only those? 

Also, bed_reader needs to be added to the list of packages installed before it is imported. This can be done by modifying the `pyproject.toml` file.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,167 @@
...
+        # Set up dataset with required vars for plink conversion
+        ds_snps_asc = ds_snps[
+            [
+                "variant_contig",
+                "variant_position",
+                "variant_allele",
+                "sample_id",
+                "call_genotype",
+            ]
+        ].isel(alleles=slice(0, 2))  # .sel(variants=ix_sites_thinned)
```
jonbrenas, line 93: The comment can probably be deleted.

tests/anoph/test_plink_converter.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # Check to see if bed, bim, fam output files exist
+    file_path = f"{tmp_path}/{plink_params['region']}.{plink_params['n_snps']}.{plink_params['min_minor_ac']}.{plink_params['thin_offset']}.{plink_params['max_missing_an']}"
+
+    if os.path.exists(f"{file_path}.bed"):
+        pass
+    if os.path.exists(f"{file_path}.bim"):
+        pass
+    if os.path.exists(f"{file_path}.fam"):
+        pass
```
jonbrenas, line 100: These should probably be `assert os.path.exists(...)`. I don't think these tests could fail even if the fonction didn't create the correct files.

jonbrenas:
It might also be a good idea to have a check that the data in the files is correct for "dummy" data. For example, looking at only the first 5 samples of 'AG1000G-AO' and the region 'X:10_000_000-10_000_500', I find only 3 biallelic sites. With such small data, it should be fairly simple to figure out what the PLINK files should be and check that it is correct. You could, obviously, use any other of "dummy" data that you like.

tristanpwdennis:
Hi Jon

Good spot! Selecting the REF and 1st ALT is a relic from before I used .biallelic_snp_calls and had to manually modify the alleles dimension. It didn't change anything being there (e.g. it was selecting the first two alleles of a dataset that already only had two alleles), but it's redundant now and I've removed it.

I've updated the tests to include the feedback above, making sure that the exported dummy data match the data pre-export, both in terms of dimensions and actual content. Please take a look and let me know if there's anything else you think I can add.

Apologies also for the long time taken and revisions needed - you can probably tell that I am quite new to this. Thanks for your feedback and help!


-t

jonbrenas:
Thank you very @tristanpwdennis. It passes the eye-test so I merged it into a shadow PR to see whether it passes the various tests. It hasn't but I haven't checked yet whether there is something big that is going wrong or if it's the result of small conflicts between recent code. I'll come back to you if there is something significant in the code that needs to be addressed.

jonbrenas (dismissed):

jonbrenas:
Hi @tristanpwdennis . Sorry, I had misinterpreted the error. `base_params.chunks_default` has been made obsolete in the main repository since you created your branch and I thought it was a parameter that you had added. I have corrected it. I will come back to you if there is anything else that needs your input. Thank you for the hard work!

malariagen_data/anoph/snp_data.py
```diff
@@ -1669,9 +1669,23 @@ def biallelic_snp_calls(
...
+            )
+
+            # Apply allele mapping blockwise to variant_allele.
+            variant_allele_out = da.map_blocks(
+                lambda allele, map: apply_allele_mapping(allele, map, max_allele=1),
+                variant_allele,
+                allele_mapping_chunked,
+                dtype=variant_allele.dtype,
+                chunks=(variant_allele.chunks[0], [2]),
+
```
alimanfoo, line 1688: Hi @tristanpwdennis, these changes are no longer necessary as the bug has been fixed in master. The helper function `dask_apply_allele_mapping()` handles the necessary transformation. 

My git-fu is not perfect but there *should* be a simple way to revert these changes so `snp_data.py` matches origin/master. Perhaps something like `git checkout origin/master -- malariagen_data/anoph/snp_data.py` might work?

tristanpwdennis: Hi @alimanfoo - sorry - in case you and @jonbrenas can tell I've been struggling a bit here (lack of git-fu, though that website has been helpful). This didn't seem to work as far as I can tell - so any suggestions welcome, thanks! I'll keep looking...

jonbrenas: Hi @tristanpwdennis, @alimanfoo, I am no expert at git-fu either but I think the easiest solution is to make the change in the shadow PR (where all tests were successful except for the new notebook that uses a local path). As a matter of fact, I did it without any real issue. I would say that, at this point, you have accomplished the job @tristanpwdennis (congrats and thank you, by the way) and all that is left is some light clean-up. @alimanfoo, do you object to considering this PR as a success and closing it (with the option to do more fine-tuning if needed in the shadow PR)?

alimanfoo: Thanks so much @jonbrenas and @tristanpwdennis 🙏🏻. Having a quick look over this PR there's a couple of small things I noticed, probably worth iterating a bit more here before calling it done. I'll add some comments tomorrow.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,167 @@
...
+                gn_ref = gn_ref.compute()
+
+        # Ensure genotypes vary
+        loc_var = np.any(gn_ref != gn_ref[:, 0, np.newaxis], axis=1)
+
+        # Load final data
+        with ProgressBar():
+            ds_snps_final = ds_snps_asc[
+                ["variant_contig", "variant_position", "variant_allele", "sample_id"]
+            ].isel(variants=loc_var)
```
alimanfoo, line 109:
```suggestion
        # Load final data
        ds_snps_final = dask_compress_dataset(ds_snps_asc, loc_var, dim="variants")
```

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,167 @@
...
+        gn_ref_final = gn_ref[loc_var]
+        val = gn_ref_final.T
+        alleles = ds_snps_final["variant_allele"].values
+        properties = {
+            "iid": ds_snps_final["sample_id"].values,
+            "chromosome": ds_snps_final["variant_contig"].values,
+            "bp_position": ds_snps_final["variant_position"].values,
+            "allele_1": alleles[:, 0],
+            "allele_2": alleles[:, 1],
+        }
```
alimanfoo, line 121: Could consider using a spinner...

```suggestion
        with self._spinner("Prepare output data"):
            alleles = ds_snps_final["variant_allele"].values
            properties = {
                "iid": ds_snps_final["sample_id"].values,
                "chromosome": ds_snps_final["variant_contig"].values,
                "bp_position": ds_snps_final["variant_position"].values,
                "allele_1": alleles[:, 0],
                "allele_2": alleles[:, 1],
            }
```

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,167 @@
...
+
+        bed_reader.to_bed(
+            filepath=bed_file_path,
+            val=val,
+            properties=properties,
+            count_A1=True,
+        )
+
+        print(f"PLINK files written to to: {plink_file_path}")
```
alimanfoo, line 130: Remove print statement.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,167 @@
...
+
+        return plink_file_path
+
+    def biallelic_snps_to_plink(
+        self,
+        results_dir,
```
alimanfoo, line 136: Nit, consider `output_dir`?

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,167 @@
...
+        region: base_params.regions,
+        n_snps: base_params.n_snps,
+        thin_offset: base_params.thin_offset = 0,
+        sample_sets: Optional[base_params.sample_sets] = None,
+        sample_query: Optional[base_params.sample_query] = None,
+        sample_indices: Optional[base_params.sample_indices] = None,
+        site_mask: Optional[base_params.site_mask] = None,
+        min_minor_ac: Optional[base_params.min_minor_ac] = 0,
+        max_missing_an: Optional[base_params.max_missing_an] = 0,
```
alimanfoo, line 145: Consider using same defaults as PCA and NJT here?

tests/anoph/test_plink_converter.py
```diff
@@ -0,0 +1,136 @@
...
+        **data_params,
+        n_snps=n_snps,
+    )
+
+    # Test to make sure that the rows and columns (no. variants and no. samples) of the .bed file match.
+    assert bed.shape[1] == ds_test.variant_position.shape[0]
+    assert bed.shape[0] == ds_test.samples.shape[0]
+
+    # Test to see if sample_id is exported correctly (stored in the .fam file).
+    assert set(bed.iid) == set(ds_test.sample_id.values)
```
alimanfoo, line 126:
```suggestion
    # Test to see if sample_id is exported correctly (stored in the .fam file).
    assert_array_equal(bed.iid, ds_test.sample_id.values)
```

alimanfoo (commented):
Looks awesome! A few minor suggestions...

tests/anoph/test_plink_converter.py
```diff
@@ -0,0 +1,136 @@
...
+    assert os.path.exists(f"{file_path}.fam")
+
+    # Read bed, bim, and fam files (bed_reader searches for the .bim and.fam files matching the prefix of the .bed file).
+    bed = bed_reader.open_bed(f"{file_path}.bed")
+
+    # Load a ds containing the same data exported to PLINK to test against.
+    ds_test = api.biallelic_snp_calls(
+        **data_params,
+        n_snps=n_snps,
+    )
```
alimanfoo, line 119: It is possible that the dataset could be slightly different from what is written out by the plink converter, because the plink converter also checks for and removes any rows with all identical genotype calls. But we may not encounter that with the test dataset, so perhaps OK to ignore.

malariagen_data/anoph/to_plink.py
alimanfoo: The function `dask_compress_dataset()` can be imported from the `util` module, and provides an optimised implementation of selecting from a dataset using a boolean indexer.

leehart:
Re: issue #248

tristanpwdennis:
Hi @jonbrenas - thank you for offering! Sorry for the delay, been a bit tied up the last few weeks but I think I can get to it over the weekend - I'll let you know if I run into any issues but looks straightforward enough!

alimanfoo:
Thanks so much @tristanpwdennis! 

I took the liberty to resolve merge conflicts and add you as a project author.

I'll take a closer look ASAP.

alimanfoo:
Also took the liberty to run `poetry update` to make sure the new dependencies are included in the poetry lock file.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,164 @@
+from typing import Optional
+
+import allel  # type: ignore
+import numpy as np
+import os
+import bed_reader
+
+from util import dask_compress_dataset
```
alimanfoo, line 8:
```suggestion
from ..util import dask_compress_dataset
```

Fix relative import.

alimanfoo:
Also took the liberty to [fix a relative import](https://github.com/malariagen/malariagen-data-python/pull/515#pullrequestreview-2445355096).

jonbrenas:
Thank you @tristanpwdennis for the quick reaction. It looks we are almost there. The same parameter name change ( `results_dir` should be `output_dir`) is creating a problem in the notebook. If you can solve that, everything should be a-OK.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,164 @@
...
+        self,
+        *,
+        output_dir,
+        region,
+        n_snps,
+        min_minor_ac,
+        thin_offset,
+        max_missing_an,
+    ):
+        return f"{output_dir}/{region}.{n_snps}.{min_minor_ac}.{thin_offset}.{max_missing_an}"
```
alimanfoo, line 36:
```suggestion
        return f"{output_dir}/{region}.{n_snps}.{min_minor_ac}.{max_missing_an}.{thin_offset}"
```

This is a very small suggestion, but I wonder if it's a little more intuitive if the `thin_offset` component of the path name comes at the end. This is because in the typical use case for an analysis like ADMIXTURE the user might want to create multiple replicate datasets with different sets of SNPs, and the way to do that here is to vary the `thin_offset`. I.e., a user might call the same function with `thin_offset` values of 0, 1, 2, etc. If this is the last path component then perhaps easier to see these are replicates?

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,164 @@
...
+        sample_sets,
+        sample_query,
+        sample_indices,
+        site_mask,
+        min_minor_ac,
+        max_missing_an,
+        random_seed,
+        inline_array,
+        chunks,
+    ):
```
alimanfoo, line 55: I'm wondering if it's worth having the code here in a separate private function, or whether actually it would be simpler to have the implementation directly within the public `biallelic_snps_to_plink()` function. There is actually nothing at all happening within the public `biallelic_snps_to_plink()` function so we are not gaining anything by splitting out the implementation into a separate private function.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,164 @@
...
+        bed_reader.to_bed(
+            filepath=bed_file_path,
+            val=val,
+            properties=properties,
+            count_A1=True,
+        )
+
+        return plink_file_path
+
+    def biallelic_snps_to_plink(
```
alimanfoo, line 125: Please could you add a docstring to this function. We usually use the `@doc` annotation from the `numpydoc_decorator` package to generate the docstrings, if you look at some other modules there should be some examples to follow.

alimanfoo: Once you have added the docstring, please could you add this function to the API docs at `docs/source/Ag3.rst` and `docs/source/Af1.rst` so that the new function will appear in the [API docs website](https://malariagen.github.io/malariagen-data-python/latest/). It would probably make sense to add the new function under the "SNP data access" section.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,164 @@
...
+        self,
+        output_dir,
+        region: base_params.regions,
+        n_snps: base_params.n_snps,
+        thin_offset: base_params.thin_offset = 0,
+        overwrite=False,
```
alimanfoo, line 131: This parameter needs a type annotation. 

You could just provide a plain type annotation here, e.g.:

```
        overwrite: bool=False,
```

...but if so you'll need to document the parameter within the docstring.

Alternatively you could declare a new type alias within the `base_params` module and document it there, then use it here like:

```
        overwrite: base_params.overwrite=False,
```

tests/anoph/test_plink_converter.py
```diff
@@ -0,0 +1,138 @@
...
+    n_snps = random.randint(1, n_snps_available)
+
+    # Define plink params.
+    plink_params = dict(output_dir=tmp_path, n_snps=n_snps, **data_params)
+
+    # Make the plink files.
+    api.biallelic_snps_to_plink(**plink_params)
+
+    # Test to see if bed, bim, fam output files exist.
+    file_path = f"{tmp_path}/{plink_params['region']}.{plink_params['n_snps']}.{plink_params['min_minor_ac']}.{plink_params['thin_offset']}.{plink_params['max_missing_an']}"
```
alimanfoo, line 108: Will need to change this if accept the suggestion above to put the thin_offset at the end.

alimanfoo (commented):
Hi @tristanpwdennis, all checks are passing in the shadow PR so we're looking great here and almost ready to merge!

Just a couple more things here to add documentation for the new function...

alimanfoo:
Btw have taken the liberty to edit the PR title and description just so this PR will look good in the release notes :)

malariagen_data/anoph/to_plink.py
tristanpwdennis: Legacy from when I copied this module from (I think) the structure of the PCA module? Amended now.

tristanpwdennis:
Addressed comments and feeling the urge to merge

malariagen_data/anoph/plink_params.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    A boolean indicating whether a previously written file with the same name ought
+    to be overwritten. Default is False.
+    """,
+]
+
+output_dir: TypeAlias = Annotated[
+    str,
+    """
+    A string indicating the desired output file location
```
alimanfoo, line 16:
```suggestion
    A string indicating the desired output file location.
```

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,129 @@
...
+        summary="""
+            Write Anopheles biallelic SNP data to the Plink binary file format.
+        """,
+        extended_summary="""
+            This function writes biallelic SNPs to the Plink binary file format. It enables
+            subsetting to specific regions (`region`), selecting specific sample sets, or lists of
+            samples, randomly downsampling sites, and specifying filters based on missing data and
+            minimum minor allele count (see the docs for `biallelic_snp_calls` for more information).
+            The `overwrite` parameter, set to true, will enable overwrite of data with the same
+            hash values (e.g. with a different randomly selected number of sites).
```
alimanfoo, line 38:
```suggestion
            SNP selection parameter values.
```

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,129 @@
...
+        """,
+        returns=("Binary Plink output files: .bed, .bim, .fam"),
+        notes="""
+            This computation may take some time to run, depending on your computing
+            environment. Unless the `overwrite` parameter is set to `True`, results will be loaded
+            from a previous computation, if performed, and if the `results_cache` parameter was
+            set when instantiating the API client.
```
alimanfoo, line 45:
```suggestion
            This computation may take some time to run, depending on your computing
            environment. Unless the `overwrite` parameter is set to `True`, results will be returned
            from a previous computation, if available.
```

Actually results will be re-used regardless of whether the `results_cache` parameter was set. I.e., this function doesn't make use of the `results_cache` machinery, it handles it's own kind of caching.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,129 @@
...
+        self,
+        output_dir: plink_params.output_dir,
+        region: base_params.regions,
+        n_snps: base_params.n_snps,
+        overwrite: plink_params.overwrite = False,
+        thin_offset: base_params.thin_offset = 0,
+        sample_sets: Optional[base_params.sample_sets] = None,
+        sample_query: Optional[base_params.sample_query] = None,
+        sample_indices: Optional[base_params.sample_indices] = None,
+        site_mask: Optional[base_params.site_mask] = None,
```
alimanfoo, line 58:
```suggestion
        site_mask: Optional[base_params.site_mask] = base_params.DEFAULT,
```

Probably better to use the default site mask by default, rather than no site mask, given that this function will mostly be used to output data for population structure analyses, and having high quality SNPs is generally best for that. I.e., here we match default parameter values to the pca() function.

malariagen_data/anoph/to_plink.py
```diff
@@ -0,0 +1,129 @@
...
+        """,
+        extended_summary="""
+            This function writes biallelic SNPs to the Plink binary file format. It enables
+            subsetting to specific regions (`region`), selecting specific sample sets, or lists of
+            samples, randomly downsampling sites, and specifying filters based on missing data and
+            minimum minor allele count (see the docs for `biallelic_snp_calls` for more information).
+            The `overwrite` parameter, set to true, will enable overwrite of data with the same
+            hash values (e.g. with a different randomly selected number of sites).
+        """,
+        returns=("Binary Plink output files: .bed, .bim, .fam"),
```
alimanfoo, line 40:
```suggestion
        returns="""
        Base path to files containing binary Plink output files. Append .bed, 
        .bim or .fam to obtain paths for the binary genotype table file, variant
        information file and sample information file respectively.
        """,
```

alimanfoo (commented):
Thanks @tristanpwdennis, noticed a couple more small things. Promise then I'll stop nit picking! If we could make these changes, then get all the checks to pass over in the shadow PR, happy to merge.

tristanpwdennis:
Accepted corrections :)

jonbrenas:
Thanks @tristanpwdennis. You probably saw that the tests failed. It looks like the 'tmp_path` used during the tests ends up being a `pathlib.PosixPath` instead of the required `str`.

alimanfoo (approved):
All looks good to me, happy to be merged once checks have passed.

alimanfoo:
Wooohoo!

tristanpwdennis:
Hooray, thanks guys! I wonder if I get a GH badge for being the longest open PR on a given repo...

PR #523, opened 2024-05-14
Add automatic selection of regional GCS storage

alimanfoo:
This PR adds automatic detection of GCP region, and selects a URL for storage in the same region if available.

This should reduce some network usage costs where we are accessing data from us-central1.


malariagen_data/util.py
```diff
@@ -918,7 +917,42 @@ def plotly_discrete_legend(
...
+                    return "us-west2"
+                elif region == "Utah":
+                    return "us-west3"
+                elif region == "Nevada":
+                    return "us-west4"
+                elif region == "Texas":
+                    return "us-south1"
+            elif country == "ZA":
+                if region == "Gauteng":
+                    return "africa-south1"
```
ahernank, line 950: @alimanfoo More of a question. Is there a benefit on listing regions directly here now? Is this thinking ahead on a scenario where the number of resources/regions increases in Africa, then more of the VMs get allocated in `africa-south1` & then, we might consider a new single region bucket for this region?

leehart (approved):

malariagen_data/util.py
alimanfoo: Yeah vaguely, I was just going to do the US regions for now but then happened to notice that the African region is up, and so added it here because I had the mapping from ipinfo region ("Gauteng") to GCP region ("africa-south1") to hand. 

But yes, at some point it would be cool to add a bucket in an African region.

alimanfoo:
Thanks @ahernank and @leehart. Just to check, is all the data available in the vo_afun_release_master_us_central1 bucket now? I.e., OK to merge this PR?

ahernank (approved):

ahernank:
Yes, @alimanfoo. The data from Af1.x has been copied to `vo_afun_release_master_us_central1`. 

Unrelated to this PR, we still need to delete the non GT data from the release bucket.

leehart:
> we still need to delete the non GT data from the release bucket.

@ahernank I think we've now decided to not delete the non-GT data from the Af release bucket, and hence the Zarr metadata and this package will not need updating. With that change in mind, I've actually restored the Af1.4 non-GT data to the release bucket. Obviously, this is inconsistent with our current Ag3.x release bucket and our SNP Data Release process. See https://github.com/malariagen/vector-data-processing/issues/36

alimanfoo:
Thanks both. Surfacing discussion from this morning, we may decide to deprecated the multi-region buckets, so will leave this PR open for now.

alimanfoo:
Looking at this PR again, I think I'd be in favour of merging this anyway, as it includes some improvements to the logic around when we check the client location and how the colab location check works. 

If we then deprecate the multi-region buckets we could make a subsequent PR to simplify back down.

PR #535, opened 2024-05-19
Pandas FutureWarning: Setting an item of incompatible dtype is deprecated

### LINKED ISSUE #532: Pandas FutureWarning: Setting an item of incompatible dtype is deprecated
When a value is missing in the genome features table, an empty string is stored instead of a float64 (`nan` is probably the best value in that case).
```
/Users/jb52/GitRepos/malariagen-data-python/malariagen_data/anoph/genome_features.py:366: FutureWarning:
  
  Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
```

jonbrenas:
Resolves #532.

It is, however, by removing a piece of code that replaced `nan` by `''` so I don't know if we want to find another solution.


alimanfoo:
> It is, however, by removing a piece of code that replaced `nan` by `''` so I don't know if we want to find another solution.

Thanks @jonbrenas, yes I think we will need to find another solution. Otherwise I think the tooltips in the genes plot will contain lots of "nan" labels which is a bit ugly.

leehart:
Hi @jonbrenas . I wonder if we can use the `Float64` type, which is supports nulls, instead of `float64`, and still convert the nulls to empty strings for presentation?
- https://pandas.pydata.org/docs/reference/api/pandas.Float64Dtype.html#pandas.Float64Dtype

alimanfoo:
> Hi @jonbrenas . I wonder if we can use the `Float64` type, which is supports nulls, instead of `float64`, and still convert the nulls to empty strings for presentation?
> 
> * https://pandas.pydata.org/docs/reference/api/pandas.Float64Dtype.html#pandas.Float64Dtype

I think the warning is due to attempting to replace nulls with the empty string (""). So probably using `Float64` instead of `float64` won't help I'm guessing. Perhaps the column should be cast to a string dtype, then fill nulls with empty string?

jonbrenas:
Looking a little bit more closely at the warming, I realised that the problem comes from `phase` and `score`, two columns that are actually not used by `plot_genes` (and luckily are the named the same in both `ag3` and `af1`). I just drop them in the function and that makes the warning disappear (at least in my tests).

alimanfoo:
> Looking a little bit more closely at the warming, I realised that the problem comes from `phase` and `score`, two columns that are actually not used by `plot_genes` (and luckily are the named the same in both `ag3` and `af1`). I just drop them in the function and that makes the warning disappear (at least in my tests).

Ah cool, that sounds good.

leehart:
Re: #554

alimanfoo (approved):

PR #507, opened 2024-02-27
Implement diplotype clustering, refactor haplotype clustering, and fix random_region_str()

sanjaynagi:
In this PR, we implement a plot_diplotype_clustering() function, which works exactly like plot_haplotype_clustering(), except it works on diplotypes rather than haplotypes. 

This PR works towards #361, we aim to make an 'advanced' version of this plotting function which plots amino acid mutation heatmap below it, as well as sample heterozygosity and possibly CNV copy number and diplotype cluster assignments. For now, we start with the simple dendrogram itself. 

I have added (Alistairs) functions into util.py which calculate pairwise distances for diplotypes. As far as I can tell, these functions were marginally different to the existing biallelic diplotype functions, so i have added new multiallelic ones. It might be that this is not quite necessary, please check @alimanfoo :)

Issue #361, opened 2023-03-14
Investigate mutations on haplotypes


alimanfoo:
We commonly would like to be able to investigate which haplotypes carry which mutations, particularly non-synonymous SNPs and CNVs. This is usually in the context of investigating a locus under recent selection, where typically you want to...

(1) Discover a locus under selection
(2) Investigate how many distinct sweeps at the locus, and whether sweeps are shared between species and/or countries
(3) Investigate any mutations on the swept haplotypes which might be driving selection

Previously we have done point 3 of this workflow in two different ways...

(A) Add information about non-synonymous SNPs as an additional sub-plot on a haplotype clustering dendrogram. E.g.:

![image](https://user-images.githubusercontent.com/703554/224982029-db1ead89-341f-4d24-9641-3eb2f4f22e45.png)

(B) Add information about non-synonymous SNPs onto a haplotype network by labelling the edges. E.g.:

![image](https://user-images.githubusercontent.com/703554/224983140-c254ef31-72cf-4f3b-9600-eb05f499f314.png)

We might want to ultimately split this issue into two issues, one covering haplotype dendrograms and one covering haplotype networks, but raising this umbrella initially to discuss some common points.

alimanfoo:
Previously @sanjaynagi and I have had lots of discussion about this, will try to surface some of that here :)

alimanfoo:
A key issue here is that the phased haplotypes available from the main haplotypes data releases do not contain some SNPs that you may be interested in. For example, the haplotypes only contain biallelic SNPs, and so any multiallelic (such as Vgsc V402L) are not included. Also, any SNPs failing site filters are not included, but site filters are very conservative and may filter out some SNPs you are interested in.

Previously we resolved this by performing some additional phasing of extra SNPs of interest onto the haplotype scaffold, using mvncall. However, this is awkard and painful to do, and not possible to support within the malariagen_data package in a general way.

This remains a difficult issue, and I think requires some rethinking of the approach.

DeribaAbera1234:
Dear Alistair, Thank you so much for the clarifications

sanjaynagi:
Thanks Alistair for the summary, and great timing as I had also started to think about trying to implement something for the Af1000 project analyses, in which we will need to do this. 

As we've spoke about in the past, perhaps it's time for us to try and retrieve amino acid mutations from diplotype clusters instead :)

alimanfoo:
[...]

PR #537, opened 2024-05-22
Advanced diplotype clustering

sanjaynagi:
Builds on #507 and partially addresses #361. 

In this PR, I add an advanced diplotype clustering function which performs diplotype clustering on a given region, and plots alongside it sample heterozygosity and amino acid alleles. This is particularly useful for investigating resistance mutations at IR loci. 

It adds a few new methods:

- an `aa_individual_allele_frequencies()` function, which just runs aa_allele_frequencies but calculates within individual frequencies (0, 0.5 or 1) rather than in cohorts. It could be that transforming the SNP data directly would be more appropriate and potentially faster than this. It might make more sense to use counts of alleles (0,1,2). 
- a function to calculate and plot heterozygosity over a provided region (current methods in malariagen_data, `_sample_count_het()` are windowed and applied to one sample at a time)
- A function to concatenate plotly subplots into one figure (`concat_subplots()`)
- A function to extract the sample order from the dendrogram 

For now, I have not added CNV support, however, it should be added prior to merging this PR.


sanjaynagi:
ahhh, it seems I forgot I was working on a fork, as such, the google credentials dont work.  Whats the best approach in future? @ahernank @alimanfoo

ahernank:
Hi @sanjaynagi, a short-term solution, but you have permissions to write directly on the repo rather than using branches.

sanjaynagi:
I've added CNV support, though currently the bar is copy number, but potentially it should be # extra copies.

alimanfoo:
Thanks @sanjaynagi. Just to mention I updated this branch to bring in some unrelated test fixes from master.

alimanfoo:
Very exciting, I'll give this a proper look asap!

alimanfoo:
Hi @sanjaynagi, some investigation into the CNV plotting issue here: https://github.com/sanjaynagi/malariagen-data-python/pull/1

sanjaynagi:
Some updates:

thanks to @alimanfoo for sorting the CNV colorscale problem. 

Ive completely removed any usage of px.imshow() now and replaced with go.Heatmap() for heterozygosity and amino acid heatmaps as well as CNVs. This also simplifies the code further, as it seems that we now don't need to pass the x_range through various functions which we did previously. 

I've added a check to see if there are actually any amino acids above the MAF threshold, if there are not, an amino acid heatmap will not be created.

Think we just need to maybe write some tests, and a final check from @alimanfoo, then we may be good to go! yeehaw!! Ill have a go at zooming in the x-axis aswell (currently its slightly zoomed out). 

image

alimanfoo:
Hi @sanjaynagi, some comments and suggestions over in https://github.com/sanjaynagi/malariagen-data-python/pull/2

malariagen_data/anoph/dipclust.py
```diff
@@ -273,3 +280,418 @@ def _diplotype_pairwise_distances(
...
+        fig.update_xaxes(visible=False)
+        fig.update_layout(
+            title=title,
+            width=width,
+            height=height,
+            hovermode="closest",
+            plot_bgcolor="white",
+        )
+
+        # COMMENT: this feels overly complex - can we somehow remove the colorbar earlier,
```
sanjaynagi, line 471: Not sure - I did try that originally, but was struggling a lot to do so. Also, I think it requires making Figure objects for the heatmaps first, whereas currently we just make heatmap traces, so it kind of adds more code.

sanjaynagi: haven't got time right now to implement the other stuff but feel free to do so

malariagen_data/anoph/dipclust.py
```diff
@@ -273,3 +280,418 @@ def _diplotype_pairwise_distances(
             gt_samples=gt_samples,
             n_snps=np.array(gt_seg.shape[0]),
         )
+
+    def _dipclust_extract_dendro_sample_order(self, fig):
+        # COMMENT: Is there a simpler way to obtain the sample order?
```
sanjaynagi, line 285: something tells me there is a simpler way  😂

malariagen_data/anoph/dipclust.py
```diff
@@ -273,3 +280,418 @@ def _diplotype_pairwise_distances(
...
+            color_discrete_map=color_discrete_map,
+            category_orders=category_orders,
+            legend_sizing=legend_sizing,
+            random_seed=random_seed,
+        )
+
+        fig_dendro = res["figure"]
+        n_snps = res["n_snps"]
+        dendro_sample_id_order = (
+            # COMMENT: Can we simplify? Why do this sorting here?
```
sanjaynagi, line 570: something tells me we can 😂

sanjaynagi:
@alimanfoo I've had a go at all the comments except the removing colorbars earlier one, cos I tried to do that in the past and it took me a while and I couldn't :)

alimanfoo:
Have pushed a few suggested changes directly to this branch, hope that's OK. Still reviewing so may do a bit more.

alimanfoo:
Hi @sanjaynagi, I pushed a few more changes, includes:

* Set colorscale visibility when heatmaps are created.
* Calculate denominator for heterozygosity based on number of called genotypes.
* Remove logic for drawing lines on the SNP heatmap - that wasn't being used after recent refactoring.
* Renamed a few things from "aa_..." to "snp_..." to reflect that we are generalising to plotting any kind of SNP.

alimanfoo:
Hi @sanjaynagi, I think this is basically good to merge. Because this is a PR from a fork the CI tests won't run, so I'm thinking to merge this PR then open up a follow on PR from a branch within the main repo to confirm tests all run. How does that sound? Anything else you'd like to do here before I merge?

alimanfoo (approved):

sanjaynagi:
Hey @alimanfoo . Sounds good to me. Nothing else to add currently.

PR #540, opened 2024-05-24
plot_pairwise_average_fst may change the order of the cohorts when using a cohort dict

### LINKED ISSUE #539: plot_pairwise_average_fst may change the order of the cohorts when using a cohort dict
When using user defined cohorts using a dictionary, say:

Screenshot 2024-05-24 at 08 40 25

`plot_pairwise_average_fst` (and I assume some other functions as well) will reorder them so that they are in alphabetical order. This is a problem when, e.g., the `annotate_se` option is used because the `se` and `fst` values are swapped, e.g. using the dictionary from the above example:

Screenshot 2024-05-24 at 08 45 04

Another time when that is a problem is when the user ordered the keys for a good reason.

jonbrenas:
Resolves #539. There might be other functions affected in the same way.


malariagen_data/anoph/fst.py
```diff
@@ -493,7 +493,9 @@ def plot_pairwise_average_fst(
         **kwargs,
     ):
         # setup df
-        cohort_list = np.unique(fst_df[["cohort1", "cohort2"]].values)
+        cohort_list = pd.unique(
+            [c for cl in fst_df[["cohort1", "cohort2"]].values for c in cl]
```
alimanfoo, line 497: The for loops in this line look a little funky. What is it supposed to be doing?

alimanfoo:
Hi @jonbrenas, just to mention I've updated this branch to bring in some test fixes from master.

malariagen_data/anoph/fst.py
jonbrenas: The goal is to get the list of all cohorts in the same order they were defined in the cohort dictionary. `plot_pairwise_average_fst` does not have the list of cohorts or the dictionary as an input so we have to access it in another way. `fst_df[["cohort1", "cohort2"]]` is a list of list each one containing two cohort names. The last cohort of the dictionary is missing from the `cohort1` column and the first is missing from `cohort2`, hence why they are both accessed. `np.unique` is able to take a list of list and return every unique value but it sorts them alphabetically breaking the dictionary's order. `pd.unique` keeps the order but can't deal with a list of lists. The for loops are thus used to go from a list of list of cohorts to a list of cohorts.

ahernank:
@jonbrenas I've just re-ran the tests here to check they are all good now. It is only the coverage that needs a bit of additional work.

jonbrenas:
Thanks @ahernank. I am not quite sure what the problem with codecov/project is, though. It looks like there is an indirect change that leaves a line uncovered in `dipclust.py` but I can't really figure out why that would be.

malariagen_data/anoph/fst.py
alimanfoo: Thanks @jonbrenas. What about:

```
pd.unique(fst_df[["cohort1", "cohort2"]].values.flatten())
```

leehart: Nested list comprehensions always bake my noodle!

I _think_ it's equivalent to this:

```python
cohort_pairs = fst_df[["cohort1", "cohort2"]].values
flattened_cohorts = []
for cohort_pair in cohort_pairs:
	for cohort in cohort_pair:
		flattened_cohorts.append(cohort)
```

Or like Alistair suggested:
```python
list(fst_df[["cohort1", "cohort2"]].values.flatten())
```

leehart (changes requested):
Could we try using `flatten()` instead of nested list comprehension, just for readability?

jonbrenas:
The bottom-left half of the diagram wasn't being filled anymore. I also changed the list comprehension to .flatten(). (I personally find the list comprehension more readable but I see I am in the minority ;) )

leehart:
Cool, thanks @jonbrenas ! 

I see there's suddenly a code coverage failure in CI, which I don't quite understand, which I think we've experienced before. I wonder if it's the product of something random. I can't seem to re-run it with any success. I think in the past I've just had to add more coverage to an unrelated part of the code to satisfy the threshold, but it's annoying.

jonbrenas:
I think the problem is with dipcluster.py. It might be that the test doesn't cover some of the new functions.

leehart:
Yes, I see that `anoph/dipclust.py` is the file with the biggest "Change %", i.e. `-49.40%`, with `anoph/snp_frq.py` having `-4.79%`.

I'm not sure what these figures mean!?

From https://docs.codecov.com/docs/coverage-percentages
> Head coverage percent is the line coverage of all your files in your head commit.
> Patch coverage percent is the line coverage of all the lines changed in your commit.
> Change percent is the percent that the overall project line coverage has changed from base to head.

I guess if we can somehow claw back -2.72% test coverage, it should pass.

alimanfoo:
Hi folks, I've folded this fix into some other maintenance I was doing on the plot_pairwise_fst_function() over in #579. I'll close here and merge in #579 if approved.

PR #541, opened 2024-05-24
Updated parasite test buckets from staging to release

nw20:
Updated files containing paths to parasite GCS staging buckets,`test_pf7_integration.py` and `test_pv4_integration.py`, to use release buckets.


alimanfoo (approved):
Thanks @nw20 🙏

PR #544, opened 2024-05-28
Add clip_min parameter for Fst GWSS analyses 

### LINKED ISSUE #543: Should Fst be clipped between 0 and 1?
I was trying to use the fst functions to compute Fst between individuals, in this case putative siblings, and was receiving values back which I knew were incorrect (Fst between regions in IBD should be -.25 or -1). After I looked in more detail, I see that when computing fst between two cohorts, Fst is clipped between 0 and 1.

https://github.com/malariagen/malariagen-data-python/blob/dc89c9ceaa6e7bff3d0842b626870140a8ff6809/malariagen_data/anoph/fst.py#L83

I suggest we either:

A) do not clip Fst and return the true Fst values
B) add a parameter such as `clip_min` which changes the floor that we clip to (so I can set it to -1). 

I actually have a slight preference for A, but I'm happy to implement either. In general I think its good to return the data as it truly exists - it allows us to discover more weird/interesting stuff, rather than it being hidden.

sanjaynagi:
Resolves #543. 

I have added a parameter, clip_min, which allows the user to set an optional minimum value for the Fst GWSS functions. Previously they were automatically clipped to 0-1, which is confusing when looking at Fst data between relatives.


leehart:
Just to note that there is a minor linting failure, and other CI test failures due to not being able to authenticate with Google Cloud from a fork. Also needs an update from the master branch.

alimanfoo:
Bringing this in via #580.

PR #546, opened 2024-06-04
Ensure a site mask is used by default for Anopheles popgen functions

### LINKED ISSUE #545: Anopheles pca() should use default site mask
Previously I recall that the Anopheles pca() function used the default site mask if the `site_mask` parameter was not given.

However, now looking at the code it seems that no site mask will be used if the `site_mask` parameter is not given, because the default value of the `site_mask` parameter is `None`. It probably should be `"default"` instead which is the general way to say "use a site mask please, whichever is the default for this data resource".

alimanfoo:
Make sure that all Anopheles popgen functions use a site mask by default. This is achieved by setting the default value of the `site_mask` parameter to the sentinel `base_params.DEFAULT` which is resolved to whichever is the default site mask for the given data resource (e.g., "gamb_colu_arab" for Ag3).

Also make sure that `None` can be used as a value of the `site_mask` parameter, in case expert users want to see results without any site mask applied.

* Resolves #545


malariagen_data/anoph/pca.py
```diff
@@ -60,7 +60,7 @@ def pca(
         sample_sets: Optional[base_params.sample_sets] = None,
         sample_query: Optional[base_params.sample_query] = None,
         sample_indices: Optional[base_params.sample_indices] = None,
-        site_mask: Optional[base_params.site_mask] = None,
+        site_mask: Optional[base_params.site_mask] = base_params.DEFAULT,
```
alimanfoo, line 63: Will cause behaviour change for `pca()` when `site_mask` parameter is not given.

malariagen_data/anopheles.py
```diff
@@ -2979,7 +2978,7 @@ def biallelic_diplotype_pairwise_distances(
         sample_sets: Optional[base_params.sample_sets] = None,
         sample_query: Optional[base_params.sample_query] = None,
         sample_indices: Optional[base_params.sample_indices] = None,
-        site_mask: Optional[base_params.site_mask] = None,
+        site_mask: Optional[base_params.site_mask] = base_params.DEFAULT,
```
alimanfoo, line 2981: Will cause behaviour change for `biallelic_diplotype_pairwise_distance()` when `site_mask` parameter is not given.

malariagen_data/anopheles.py
```diff
@@ -3159,7 +3158,7 @@ def plot_njt(
...
```
alimanfoo, line 3161: Will cause behaviour change for `plot_njt()` when `site_mask` parameter is not given.

alimanfoo (commented):
Three functions will have a behaviour change, so this PR should trigger a major version bump for the next release.

leehart (approved):

leehart:
Merging soon, unless objections. Will then release v10.0.0

ahernank (approved):

Issue #555, opened 2024-06-24
Diplotype clustering fails on sample sets without CNV calls 


sanjaynagi:
![image](https://github.com/malariagen/malariagen-data-python/assets/34922269/10ae5217-3cbf-4979-8134-c61c11ca33cd)

There are no CNV calls for this sample set so it fails. Earlier versions of the diplotype clustering function had a try/except statement to get around this, could implement something like that.

sanjaynagi:
temporary workaround 

`  
sample_sets=[s for s in ag3.sample_sets()['sample_set'].to_list() if s != 'barron-2019']
`

alimanfoo:
Seems fine to handle this error internally, but how do we communicate to the user that the data are missing? Need to make sure it doesn't look like these samples have normal copy number.

leehart:
Related issue https://github.com/malariagen/malariagen-data-python/issues/363

leehart:
Would it be sufficient to simply skip sample sets that don't have CNV HMM data in the API?

Essentially:
```python
                    y = self._cnv_hmm_dataset(
                        contig=r.contig,
                        sample_set=s,
                        inline_array=inline_array,
                        chunks=chunks,
                    )

                    # If no CNV HMM dataset was found then skip
                    if y is None:
                        continue

                    ly.append(y)
```

For example, I can submit a PR that would allow:
```python
ag3.plot_diplotype_clustering_advanced(
  region='2L:28,535,000-28,552,000',
  cnv_region='2L:28,535,000-28,552,000',
  sample_query='taxon == "gambiae" and year > 2019',
  site_mask='gamb_colu_arab',
  color='taxon',
  snp_transcript='AGAP006227-RA',
)
```
![newplot](https://github.com/malariagen/malariagen-data-python/assets/4256466/b297c3f2-16e8-4bb0-afcd-6626a8491e6f)

leehart:
We now skip sample sets that don't have CNV HMM data but still raise `ValueError` when no CNV HMM data are found at all.

Issue #551, opened 2024-06-17
Handle missing discordant read calls for Af1.x


leehart:
Af1.x CNV releases will be considered complete without DRCs, only including HMM data and coverage calls.

I think the meaning of the pre-release `pre` arg, e.g. `Af1(pre=True)` will correspond to a known temporary incompleteness of data, rather than any embargo status, which will be handled separately via forthcoming terms-of-use info instead.

Possibly related: issue https://github.com/malariagen/malariagen-data-python/issues/89

leehart:
Possibly connectable: issue https://github.com/malariagen/malariagen-data-python/issues/363 PR https://github.com/malariagen/malariagen-data-python/pull/409

leehart:
Currently, attempting to run something like:
```python
af1.cnv_discordant_read_calls(contig='2RL')
```
will result in `FileNotFoundError: '.zmetadata'`.

Compare to CNV HMM code, which currently handles missing data using `None`, i.e.
```python
    def open_cnv_hmm(self, sample_set: base_params.sample_set) -> zarr.hierarchy.Group:
        try:
            return self._cache_cnv_hmm[sample_set]
        except KeyError:
            release = self.lookup_release(sample_set=sample_set)
            release_path = self._release_to_path(release)
            path = f"{self._base_path}/{release_path}/cnv/{sample_set}/hmm/zarr"

            # If CNV HMM data exists for this sample set then return the zarr,
            # Otherwise return None.
            marker = path + "/.zmetadata"
            if self._fs.exists(marker):
                store = init_zarr_store(fs=self._fs, path=path)
                root = zarr.open_consolidated(store=store)
            else:
                root = None

            self._cache_cnv_hmm[sample_set] = root
        return root
```

Compare to CNV coverage calls code, which currently handles missing data using `ValueError`, i.e.
```python
    def open_cnv_coverage_calls(
        self,
        sample_set: base_params.sample_set,
        analysis: cnv_params.coverage_calls_analysis,
    ) -> zarr.hierarchy.Group:
        key = (sample_set, analysis)
        try:
            return self._cache_cnv_coverage_calls[key]
        except KeyError:
            release = self.lookup_release(sample_set=sample_set)
            release_path = self._release_to_path(release)
            path = f"{self._base_path}/{release_path}/cnv/{sample_set}/coverage_calls/{analysis}/zarr"
            # N.B., not all sample_set/analysis combinations exist, need to check
            marker = path + "/.zmetadata"
[...]

Issue #89, opened 2021-11-25
Simplify handling of empty datasets


alimanfoo:
When access haplotypes data, there are cases where there are no data for a given sample set and analysis, e.g., for the "arab" analysis where a sample set has no arabiensis samples. Currently the approach is to return `None` in cases like this.

Similarly, when accessing CNV discordant read calls data, there are no data for some contigs. Again the current approach is to return `None`.

However, it would probably be simpler, internally at least, if in these cases a dataset was returned, but with a 0 length dimension. E.g., if there are no samples for a given haplotypes dataset, return a dataset with 0 length samples dimension. E.g., if there are no variants for a CNV dataset for a given contig, return a dataset with a 0 length variants dimension.

leehart:
Noting here

`haplotypes()` has:
```
for r in region:
    ly = []

    for s in sample_sets:
        y = self._haplotypes_dataset(
            contig=r.contig,
            sample_set=s,
            analysis=analysis,
            inline_array=inline_array,
            chunks=chunks,
        )
        if y is not None:
            ly.append(y)

    if len(ly) == 0:
        debug("early out, no data for given sample sets and analysis")
        return None
```

`_haplotypes_dataset()` has:
```
root = self.open_haplotypes(sample_set=sample_set, analysis=analysis)
sites = self.open_haplotype_sites(analysis=analysis)
```
```
if root is None:
    return None
```

In contrast, `cnv_discordant_read_calls()` has:
```
lx = []
for c in contig:

    ly = []
    for s in sample_sets:
        y = self._cnv_discordant_read_calls_dataset(
            contig=c,
            sample_set=s,
            inline_array=inline_array,
            chunks=chunks,
        )
        ly.append(y)

    x = xarray_concat(ly, dim=DIM_SAMPLE)
    lx.append(x)
```


and `_cnv_discordant_read_calls_dataset()` has:
```
root = self.open_cnv_discordant_read_calls(sample_set=sample_set)
```
```
if contig not in root:
    raise ValueError(f"no CNVs available for contig {contig!r}")
```

leehart:
Since previous comments, the code for `haplotypes()` appears to have changed to raise a `ValueError` instead of returning `None`.
```python
                if len(ly) == 0:
                    # Bail out, no data for given sample sets and analysis.
                    raise ValueError(
                        f"No samples found for phasing analysis {analysis!r}"
                    )
```

leehart:
[...]

PR #565, opened 2024-07-09
Handle missing CNV data

leehart:
Re: issues #363, #555

Note: This currently won't address #89 nor #551


leehart:
Superseded by #575

Issue #568, opened 2024-07-19
Anopheles pca() and plot_njt() functions should have sensible defaults for min_minor_ac and max_missing_an


alimanfoo:
Previously ([in version 7](https://github.com/malariagen/malariagen-data-python/blob/5daf1c32844f50a0fbe6092fb30deced2316a796/malariagen_data/anopheles.py#L3218)) we set defaults for these parameters:

```python
    def pca(
        self,
        region,
        n_snps,
        thin_offset=0,
        sample_sets=None,
        sample_query=None,
        site_mask="default",
        min_minor_ac=2,
        max_missing_an=0,
        n_components=20,
    ):
```

When some refactoring happened internally, we also changed these default values (unintentionally?) to None. This means that PCAs being run with default parameters may be using lots of uninformative SNPs (non-segregating and singletons) and also SNPs with lots of missingness, which is not ideal.

Propose to set these parameters back to their default values in the pca() function as of version 7.

alimanfoo:
We should also set the same defaults for the plot_njt() function.

alimanfoo:
cc @leehart

leehart:
Investigating test failures:
```
FAILED tests/anoph/test_pca.py::test_pca_plotting[ag3_sim] - ValueError: Not enough SNPs.
FAILED tests/anoph/test_pca.py::test_pca_plotting[af1_sim] - ValueError: Not enough SNPs.
```

leehart:
For some reason `biallelic_snp_calls()` isn't getting many variants, e.g. 
```
test_pca n_snps_available 19262
test_pca n_snps 17593

biallelic_snp_calls ds_out.sizes["variants"] 333
biallelic_snp_calls n_snps 17593
```
🤔

leehart:
Ah, I suspect we might need to apply the defaults to these functions too:
- `biallelic_snp_calls()`
- `biallelic_diplotypes()`
- `biallelic_diplotype_pairwise_distances()`

alimanfoo:
> Ah, I suspect we might need to apply the defaults to these functions too:
> 
> * `biallelic_snp_calls()`
> * `biallelic_diplotypes()`
> * `biallelic_diplotype_pairwise_distances()`

FWIW I would leave default values as None for these functions. These are more general functions to obtain SNP data. The pca() and plot_njt() are more specific functions where it matters more to have low missingness and segregating variation.

alimanfoo:
[...]

PR #569, opened 2024-07-19
Add defaults for min_minor_ac, max_missing_an

leehart:
Re: issue #568

### DIFF
--- a/malariagen_data/anoph/base_params.py
+++ b/malariagen_data/anoph/base_params.py
@@ -270,8 +270,6 @@ min_minor_ac: TypeAlias = Annotated[
     """,
 ]
 
-min_minor_ac_default: min_minor_ac = 2
-
 max_missing_an: TypeAlias = Annotated[
     int,
     """
@@ -280,8 +278,6 @@ max_missing_an: TypeAlias = Annotated[
     missing calls.
     """,
 ]
-
-max_missing_an_default: max_missing_an = 0
 
 snp_query: TypeAlias = Annotated[
     str,
--- a/malariagen_data/anoph/base_params.py
+++ b/malariagen_data/anoph/base_params.py
@@ -270,6 +270,8 @@ min_minor_ac: TypeAlias = Annotated[
     """,
 ]
 
+min_minor_ac_default: min_minor_ac = 2
+
 max_missing_an: TypeAlias = Annotated[
     int,
     """
@@ -278,6 +280,8 @@ max_missing_an: TypeAlias = Annotated[
     missing calls.
     """,
 ]
+
+max_missing_an_default: max_missing_an = 0
 
 snp_query: TypeAlias = Annotated[
     str,
--- a/malariagen_data/anoph/snp_data.py
+++ b/malariagen_data/anoph/snp_data.py
@@ -1565,12 +1565,8 @@ class AnophelesSnpData(
         min_cohort_size: Optional[base_params.min_cohort_size] = None,
         max_cohort_size: Optional[base_params.max_cohort_size] = None,
         random_seed: base_params.random_seed = 42,
-        min_minor_ac: Optional[
-            base_params.min_minor_ac
-        ] = base_params.min_minor_ac_default,
-        max_missing_an: Optional[
-            base_params.max_missing_an
-        ] = base_params.max_missing_an_default,
+        min_minor_ac: Optional[base_params.min_minor_ac] = None,
+        max_missing_an: Optional[base_params.max_missing_an] = None,
         n_snps: Optional[base_params.n_snps] = None,
         thin_offset: base_params.thin_offset = 0,
     ) -> xr.Dataset:
@@ -1713,12 +1709,8 @@ class AnophelesSnpData(
         min_cohort_size: Optional[base_params.min_cohort_size] = None,
         max_cohort_size: Optional[base_params.max_cohort_size] = None,
         random_seed: base_params.random_seed = 42,
-        min_minor_ac: Optional[
-            base_params.min_minor_ac
-        ] = base_params.min_minor_ac_default,
-        max_missing_an: Optional[
-            base_params.max_missing_an
-        ] = base_params.max_missing_an_default,
+        min_minor_ac: Optional[base_params.min_minor_ac] = None,
+        max_missing_an: Optional[base_params.max_missing_an] = None,
         n_snps: Optional[base_params.n_snps] = None,
         thin_offset: base_params.thin_offset = 0,
         inline_array: base_params.inline_array = base_params.inline_array_default,
--- a/malariagen_data/anoph/snp_data.py
+++ b/malariagen_data/anoph/snp_data.py
@@ -1565,8 +1565,12 @@ class AnophelesSnpData(
         min_cohort_size: Optional[base_params.min_cohort_size] = None,
         max_cohort_size: Optional[base_params.max_cohort_size] = None,
         random_seed: base_params.random_seed = 42,
-        min_minor_ac: Optional[base_params.min_minor_ac] = None,
-        max_missing_an: Optional[base_params.max_missing_an] = None,
+        min_minor_ac: Optional[
+            base_params.min_minor_ac
+        ] = base_params.min_minor_ac_default,
+        max_missing_an: Optional[
+            base_params.max_missing_an
+        ] = base_params.max_missing_an_default,
         n_snps: Optional[base_params.n_snps] = None,
         thin_offset: base_params.thin_offset = 0,
     ) -> xr.Dataset:
@@ -1709,8 +1713,12 @@ class AnophelesSnpData(
         min_cohort_size: Optional[base_params.min_cohort_size] = None,
         max_cohort_size: Optional[base_params.max_cohort_size] = None,
         random_seed: base_params.random_seed = 42,
-        min_minor_ac: Optional[base_params.min_minor_ac] = None,
-        max_missing_an: Optional[base_params.max_missing_an] = None,
+        min_minor_ac: Optional[
+            base_params.min_minor_ac
+        ] = base_params.min_minor_ac_default,
+        max_missing_an: Optional[
+            base_params.max_missing_an
+        ] = base_params.max_missing_an_default,
         n_snps: Optional[base_params.n_snps] = None,
         thin_offset: base_params.thin_offset = 0,
         inline_array: base_params.inline_array = base_params.inline_array_default,
--- a/malariagen_data/anoph/snp_data.py
+++ b/malariagen_data/anoph/snp_data.py
@@ -1655,7 +1655,7 @@ class AnophelesSnpData(
             ds_out = xr.Dataset(coords=coords, data_vars=data_vars, attrs=ds.attrs)
 
             # Apply conditions.
-            if max_missing_an or min_minor_ac:
+            if max_missing_an is not None or min_minor_ac is not None:
                 loc_out = np.ones(ds_out.sizes["variants"], dtype=bool)
 
                 # Apply missingness condition.
--- a/tests/anoph/test_pca.py
+++ b/tests/anoph/test_pca.py
@@ -9,6 +9,7 @@ from pytest_cases import parametrize_with_cases
 from malariagen_data import af1 as _af1
 from malariagen_data import ag3 as _ag3
 from malariagen_data.anoph.pca import AnophelesPca
+from malariagen_data.anoph import pca_params
 
 
 @pytest.fixture
@@ -83,7 +84,11 @@ def test_pca_plotting(fixture, api: AnophelesPca):
         sample_sets=random.sample(all_sample_sets, 2),
         site_mask=random.choice((None,) + api.site_mask_ids),
     )
-    ds = api.biallelic_snp_calls(**data_params)
+    ds = api.biallelic_snp_calls(
+        min_minor_ac=pca_params.min_minor_ac_default,
+        max_missing_an=pca_params.max_missing_an_default,
+        **data_params,
+    )
 
     # PCA parameters.
     n_samples = ds.sizes["samples"]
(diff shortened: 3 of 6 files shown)


leehart:
Investigating coverage test failures, e.g.
```
__________________________ test_pca_plotting[ag3_sim] __________________________

fixture = <tests.anoph.conftest.Ag3Simulator object at 0x7f6794114710>
api = <malariagen_data.anoph.pca.AnophelesPca object at 0x7f676cc47260>

    @parametrize_with_cases("fixture,api", cases=".")
    def test_pca_plotting(fixture, api: AnophelesPca):
        # Parameters for selecting input data.
        all_sample_sets = api.sample_sets()["sample_set"].to_list()
        data_params = dict(
            region=random.choice(api.contigs),
            sample_sets=random.sample(all_sample_sets, 2),
            site_mask=random.choice((None,) + api.site_mask_ids),
        )
        ds = api.biallelic_snp_calls(**data_params)
    
        # PCA parameters.
        n_samples = ds.sizes["samples"]
        n_snps_available = ds.sizes["variants"]
>       n_snps = random.randint(n_samples, n_snps_available)

tests/anoph/test_pca.py:91: 
```
```
FAILED tests/anoph/test_pca.py::test_pca_plotting[ag3_sim] - ValueError: empty range in randrange(97, 69)
```

leehart:
@alimanfoo Thanks for your replies on my issue comments. I'm not sure I understand.

If I revert the default values (to `None`) for `biallelic_snp_calls()`, `biallelic_diplotypes()` and `biallelic_diplotype_pairwise_distances()`, as you suggest, then I expect I'll get the same `ValueError: Not enough SNPs.` from `test_pca_plotting`.

This PR currently applies the defaults to those "more general" functions too, including `max_missing_an=0`, which understand is not ideal, but somehow satisfied the tests.

So I gather we need to find a way of leaving the default values as `None` for those functions, while still avoiding `ValueError: Not enough SNPs.` from `test_pca_plotting`, unless I'm missing something.

With the default values of `None`, the `biallelic_snp_calls()` function was apparently getting low numbers of `ds_out.sizes["variants"]`, which seemed to cause the test failures, but I think you're saying that this would also be expected when using the default value of `max_missing_an=0`, which requires zero missingness.

leehart:
To demonstrate or clarify, and perhaps help debug, the test failure mentioned above, i.e. `ValueError: Not enough SNPs.`, when keeping the default values as `None` for `biallelic_snp_calls()`, `biallelic_diplotypes()` and `biallelic_diplotype_pairwise_distances()`, I've created a draft PR here: https://github.com/malariagen/malariagen-data-python/pull/570

malariagen_data/anoph/snp_data.py
```diff
@@ -1565,8 +1565,12 @@ def biallelic_snp_calls(
...
         max_cohort_size: Optional[base_params.max_cohort_size] = None,
         random_seed: base_params.random_seed = 42,
-        min_minor_ac: Optional[base_params.min_minor_ac] = None,
-        max_missing_an: Optional[base_params.max_missing_an] = None,
+        min_minor_ac: Optional[
+            base_params.min_minor_ac
+        ] = base_params.min_minor_ac_default,
+        max_missing_an: Optional[
+            base_params.max_missing_an
+        ] = base_params.max_missing_an_default,
```
alimanfoo, line 1573: Suggest to leave defaults of None for this function, biallelic_snp_calls() is a more general function and so these defaults don't necessarily make sense.

malariagen_data/anoph/snp_data.py
```diff
@@ -1709,8 +1713,12 @@ def biallelic_diplotypes(
...
```
alimanfoo, line 1721: As above, suggest to leave defaults of None for this function, biallelic_diplotypes() is a more general function and so these defaults don't necessarily make sense.

tests/anoph/test_pca.py
```diff
@@ -88,9 +88,14 @@ def test_pca_plotting(fixture, api: AnophelesPca):
...
     n_snps_available = ds.sizes["variants"]
-    n_snps = random.randint(n_samples, n_snps_available)
     n_components = random.randint(3, n_samples)
 
+    # If there are more samples than SNPs available, use all SNPs.
+    if n_samples > n_snps_available:
+        n_snps = n_snps_available
+    else:
+        n_snps = random.randint(n_samples, n_snps_available)
+
```
alimanfoo, line 98: I don't think we properly considered the case where the number of SNPs could be very small, less than the number of samples.

I would guess this is getting triggered because the simulated data has a certain degree of missingness, which means that under the default parameter value of 0 for `max_missing_an` then very few SNPs meet that requirement.

One option here could be to set a less strict value of `max_missing_an` for the purposes of this test. 

You could also keep the changes you've made here, but you might also need to be careful about handling the number of components (`n_components`). In the extreme case of very small number of SNPs or samples, the total number of components available in the output of the PCA will be `min(n_samples, n_snps)`. So, e.g., even if you requested `n_components = 4`, if there are only 3 SNPs then you will only get 3 components in the output.

Perhaps this could all be simplified to:

```
    n_snps = random.randint(1, n_snps_available)
    n_components = random.randint(2, min(n_samples, n_snps))
```

...which should be pretty robust to running with a small number of SNPs.

malariagen_data/anoph/base_params.py
```diff
@@ -279,6 +281,8 @@ def validate_sample_selection_params(
     """,
 ]
 
+max_missing_an_default: max_missing_an = 0
+
```
alimanfoo, line 285: I'm fine to have these here, but it could be argued that these defaults are less "base" defaults but rather more specific defaults that only apply when you want to run certain types of population structure analyses such as PCA or NJT, where it matters to have segregating variants and you want to minimise missingness. As such they could possibly live better in the pca_params module.

alimanfoo:
> To demonstrate or clarify, and perhaps help debug, the test failure mentioned above, i.e. `ValueError: Not enough SNPs.`, when keeping the default values as `None` for `biallelic_snp_calls()`, `biallelic_diplotypes()` and `biallelic_diplotype_pairwise_distances()`, I've created a draft PR here: #570

This error will be unrelated to the default values that are set for `max_missing_an` and `min_minor_ac` within the `biallelic_snps_calls()` function.

The test triggering the error is `test_pca_plotting`. This calls `pca()` where there are default values set for `max_missing_an` and `min_minor_ac`. These values then get passed through when `pca()` calls `biallelic_diplotypes()` internally.

alimanfoo:
FWIW I would say the main hurdle here is that the default value of `max_missing_an` doesn't work very well for the amount of missingness in the simulated data.

Perhaps the best way to deal with that would be to provide a more sensible value of `max_missing_an` when calling functions within the tests, or adjust the simulated data to reduce the amount of missingness.

alimanfoo:
This does also highlight a general weakness with setting `max_missing_an=0` as default for all datasets. In real datasets, as the number of samples gets larger, so the chance that you'll find variants with no missingness at all get smaller. 

Rather it would be better to set a fractional value, i.e., to allow something like `max_missing_an=0.01` which would mean keep all variants with at most 1% missing genotype calls.

malariagen_data/anoph/base_params.py
leehart: Thanks @alimanfoo . Originally, I put `max_missing_an_default` and `min_minor_ac_default` in `pca_params.py` before I realised that `min_minor_ac` and `max_missing_an` were already in `base_params.py`.
I assumed we shouldn't have duplicate params in different modules.

tests/anoph/test_pca.py
leehart: Sounds right. Thanks @alimanfoo .

malariagen_data/anoph/snp_data.py
leehart: Will do. Thanks @alimanfoo .

malariagen_data/anoph/snp_data.py
leehart: Will do. Thanks @alimanfoo .

leehart:
Tests failing (for 3.12) after merging in master branch:
```
=========================== short test summary info ============================
FAILED tests/anoph/test_g123.py::test_g123_calibration[af1_sim] - IndexError: index -1 is out of bounds for axis 0 with size 0
=========== 1 failed, 373 passed, 174 warnings in 179.66s (0:02:59) ============
```
I'll try rerunning.

malariagen_data/anoph/base_params.py
leehart: I wonder if something like this works in `pca_params.py`:
```python
from . import base_params
min_minor_ac_default: base_params.min_minor_ac = 2
```
To allow:
```python
        min_minor_ac: Optional[
            base_params.min_minor_ac
        ] = pca_params.min_minor_ac_default,
```

leehart:
Tests failing (for 3.9 instead now, after 3.11 passed):
```
=========================== short test summary info ============================
FAILED tests/anoph/test_pca.py::test_pca_plotting[ag3_sim] - AssertionError: ('n_components', 74, 'n_samples', 98, 'n_snps_available', 54, ...)
assert 'PC55' in Index(['sample_id', 'partner_sample_id', 'contributor', 'country', 'location',\n       'year', 'month', 'latitude', 'longitude', 'sex_call',\n       ...\n       'PC45', 'PC46', 'PC47', 'PC48', 'PC49', 'PC50', 'PC51', 'PC52', 'PC53',\n       'PC54'],\n      dtype='object', length=108)
 +  where Index(['sample_id', 'partner_sample_id', 'contributor', 'country', 'location',\n       'year', 'month', 'latitude', 'longitude', 'sex_call',\n       ...\n       'PC45', 'PC46', 'PC47', 'PC48', 'PC49', 'PC50', 'PC51', 'PC52', 'PC53',\n       'PC54'],\n      dtype='object', length=108) =                    sample_id partner_sample_id  ...      PC53      PC54\n0   VBS01780-4651STDY7017723            GP1621  ... -0.872420  0.109044\n1   VBS01250-4651STDY7017429            GP1091  ... -0.235901 -0.184651\n2   VBS01789-4651STDY7017732            GP1630  ... -0.832990 -0.428277\n3   VBS01222-4651STDY7017404            GP1063  ... -0.373940 -0.054506\n4   VBS01943-4651STDY7017870            GP1784  ...  0.047812 -0.124096\n..                       ...               ...  ...       ...       ...\n93                  AR0052-C            LUA052  ... -0.025425  0.003523\n94                  AR0080-C            LUA080  ... -0.274052  0.051305\n95                  AR0076-C            LUA076  ... -0.141725 -0.279961\n96                  AR0066-C            LUA066  ...  0.632883  0.031980\n97                  AR0017-C            LUA017  ... -0.246536  0.037848\n\n[98 rows x 108 columns].columns
=========== 1 failed, 373 passed, 171 warnings in 221.99s (0:03:41) ============
```
I'll try rerunning again.

I'm not liking these random test fail/pass scenarios! I wonder we can do to make tests more consistent and deterministic.

leehart:
> This does also highlight a general weakness with setting `max_missing_an=0` as default for all datasets. In real datasets, as the number of samples gets larger, so the chance that you'll find variants with no missingness at all get smaller.
> 
> Rather it would be better to set a fractional value, i.e., to allow something like `max_missing_an=0.01` which would mean keep all variants with at most 1% missing genotype calls.

I suppose we should open a separate issue for that; it might be going a little beyond this mission. Or do you think we should try to fit it in here?

leehart:
> FWIW I would say the main hurdle here is that the default value of `max_missing_an` doesn't work very well for the amount of missingness in the simulated data.
> 
> Perhaps the best way to deal with that would be to provide a more sensible value of `max_missing_an` when calling functions within the tests, or adjust the simulated data to reduce the amount of missingness.

Even if I set `max_missing_an` to something ridiculous like `1_000_000_000_000`, I'm still getting the error. I must be missing something... Apart from brain cells!
```
================================ short test summary info =================================
FAILED tests/anoph/test_pca.py::test_pca_plotting[ag3_sim] - ValueError: Not enough SNPs.
FAILED tests/anoph/test_pca.py::test_pca_plotting[af1_sim] - ValueError: Not enough SNPs.
=================================== 2 failed in 19.00s ===================================
```

leehart:
> > To demonstrate or clarify, and perhaps help debug, the test failure mentioned above, i.e. `ValueError: Not enough SNPs.`, when keeping the default values as `None` for `biallelic_snp_calls()`, `biallelic_diplotypes()` and `biallelic_diplotype_pairwise_distances()`, I've created a draft PR here: #570
> 
> This error will be unrelated to the default values that are set for `max_missing_an` and `min_minor_ac` within the `biallelic_snps_calls()` or `biallelic_diplotypes()` functions.
> 
> The test triggering the error is `test_pca_plotting`. This calls `pca()` where there are default values set for `max_missing_an` and `min_minor_ac`. These values then get passed through when `pca()` calls `biallelic_diplotypes()` internally.

`raise ValueError("Not enough SNPs.")` is only in `biallelic_snp_calls()` from `snp_data.py`, which is called in `test_pca_plotting()` before `pca()` from `pca.py`.

leehart:
test_pca_plotting runs `biallelic_snp_calls()` with defaults as `None` without fail.

test_pca_plotting fails (for both `ag3_sim` and `af1_sim`) when running `pca()` with `min_minor_ac` set to `2` (default) and `max_missing_an` set to something ridiculously high, e.g. `1_000_000_000_000`. 

Curiously, tests don't fail when `min_minor_ac` is `0` and `max_missing_an` is `0`, but they do fail when `min_minor_ac` is `0` and `max_missing_an` is `1`. I would have expected `max_missing_an` as `1` to be more lenient than `max_missing_an` as `0`, i.e. it would allow more missingness, so I'm starting to suspect some sort of logic failure.

leehart:
Test failure under Python v3.11
```
=========================== short test summary info ============================
FAILED tests/anoph/test_pca.py::test_pca_plotting[af1_sim] - KeyError: 'PC3'
=========== 1 failed, 373 passed, 171 warnings in 161.34s (0:02:41) ============
```
Might be related to the recent `n_components = random.randint(2, min(n_samples, n_snps))`.
Succeeded after re-run, implying that this failure has a random component.

leehart:
@cclarkson This one's ready for (re)review.

cclarkson (approved):
LGTM, thanks @leehart

PR #575, opened 2024-08-02
Skip sample sets missing CNV HMM

leehart:
Re: issues #363, #555

~Alternative to PR #577~

### DIFF
--- a/malariagen_data/anoph/cnv_data.py
+++ b/malariagen_data/anoph/cnv_data.py
@@ -207,6 +207,12 @@ class AnophelesCnvData(
 
                     ly.append(y)
 
+                if len(ly) == 0:
+                    # Bail out, no data for given sample sets and analysis.
+                    raise ValueError(
+                        f"No data found for requested sample sets."
+                    )
+
                 debug("concatenate data from multiple sample sets")
                 x = simple_xarray_concat(ly, dim=DIM_SAMPLE)
 
--- a/malariagen_data/anoph/cnv_data.py
+++ b/malariagen_data/anoph/cnv_data.py
@@ -83,11 +83,10 @@ class AnophelesCnvData(
 
             # If CNV HMM data exists for this sample set then return the zarr,
             # Otherwise return None.
-            marker = path + "/.zmetadata"
-            if self._fs.exists(marker):
-                store = init_zarr_store(fs=self._fs, path=path)
+            store = init_zarr_store(fs=self._fs, path=path)
+            try:
                 root = zarr.open_consolidated(store=store)
-            else:
+            except FileNotFoundError:
                 root = None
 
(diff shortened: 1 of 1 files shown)


malariagen_data/anoph/cnv_data.py
```diff
@@ -80,8 +80,16 @@ def open_cnv_hmm(self, sample_set: base_params.sample_set) -> zarr.hierarchy.Gro
...
-            root = zarr.open_consolidated(store=store)
+
+            # If CNV HMM data exists for this sample set then return the zarr,
+            # Otherwise return None.
+            marker = path + "/.zmetadata"
+            if self._fs.exists(marker):
+                store = init_zarr_store(fs=self._fs, path=path)
+                root = zarr.open_consolidated(store=store)
+            else:
+                root = None
```
alimanfoo, line 91:
```suggestion
            store = init_zarr_store(fs=self._fs, path=path)
            try:
                root = zarr.open_consolidated(store=store)
            except FileNotFoundError:
                root = None
```

Suggestion above follows approach taken in `open_haplotypes()`.

It's a bit better to ask forgiveness here rather than permission. The reason is that an `exists()` check will add an extra HTTP request which could take maybe 100ms, and these could add up if you have lots of sample sets to check.

malariagen_data/anoph/cnv_data.py
```diff
@@ -189,6 +201,11 @@ def cnv_hmm(
                         inline_array=inline_array,
                         chunks=chunks,
                     )
+
+                    # If no CNV HMM dataset was found then skip
+                    if y is None:
+                        continue
+
                     ly.append(y)
```
alimanfoo, line 210:
```suggestion

                    # If no CNV HMM dataset was found then skip
                    if y is None:
                        continue

                    ly.append(y)

                if len(ly) == 0:
                    # Bail out, no data for given sample sets and analysis.
                    raise ValueError(
                        f"No data found for requested sample sets."
                    )

```

Comparing with the `haplotypes()` function, there is an extra check here just in case there is no data at all for the requested sample sets.

alimanfoo:
Hi @leehart, I think I'd favour this approach over the alternative in #577, just because what happens here parallels what happens for haplotype data, and will be more convenient for the user.

malariagen_data/anoph/cnv_data.py
leehart: Thanks @alimanfoo , will do. I agree with the reason but not the maxim!

malariagen_data/anoph/cnv_data.py
leehart: Thanks @alimanfoo , will do.

alimanfoo (approved):
LGTM :)

PR #576, opened 2024-08-02
Raise ValueError for missing CNV DRCs

leehart:
Re: issue #551

### DIFF
--- a/malariagen_data/anoph/cnv_data.py
+++ b/malariagen_data/anoph/cnv_data.py
@@ -256,7 +256,7 @@ class AnophelesCnvData(
             marker = path + "/.zmetadata"
             if not self._fs.exists(marker):
                 raise ValueError(
-                    f"CNV coverage calls analysis f{analysis!r} not implemented for sample set {sample_set!r}"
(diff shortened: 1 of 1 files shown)


alimanfoo:
Hi @leehart, maybe it would be better to handle this the same way as for HMM CNV data in #575? This will be more consistent internally, and in the extreme case where funestus never has an DRC data, it will still raise an error (if you add in the additional check I suggested [here](https://github.com/malariagen/malariagen-data-python/pull/575#discussion_r1711804403)).

leehart:
@alimanfoo Agreed. At one time it seemed that returning `None` would be consistent, and then raising `ValueError` like this would be consistent, but now it looks like everything is falling into line, along the same one true track, moving in the right direction! 👍

leehart:
Plan to handle this the same way as for HMM data in #575 and [`haplotypes()`](https://github.com/malariagen/malariagen-data-python/blob/3b1640b0acdf61b2c844ea8ae650ac0219ced340/malariagen_data/anoph/hap_data.py#L326).

PR #577, opened 2024-08-02
Raise ValueError for missing CNV HMM

leehart:
Re: issue #363

Alternative approach to PR #575

Regarding issue #555, this would still require manual filtering of sample sets that don't have CNV HMM.

The user might not know ahead of time which sample sets have or don't have CNV HMM.


leehart:
Closing in favour of PR #575 , which is more consistent with [`haplotypes()`](https://github.com/malariagen/malariagen-data-python/blob/3b1640b0acdf61b2c844ea8ae650ac0219ced340/malariagen_data/anoph/hap_data.py#L326) and more convenient for the user.

PR #579, opened 2024-08-09
Fix and improve plotting pairwise average fst

### LINKED ISSUE #567: FST plotting function error
The FST function should give the standard error or Z score when called on the diagonal. However the first row is swapped with the first column. Please see the figures below. The first is the plot with just FST and the second is a plot which should show Z score on the diagonal. 

Screenshot 2024-07-10 at 16 22 50

Screenshot 2024-07-10 at 16 24 29

alimanfoo:
* Resolves #567.
* Removes warnings associated with pandas chained assignments.
* Dynamically sizes plot based on number of cohorts.
* Adds a title to the plot.
* Disable colouring by default if an annotation is requested in the upper triangle (as the table then contains a mix of two different types of values and the colouring becomes confusing).
* Code simplification.

### DIFF
--- a/malariagen_data/anoph/fst.py
+++ b/malariagen_data/anoph/fst.py
@@ -485,32 +485,49 @@ class AnophelesFstAnalysis(
         zmax: Optional[plotly_params.zmax] = None,
         text_auto: plotly_params.text_auto = ".3f",
         color_continuous_scale: plotly_params.color_continuous_scale = "gray_r",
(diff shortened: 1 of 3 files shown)


alimanfoo:
Applies the fix from #540 to resolve the error raised by @KellyLBennett in #567. Here's the fixed plot:

image

ahernank (approved):

alimanfoo:
Thanks @leehart and @ahernank

Issue #583, opened 2024-08-16
Colors in GWSS plots


jonbrenas:
@alimanfoo mentioned that we now offer the possibility to plot the various WGSS analyses for whole chromosomes (and even for the whole genome) but that, because they are all drawn the same way, it is hard to see where chromosome arms change. The idea would be to have a different colour for each one.

jonbrenas:
Looking at this, I have several ideas for small enhancements:

1) Allowing users more freedom with how the stats are represented by letting them select the markers, their colour(s), their sizes, ... This is fairly straightforward.
2) The above-mentioned use of different parameters for different contigs (or even maybe regions?)
3) The option to have more than one trace on the same plot (is that going to be readable though?). This might be a different function.
4) Maybe an option to plot the difference between two traces to see differential signals of selection. This would probably be a new function.

alimanfoo:
Hi @jonbrenas, thanks for raising this. Some comments inline...

> Looking at this, I have several ideas for small enhancements:
> 
> 1. Allowing users more freedom with how the stats are represented by letting them select the markers, their colour(s), their sizes, ... This is fairly straightforward.

This seems like a good idea. However, the parameters would need to account for point 2 below somehow, i.e., how does the user control what happens when multiple contigs are being shown.

> 2. The above-mentioned use of different parameters for different contigs (or even maybe regions?)

I would suggest to focus this issue on being able to plot different colours for different contigs, and to have this be the default behaviour.

> 3. The option to have more than one trace on the same plot (is that going to be readable though?). This might be a different function.

I'd suggest raising as a separate issue - it is a common requirement to want to make a figure with multiple GWSS tracks above each other.

> 4. Maybe an option to plot the difference between two traces to see differential signals of selection. This would probably be a new function.

Eric Lucas has in the past used the difference of H12 between two cohorts. Suggest to raise as a separate issue.

jonbrenas:
Thanks @alimanfoo . I will create the extra issues.

As for this one:

>> Looking at this, I have several ideas for small enhancements:
>> 
>> 1. Allowing users more freedom with how the stats are represented by letting them select the markers, their colour(s), their sizes, ... This is fairly straightforward.
>
>This seems like a good idea. However, the parameters would need to account for point 2 below somehow, i.e., how does the user control what happens when multiple contigs are being shown.
>
>> 2. The above-mentioned use of different parameters for different contigs (or even maybe regions?)
>
[...]

Issue #554, opened 2024-06-20
codecov/project CI check fails but seems to require additional coverage on code unrelated to the PR


leehart:
I think it's happened a few times where a PR has been submitted but the CI checks for code coverage appears to fail on unrelated code, or the only way to satisfy the requirement is to add more test coverage to code not related to the PR. Or perhaps there's some other requirement that isn't clear?

I'm not sure if this is an issue with the way the CI is configured or a problem associated with the coverage tests themselves, but it always seems unfair and inappropriate to require additional code coverage tests to be added to parts of the codebase that are not related to the original PR.

It causes PRs to get stuck in limbo, despite having passed all other reasonable requirements, e.g. #540

alimanfoo:
I think codecov is using the wrong base. This possibly relevant: https://community.codecov.com/t/set-base-commit/1014/4

PR #591, opened 2024-08-29
Colors in GWSS plots

jonbrenas:
Adresses #583.

This commit works on the first point, more discussion is needed for the rest.

### DIFF
--- a/malariagen_data/anoph/h12.py
+++ b/malariagen_data/anoph/h12.py
@@ -325,6 +325,7 @@ class AnophelesH12Analysis(
         sizing_mode: gplt_params.sizing_mode = gplt_params.sizing_mode_default,
         width: gplt_params.width = gplt_params.width_default,
         height: gplt_params.height = 200,
+        circle_kwargs: Optional[gplt_params.circle_kwargs] = None,
         show: gplt_params.show = True,
         x_range: Optional[gplt_params.x_range] = None,
         output_backend: gplt_params.output_backend = gplt_params.output_backend_default,
@@ -377,15 +378,22 @@ class AnophelesH12Analysis(
             output_backend=output_backend,
         )
 
+        circle_kwargs_mutable = dict(circle_kwargs) if circle_kwargs else {}
+        circle_kwargs_mutable["size"] = circle_kwargs_mutable.get("size", 3)
+        circle_kwargs_mutable["line_width"] = circle_kwargs_mutable.get("line_width", 1)
+        circle_kwargs_mutable["line_color"] = circle_kwargs_mutable.get(
+            "line_color", "black"
+        )
+        circle_kwargs_mutable["fill_color"] = circle_kwargs_mutable.get(
+            "fill_color", None
+        )
+
         # Plot H12.
         fig.scatter(
             x=x,
             y=h12,
             marker="circle",
-            size=3,
-            line_width=1,
-            line_color="black",
-            fill_color=None,
+            **circle_kwargs_mutable,
         )
 
         # Tidy up the plot.
@@ -422,6 +430,7 @@ class AnophelesH12Analysis(
         sizing_mode: gplt_params.sizing_mode = gplt_params.sizing_mode_default,
         width: gplt_params.width = gplt_params.width_default,
         track_height: gplt_params.track_height = 170,
+        circle_kwargs: Optional[gplt_params.circle_kwargs] = None,
         genes_height: gplt_params.genes_height = gplt_params.genes_height_default,
         show: gplt_params.show = True,
         output_backend: gplt_params.output_backend = gplt_params.output_backend_default,
@@ -441,6 +450,7 @@ class AnophelesH12Analysis(
             sizing_mode=sizing_mode,
             width=width,
             height=track_height,
+            circle_kwargs=circle_kwargs,
             show=False,
             output_backend=output_backend,
         )
--- a/malariagen_data/anoph/h1x.py
+++ b/malariagen_data/anoph/h1x.py
@@ -166,6 +166,7 @@ class AnophelesH1XAnalysis(
         sizing_mode: gplt_params.sizing_mode = gplt_params.sizing_mode_default,
         width: gplt_params.width = gplt_params.width_default,
         height: gplt_params.height = 200,
+        circle_kwargs: Optional[gplt_params.circle_kwargs] = None,
         show: gplt_params.show = True,
         x_range: Optional[gplt_params.x_range] = None,
         output_backend: gplt_params.output_backend = gplt_params.output_backend_default,
@@ -219,15 +220,22 @@ class AnophelesH1XAnalysis(
             output_backend=output_backend,
         )
 
+        circle_kwargs_mutable = dict(circle_kwargs) if circle_kwargs else {}
+        circle_kwargs_mutable["size"] = circle_kwargs_mutable.get("size", 3)
+        circle_kwargs_mutable["line_width"] = circle_kwargs_mutable.get("line_width", 1)
+        circle_kwargs_mutable["line_color"] = circle_kwargs_mutable.get(
+            "line_color", "black"
+        )
+        circle_kwargs_mutable["fill_color"] = circle_kwargs_mutable.get(
+            "fill_color", None
+        )
+
         # Plot H1X.
         fig.scatter(
             x=x,
             y=h1x,
             marker="circle",
-            size=3,
-            line_width=1,
-            line_color="black",
-            fill_color=None,
+            **circle_kwargs_mutable,
         )
 
         # Tidy up the plot.
@@ -268,6 +276,7 @@ class AnophelesH1XAnalysis(
         sizing_mode: gplt_params.sizing_mode = gplt_params.sizing_mode_default,
         width: gplt_params.width = gplt_params.width_default,
         track_height: gplt_params.track_height = 190,
+        circle_kwargs: Optional[gplt_params.circle_kwargs] = None,
         genes_height: gplt_params.genes_height = gplt_params.genes_height_default,
         show: gplt_params.show = True,
         output_backend: gplt_params.output_backend = gplt_params.output_backend_default,
@@ -288,6 +297,7 @@ class AnophelesH1XAnalysis(
             sizing_mode=sizing_mode,
             width=width,
             height=track_height,
+            circle_kwargs=circle_kwargs,
             show=False,
             output_backend=output_backend,
         )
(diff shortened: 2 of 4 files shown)


leehart:
Thanks @jonbrenas . Just checking whether this PR is ready for review, or if there is more to do on this part (in which case I might convert it to draft) ?

leehart:
@jonbrenas

leehart:
@leehart

jonbrenas:
OK, so it looks like I dreamt that I had added the code to this PR. I am going to fish for it.

jonbrenas:
There are still some issues. Some are pretty minor (the colour map is hard-coded within the function which is dumb, the same needs to be done for the other GWSS functions, ...) but there is something that we need to discuss: the "different colours for different contigs"-thing and the "users can choose how their plots look"-thing don't work particularly well together yet and we need to decide what the parameters should look like.

leehart:
Thanks @jonbrenas . Yeah, it would be nice to get the default `color_dict` into a better place. I wonder if there are existing colour palette mechanisms that can be copied, for consistency? I notice `plot_ihs_gwss_track` supports a `palette` arg:
```python
palette: ihs_params.palette = ihs_params.palette_default,
```
```python
palette_default: palette = "Blues"
````

Regarding supporting both alternating contig colours and user-specified settings, what is the current situation with this PR's code? For instance, if the user doesn't specify any `circle_kwargs` then I guess `plot_g123_gwss_track()` should show contigs in alternating colours. Whereas, if the user does want to override the default behaviour, even just to change the colours, then I guess they would specify `circle_kwargs` accordingly, and they would need to know which settings. Is that right? Are you able to provide a couple of examples here, or maybe we need to chat about that specifically?

**Comment by jonbrenas** [comment]
I am having troubles with the `mypy` part of the linting tests. More specifically, I want the user to be able to specify the `circle_kwargs` for the various contigs, called `circle_kwargs_param`, in one of the following ways:

1. As a `dict[int, circle_kwargs]` as `contigs` are returned as `int`s by, for example, `ag3.haplotypes(...)`
2. As a `dict[str, circle_kwargs]` with the `contigs` names as keys
3. As a single `circle_kwargs` that is going to be used everywhere
4. As a `list[circle_kwargs]` where the first value is used for contig 0 and so on.

The problem is the first 3 options are, basically, `dict[X,Y]` and when mypy sees code such as:
```
circle_kwargs_param_dict[i] = circle_kwargs_param["2R"]
```
it returns:
```
malariagen_data/anoph/h12.py:495: error: Invalid index type "str" for "dict[Any, Any] | dict[int, Mapping[Any, Any]] | dict[str, Mapping[Any, Any]]"; expected type "int"  [index]
```
and when it sees code such as:
```
circle_kwargs_param_dict[i] = circle_kwargs_param[1]
```
it returns:
```
malariagen_data/anoph/h12.py:549: error: Invalid index type "int" for "dict[Any, Any] | dict[int, Mapping[Any, Any]] | dict[str, Mapping[Any, Any]]"; expected type "str"  [index]
```

By the way, the tests run fine with all 3 dict options, it is just the type checking that crashes the tests.

@leehart , @cclarkson , @alimanfoo , @ahernank, any one has an idea of what I am doing wrong (or of what I can do to prevent mypy from panicking?)?

**Comment by leehart** [comment]
Hi @jonbrenas . Aside from the data type problems, maybe it's just me but the options for the end user seems a bit complicated, although it does also seem very accommodating.

My feeling is still, in light of the issue at hand, in agreement with @alimanfoo 's suggestion, that there are probably only a few scenarios to cater for:
1. The user doesn't care about the particular colours, as long as they can distinguish adjacent contigs, so maybe we should default to 2 contrasting colours, preferably consistent with other colour schemes/themes used in the package.
2. The user wants everything to be particular colour, so maybe we need to provide a way to override the default 2 colours with 1 user-specified colour. I suppose this is equivalent to the current behaviour.
3. The user wants to specify their own set of colours (like a palette - I didn't mean to suggest the ihs one specifically!) that still alternates per contig, but any number of colours, again overriding the default 2 colours. This option already seems like a nice-to-have feature, to me, given the original issue.

Sorry I haven't had time to look at this more closely yet; that could be exactly what your changes actually gives to the user!

In any case, I'm still worried that the end user won't be able to easily figure out how to use an argument named `circle_kwargs_param` with multiple different data types, in order to just apply a slight tweak to the plot colours. I get the feeling I'm overlooking some important requirements or failing to appreciate some subtle complications here! I'll try to look at this more closely this week... tomorrow. I can see that the package already uses `circle_kwargs` params in various places, so I gather this name choice is a consistency thing.

**Comment by jonbrenas** [comment]
Thanks @leehart . 

Case 1 is the default, except that there are 5 different colours because users could request to plot '23X' in a row. Alternating colours would be possible (that's what is done typically for plasmodium) but 5 colours (or 3 for funestus) is still manageable and clearer IMHO.

Case 2 is the current default. I personally don't really like it. It can be achieved by giving only one `circle_kwargs` that can be as simple as `{'color': 'black'}`, for instance.

Case 3 is where things get trickier. The main issue is that there are (at least) 2 uses for (almost) every function in `malariagen_data`: exploratory and showcase. The typical example is `snp_allele_frequencies`. In the exploratory use, you run the code with only the transcript that you are interested in and you get an idea of what you are looking at. But, you get a lot of SNPs that are of no interest, the cohorts are not exactly what you want, ... so you have to define your own cohorts and do some filtering before you can use it as a result. The idea is to had "showcase" functionalities to the code so that users don't have to copy/paste the code for the functions with the one change being how the plots are displayed. It is a little egotistical, maybe, but I am starting to have a hard time choosing between the many copies of h12 that I have in various notebooks to know which one I want to use in each analysis.

That said, I agree with you that `circle_kwargs_param` is a complex parameter but it is not one that most users will ever try to use, it is only for people who were going to use that kind of parameter when they recreated the code and needed to give the correct parameters to bokeh. In the same way that I am sure that 99% of the trainees at our workshop have never redefined the cohorts used in `snp_allele_frequencies` despite the fact that they have all ran the function dozens of times on data that would have definitely benefited from better defined cohorts because they didn't see the need (yet).

**Comment by alimanfoo** [comment]
Hi @jonbrenas, thanks so much, I think this is moving in a good direction :)

We could simplify a little further here I think, because rather than create new functions `h12_gwss_contig()` and `h1x_gwss_contig()` I think it would be fine just to change the return value from the existing `h12_gwss()` and `h1x_gwss()` functions to add the contig values as a third array.

Most users will rarely if ever call these functions directly, rather they will call `plot_h12_gwss()` instead, and so changing the return values of `h12_gwss()` and `h1x_gwss()` I don't think will impact much.

**Comment by jonbrenas** [comment]
Thanks @alimanfoo . I indeed created these functions to be backwards-compatible with the idea that we would move to the solution you proposed for the next major release (I thought that it was the idea, that minor releases and patches do not break any previously working code but that major releases can). That said, I am not sure I would have remembered ;) !

**Comment by alimanfoo** [comment]
> Thanks @alimanfoo . I indeed created these functions to be backwards-compatible with the idea that we would move to the solution you proposed for the next major release (I thought that it was the idea, that minor releases and patches do not break any previously working code but that major releases can). That said, I am not sure I would have remembered ;) !

Thanks @jonbrenas. I'm happy to change the API and make a major release release. In general I don't think making a major release is fine, the main thing is to be aware that we have made an API change and so a major version bump is needed. Also it is good to think about whether the API change will impact lots of user code, and so only do it if there is clear value. But in this case I think the impact to user code will be minimal.

**Comment by jonbrenas** [comment]
Thanks @alimanfoo. This should be better now.

**Comment by leehart** [comment]
@jonbrenas to resolve conflicts and provide examples here and in notebooks.

**Comment by jonbrenas** [comment]
Here are some pictures from the new cells in the notebook:

H12 (default colours):
Screenshot 2024-11-05 at 15 39 12

H1X (manual choice of colours):
Screenshot 2024-11-05 at 15 45 59

I didn't modify the part for Af1 as there is no option for multiple contigs (we might want to colour 2L and 2R differently for example but that would be more complex because they are 1 contig for funestus).

**Comment by jonbrenas** [comment]
@alimanfoo, @leehart, I am having trouble with codecov. I don't think the part that is not covered is an issue as it is exclusively code that was already in master.

**Comment by leehart** [comment]
@jonbrenas I think we have ignored and merged regardless before, e.g. #647, but I reckon it's a persistent issue, i.e. #554

**Comment by alimanfoo** [comment]
> Here are some pictures from the new cells in the notebook

Awesome, looking good!

Could you post an example of what it looks like with default colours if you do `region="23X"` just for interest, i.e., whole genome.

File: malariagen_data/anoph/h12.py [hunk 1]
```diff
@@ -297,10 +306,10 @@ def h12_gwss(
         random_seed: base_params.random_seed = 42,
         chunks: base_params.chunks = base_params.native_chunks,
         inline_array: base_params.inline_array = base_params.inline_array_default,
-    ) -> Tuple[np.ndarray, np.ndarray]:
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
         # Change this name if you ever change the behaviour of this function, to
         # invalidate any previously cached data.
-        name = "h12_gwss_v1"
+        name = "h12_gwss_contig_v1"
```
alimanfoo, line 312:
```suggestion
        name = "h12_gwss_v2"
```

Nit, convention is to use the function name then add a version number.

malariagen_data/anoph/h1x.py
```diff
@@ -119,10 +128,10 @@ def h1x_gwss(
         random_seed: base_params.random_seed = 42,
         chunks: base_params.chunks = base_params.native_chunks,
         inline_array: base_params.inline_array = base_params.inline_array_default,
-    ) -> Tuple[np.ndarray, np.ndarray]:
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
         # Change this name if you ever change the behaviour of this function, to
         # invalidate any previously cached data.
-        name = "h1x_gwss_v1"
+        name = "h1x_gwss_contig_v1"
```
alimanfoo, line 134:
```suggestion
        name = "h1x_gwss_v2"
```

alimanfoo (commented):
Looks great, just a couple of really tiny suggestions.

alimanfoo:
Do you notice any performance difference at all from this change? Just curious, if you aren't noticing any obvious slowdown then I'm not worried.

alimanfoo:
> @jonbrenas I think we have ignored and merged regardless before, e.g. #647, but I reckon it's a persistent issue, i.e. #554

Yeah I think we can ignore the codecov/project check for now. The main thing is that the codecov/patch check is passing, which shows you have good coverage of the code that is being changed in this PR.

A previous PR reduced overall coverage for the project slightly, and the codecov/project check keeps complaining about this in every subsequent PR. Feels like there should be some way to reset the base for the codecov/project check, but not sure how.

Issue #548, opened 2024-06-11
explanation of site_annotations output


EricRLucas:
Please could the the docs include information on how to interpret the output of site_annotations()? I can't work out what the different levels of codon_nonsyn mean (effect?). I also can't quite work out what degeneracy shows (all 4-fold degenerate positions have a 0 score for nonsyn, which makes sense, but none of the other degeneracies have a 0; I would have thought that a 3-fold degenerate position would have a lot of 0 values).

leehart:
Internal ref: https://github.com/malariagen/vector-ops/issues/1278

jonbrenas:
I agree that the docs could use more details as I was quite unsure what the various values meant ... and I think I am the one who generated the file that gets read for gambiae. I want to check that my understanding is correct before I change the docs, though, so I hope @alimanfoo and @cclarkson can check my homework.

Currently, the "Returns"   part just says:

```
Dataset
A dataset of site annotations.
```

My suggestion would be something like:

```
Dataset
A dataset of site annotations containing 7 variables:
seq_cls: The feature class. There are 11 possible values:
        1: Upstream
        2: Downstream
        3: 5' UTR
        4: 3' UTR
        5: CDS (first)
        6: CDS (mid)
        7: CDS (last)
        8: Intron (first)
        9: Intron (mid)
        10: Intron (last)
        0: Unknown
seq_flen: The length of the feature.
seq_relpos_start: Relative position to the start of the feature. 0 if not in a feature.
seq_relpos_stop: Relative position to the end of the feature. 0 if not in a feature.
codon_position: Position within a triplet codon. -1 if not in a CDS.
codon_nonsyn: Number of different amino acids that can be obtained by changing this codon (does not include the amino acid that is encoded by the triplet codon as is). 0 if not in a CDS.
codon_degeneracy: The redundancy of the codon. Can take 5 different values:
        1: 0-fold degenerate, i.e., all nucleotides encode different amino acids
        2: simple 2-fold degenerate, i.e., 2 different amino acids can be encoded depending on which nucleotide is present; each amino acid is encoded by two different nucleotides
        3: complex 2-fold degenerate, i.e., 2 different amino acids can be encoded depending on which nucleotide is present but they are not paired
        4: 4-fold degenerate, i.e., all nucleotides encode the same amino acid
        -1: not in a CDS
```

Is that correct? Is that understandable?

alimanfoo:
[...]

PR #592, opened 2024-09-02
Expansion of the documentation

### LINKED ISSUE #662: Adding detailed descriptions of the DataFrames and Datasets returned by the various functions
Supersedes #548,  #553 and maybe more.

In almost all cases the description is very short and not very helpful, e.g.:

> A dataframe of sample metadata, one row per sample.

jonbrenas:
Resolves #662.

Quite a few functions are shown in the documentation as returning a DataFrame or a Dataset. It is not always obvious what the returned structure contains (e.g., how many columns in the DataFrame, how to interpret their contents, ...). I have given it a try for  `sample_sets` (one of the simplest functions) but I am not quite sure how to test what the result looks like. Ideally, I would have had the return be a String (roughly the current documentation) followed by a dict of the columns but that is not an option supported by numpydoc_decorator (I guess I could ask its creator if that can be changed or if he has a better idea @alimanfoo ).


jonbrenas:
Looking at some of the functions that return DataFrames (e.g., `sample_metadata()`), the columns are dependent on whether it is the Ag3 or Af1 version that is run. They have different pages in the documentation but the same code definition and thus the same `@doc` text.

jonbrenas:
I have s small issue with cohorts(). The docs say that it accepts as a parameter `str | Mapping[str, str]`(because it is the standard type of a cohort) but it really only accepts the 6 pre-defined cohorts specifications.

leehart:
Re: issue #548

jonbrenas:
There were some cases where the datasets were described (e.g., `snp_allele_frequencies_advanced`). I left them as is. They are likely to be stylistically different but I doubt I have been consistent through the last few months so there will probably be some work needed during review to harmonise everything.

jonbrenas:
I am done with this first run through. It is possible that new functions need a new doc too but I feel like it is a good time to ask for a bit of a review.

leehart (approved):
Thanks for all this work @jonbrenas ! 
It will be good to see how this actually manifests in the deployed docs. Did you establish a way to preview it? Then we can do another round of additions and tweaks, at some point.

malariagen_data/anoph/cnv_data.py
```diff
@@ -377,7 +389,19 @@ def _cnv_coverage_calls_dataset(
...
+        **variants** the number of CNV regions in the selected region,
+        **samples** the number of samples. There are 5 coordinates:
+        **variant_position** has **variants** values and contains the initial position of each CNV region,
+        **variant_end** has **variants** values and contains the final position of each CNV region,
+        **variant_contig** has **variants** values and contains the contig of each CNV region,
+        **variant_id** has **variants** values and contains the identifier for each CNV region,
+        **sample_id** has **samples** values and contains the identifier of each sample. It contains 4 data variables:
+        **variant_CIPOS**, it has **variants** values and contains ??? for each CNV region,
+        **variant_CIEND**, it has **variants** values and contains ??? for each CNV region,
+        **variant_filter_pass**, it has **variants** values and contains ??? for each CNV region,
```
alimanfoo, line 402:
```suggestion
        **variant_CIPOS**, it has **variants** values and contains the confidence interval for the start position for each CNV region,
        **variant_CIEND**, it has **variants** values and contains the confidence interval for the end position for each CNV region,
        **variant_filter_pass**, it has **variants** values and is True for each CNV region that passes quality filters,
```

malariagen_data/anoph/cnv_data.py
```diff
@@ -533,7 +557,21 @@ def _cnv_discordant_read_calls_dataset(
...
+        **variants** the number of discordant read calls in the selected region,
+        **samples** the number of samples. There are 5 coordinates:
+        **variant_position** has **variants** values and contains the initial position of each discordant read call,
+        **variant_end** has **variants** values and contains the final position of each discordant read call,
+        **variant_id** has **variants** values and contains the identifier of each discordant read call,
+        **variant_contig** has **variants** values and contains the contig of each discordant read call,
+        **sample_id** has **samples** values and contains the identifier of each sample. It contains 6 data variables:
+        **variant_Region**, it has **variants** values and contains the identifier of the region covered by each discordant read call,
+        **variant_StartBreakpointMethod**, it has **variants** values and contains ? for each discordant read call,
+        **variant_EndBreakpointMethod**, it has **variants** values and contains ? for each discordant read call,
```
alimanfoo, line 570:
```suggestion
        **variant_StartBreakpointMethod**, it has **variants** values and specifies how the start breakpoint was determined for each discordant read call,
        **variant_EndBreakpointMethod**, it has **variants** values and specifies how the end breakpoint was determined for each discordant read call,
```

malariagen_data/anoph/genome_features.py
```diff
@@ -119,7 +119,7 @@ def _prep_gff_attributes(
     @check_types
     @doc(
         summary="Access genome feature annotations.",
-        returns="A dataframe of genome annotations, one row per feature.",
+        returns="A dataframe of genome annotations, one row per feature. The dataframe follows the GFF3 format (https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md), including **ID**, **Parent**, **Name** (only for *Ag3*, replaced by **Note** which gives the Enzyme Commission (EC) number and associated nomenclature when relevant), and **description** (replacing Note) as extra attributes.",
```
alimanfoo, line 122:
```suggestion
        returns="A dataframe of genome annotations, one row per feature. The dataframe follows the GFF3 format (https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md), including extra attributes **ID**, **Parent**, **Name** and **description** depending on the dataset.",
```

Maybe a little simpler?

malariagen_data/anoph/sample_metadata.py
```diff
@@ -584,7 +642,7 @@ def sample_metadata(
             index="Sample metadata columns to use for the pivot table index.",
             columns="Sample metadata columns to use for the pivot table columns.",
         ),
-        returns="Pivot table of sample counts.",
+        returns="Pivot table of sample counts. One raw per admin2_year cohort. Unless otherwise specified using the `columns` parameters, the samples are grouped according to their taxon and then counted.",
```
alimanfoo, line 645:
```suggestion
        returns="Pivot table of sample counts. One row per admin2_year cohort. Unless otherwise specified using the `columns` parameters, the samples are grouped according to their taxon and then counted.",
```

malariagen_data/anopheles.py
```diff
@@ -849,8 +849,22 @@ def gene_cnv(
...
+        **gene_id** has **genes** values and contains the identifier of each gene,
+        **sample_id** has **samples** values and contains the identifier of each sample. It contains 11 data variables:
+        **gene_contig**, it has **genes** values and contains the contig of each gene,
+        **gene_start**, it has **genes** values and contains the position of the start of each gene,
+        **gene_end**, it has **genes** values and contains the position of the end of each gene,
+        **gene_windows**, it has **genes** values and contains the number of windows of each gene,
+        **gene_name**, it has **genes** values and contains the name of each gene,
+        **gene_strand**, it has **genes** values and contains on which strand each gene is found,
+        **gene_description**, it has **genes** values and contains the description of each gene,
+        **CN_mode**, it has (**genes**, **samples**) values and contains ? for each gene and each sample,
```
alimanfoo, line 864:
```suggestion
        **CN_mode**, it has (**genes**, **samples**) values and contains the modal copy number for each gene and each sample,
```

alimanfoo (commented):
Wow, so much work here @jonbrenas!

Added a couple of suggestions to fill in blanks.

Only other suggestion is to consistently use either \*\*foo\*\* or \`foo\` when referring to column/variable/dimension names. I'd have a small preference for \`foo\` as it's a bit less distracting when reading the docs via `help()`.

malariagen_data/anoph/aim_data.py
```diff
@@ -65,7 +65,11 @@ def _prep_aims_param(self, *, aims: aim_params.aims) -> str:
     @check_types
     @doc(
         summary="Access ancestry informative marker variants.",
-        returns="A dataset containing AIM positions and discriminating alleles.",
+        returns="""
+        A dataset with 2 dimensions: **variants** the number of AIMs sites, and **alleles** which will always be 2, each representing one of the species. It contains 2 coordinates:
+        **variant_contig** has **variants** values and contains the chromosome arm of each AIM, and **variant_position** has **variants** values and contains the position of each AIM. It contains 1 data variable:
+        **variant_allele** has (**variants**, **allele**) values and contains the discriminating alleles for each AIM.
```
alimanfoo, line 71: Swap \*\* for \`.

malariagen_data/anoph/aim_data.py
```diff
@@ -113,7 +117,16 @@ def _aim_calls_dataset(self, *, aims, sample_set):
...
+        A dataset with 4 dimensions:
+        **variants** the number of AIMs sites,
+        **samples** the number of samples,
+        **ploidy** the ploidy (2),
+        and **alleles** which will always be 2, each representing one of the species. It contains 3 coordinates:
+        **sample_id** has **samples** values and contains the identifier of each sample,
+        **variant_contig** has **variants** values and contains the chromosome arm of each AIM,
+        and **variant_position** has **variants** values and contains the position of each AIM. It contains 2 data variables:
+        **call_genotype** has (**variants**, **samples**, **ploidy**) values and contains both calls for each sample and each AIM,
+        **variant_allele** has (**variants**, **allele**) values and contains the discriminating alleles for each AIM.
```
alimanfoo, line 129: Swap \*\* for \`.

malariagen_data/anoph/base.py
```diff
@@ -397,7 +397,16 @@ def _read_sample_sets(self, *, single_release: str):
...
-        returns="A dataframe of sample sets, one row per sample set.",
+        returns="""A dataframe of sample sets, one row per sample set. It contains five columns:
+         **sample_set** is the name of the sample set,
+         **sample_count** is the number of samples the sample set contains,
+         **study_id** is the identifier for the study that generated the sample set,
+         **study_url** is the URL of the study on the MalariaGEN website,
+         **term_of_use_expiry** is the date when the terms of use expire,
+         **terms_of_use_url** is the URL of the terms of use,
+         **release** is the identifier of the release containing the sample set,
+         **unrestricted_use** whether the sample set can be without restriction (e.g., if the terms of use have expired).
```
alimanfoo, line 408: Swap \*\* for \`.

alimanfoo (changes requested):
Noticed a couple of files where still using \*\*...

malariagen_data/anoph/base.py
jonbrenas: Yep, it took me 3 files to figure out how to do a find and replace on my Mac ;). I should have rechecked the files where I used another method.

alimanfoo (approved):

alimanfoo:
Nice one, thanks @jonbrenas 🙏

PR #593, opened 2024-09-02
Update GCS bucket to use single-region as default

### LINKED ISSUE #585: Support single-region buckets only
As part of the move to support authenticated access in GCS, we have also assessed our use of multi-region vs single-region buckets. Currently, both buckets are supported by the API. 

After some testing, ensuring the change won't have major impacts to speed, we will move towards supporting the single-region buckets only. We need to reflect this on the API configuration.

ahernank:
* Resolves #585

I've had a couple of goes at changing this but actually ended up thinking that we could just update the default, and keep all the "machinery" for regional buckets switching within the API, in case we ever decide to mirror or create more buckets (e.g. within the South Africa region)..


alimanfoo (approved):
LGTM, thanks @ahernank.

PR #595, opened 2024-09-05
Plot H12 GWSS traces for multiple cohorts together in a single figure

### LINKED ISSUE #594: Multiple H12 traces
It is not yet clear to me if it is better to have several linked tracks or several traces on the same track. So I attempted both.

jonbrenas:
This PR adds new functions for performing H12 GWSS for multiple cohorts and plotting the results together in a single figure. 

* Adds a `plot_h12_gwss_multi_overlay()` function to plot multiple cohorts overlaid in a single track.
* Adds a `plot_h12_gwss_multi_panel()` function to plot multiple cohorts in separate tracks with linked pan and zoom.

Resolves #594.


leehart:
Thanks @jonbrenas . I'll convert this to draft because it's still "work in progress". Press the "ready for review" button when it's ready, of feel free to ping reviewers.

leehart:
Thanks @jonbrenas . Do you have some example code (or maybe screenshots) to help me understand this better, then I can try to give it a quick review?

jonbrenas:
Thanks @leehart. Here is the code that I showed previously. It should be pretty close to what ended up in the files.

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,269 @@ def plot_h12_gwss(
...
             return fig
 
+    @check_types
+    @doc(
+        summary="Plot h12 GWSS data track with multiple traces.",
+    )
+    def plot_h12_gwss_track_multi(
+        self,
+        contig: base_params.contig,
+        sample_queries: h12_params.sample_queries,
```
alimanfoo, line 524:
```suggestion
        cohorts: base_params.cohorts,
```

Suggest to use a "cohorts" parameter instead, which is the same as already used elsewhere in the API when multiple cohorts are required. 

This could then be handled via the `_setup_cohort_queries()` helper function, like is done within the `diversity_stats()` method.

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,269 @@ def plot_h12_gwss(
...
+        for i, (x, h12) in enumerate(res):
+            fig.scatter(
+                x=x,
+                y=h12,
+                marker="circle",
+                size=3,
+                line_width=1,
+                line_color=colors[i % len(colors)],
+                fill_color=None,
+                legend_label=sample_queries[i],
```
alimanfoo, line 600: Could use cohort identifiers here rather than sample queries, would be more compact.

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,269 @@ def plot_h12_gwss(
...
+            return None
+
+    @check_types
+    @doc(
+        summary="Plot h12 GWSS data with multiple tracks.",
+    )
+    def plot_h12_gwss_multi(
+        self,
+        contig: base_params.contig,
+        sample_queries: h12_params.sample_queries,
```
alimanfoo, line 692: Suggest to change to a `cohorts` parameter as above.

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,269 @@ def plot_h12_gwss(
...
+                return fig
+        except AssertionError as msg:
+            print(msg)
+            return None
+
+    @check_types
+    @doc(
+        summary="Plot h12 GWSS data with multiple tracks.",
+    )
+    def plot_h12_gwss_multi(
```
alimanfoo, line 689:
```suggestion
    def plot_h12_gwss_multi_panel(
```

Maybe make name more descriptive?

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,269 @@ def plot_h12_gwss(
...
+            bokeh.plotting.show(fig)
+            return None
+        else:
+            return fig
+
+    @check_types
+    @doc(
+        summary="Plot h12 GWSS data with multiple traces.",
+    )
+    def plot_h12_gwss_multitraces(
```
alimanfoo, line 618:
```suggestion
    def plot_h12_gwss_multi_overlay(
```

Maybe make name more descriptive?

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,269 @@ def plot_h12_gwss(
         else:
             return fig
 
+    @check_types
+    @doc(
+        summary="Plot h12 GWSS data track with multiple traces.",
+    )
+    def plot_h12_gwss_track_multi(
```
alimanfoo, line 521:
```suggestion
    def plot_h12_gwss_multi_overlay_track(
```

malariagen_data/anoph/h12_params.py
```diff
@@ -27,3 +27,38 @@
...
+        self.min_cohort_size = min_cohort_size
+        self.max_cohort_size = max_cohort_size
+
+
+sample_queries: TypeAlias = Annotated[
+    Sequence[sample_query_params],
+    """
+    A set of sample queries parameters. These include actual sample queries, the title associated to each one, the window size for the analysis, the site filter analysis that needs to be used, the cohort size, the minimum and the maximum cohort sizes.
+    """,
+]
```
alimanfoo, line 64: I think this could be simplified, because there is only one parameter here that might need to be varied between cohorts, which is the `window_size` parameter.

I think I would suggest to remove this, and just expose these parameters (`window_size`, `analysis`, `cohort_size`, `min_cohort_size`, `max_cohort_size`) directly on the function being called by the user. For `analysis`, `cohort_size`, `min_cohort_size`, `max_cohort_size` we would expect a single value and pass this through for all cohorts. For `window_size` it could be a single value, or it could be a dictionary mapping cohort keys to window size values.

Does that make sense?

So, e.g., a user would call:

```python
ag3.plot_h12_gwss_multi_panel(
    contig="2R",
    sample_sets=[...],
    cohorts="admin1_year",
    analysis="gamb_colu_arab",
    window_size=2000,
    min_cohort_size=20,
    ...
)
```

...or if different window sizes are required, a user would call:

```python
ag3.plot_h12_gwss_multi_panel(
    contig="2R",
    sample_sets=[...],
    cohorts="admin1_year",
    analysis="gamb_colu_arab",
    window_size={"...": 2000, "...": 1000, ...},
    min_cohort_size=20,
    ...
)

File: malariagen_data/anoph/h12.py (hunk 1 above)
Comment (alimanfoo) [reply malariagen_data/anoph/h12.py:524]: Also suggest to add a `sample_query` parameter, with behaviour as per `diversity_stats()`. I.e., the `sample_query` defines a base query, then the `cohorts` defines groups of samples on top of that. Should all be handled via the `_setup_cohort_queries()` helper.

**Review by alimanfoo — COMMENTED** [review]
Hi @jonbrenas, looking good, a few suggestions...

**Comment by jonbrenas** [comment]
Thanks @alimanfoo . I think I addressed all of your comments and I added some tests. I made the definition of `h12_params.window_size` a little more complex by including the option for a dict of window sizes which is not really ideal because other functions don't check that the value is an int. That probably needs to be done.

File: malariagen_data/anoph/h12.py [hunk 8]
```diff
@@ -514,6 +514,325 @@ def plot_h12_gwss(
...
+                        max_cohort_size=max_cohort_size,
+                        random_seed=random_seed,
+                        title=cohort_label,  # Same as previously
+                        sizing_mode=sizing_mode,
+                        width=width,
+                        height=track_height,
+                        show=False,
+                        output_backend=output_backend,
+                    )
+                )
```
alimanfoo, line 804:
```suggestion
            params = dict(
                contig=contig,
                analysis=analysis,
                window_size=window_size[cohort_label],
                sample_sets=sample_sets,
                sample_query=cohort_query,
                cohort_size=cohort_size,
                min_cohort_size=min_cohort_size,
                max_cohort_size=max_cohort_size,
                random_seed=random_seed,
                title=cohort_label,  # Deal with a choice of titles later
                sizing_mode=sizing_mode,
                width=width,
                height=track_height,
                show=False,
                output_backend=output_backend,
            )
            if i > 0:
                figs.append(
                    self.plot_h12_gwss_track(
                        x_range=figs[0].x_range,
                        **params
                    )
                )
            else:
                figs.append(
                    self.plot_h12_gwss_track(**params)
                )
```

Small suggestion to make code a bit more concise.

malariagen_data/anoph/h12_params.py
```diff
@@ -22,6 +22,13 @@
     """,
 ]
 
+multi_window_size: TypeAlias = Annotated[
+    Union[window_size, dict[str, int]],
+    """
+    The size of windows (number of SNPs) used to calculate statistics within.
```
alimanfoo, line 28:
```suggestion
    The size of windows (number of SNPs) used to calculate statistics within. Can
    be a single value, in which case the same window size will be used for all
    cohorts. Can also be a mapping from cohort identifiers to values, in case
    you need to provide different window sizes for different cohorts.
```

alimanfoo:
It works :)

image

alimanfoo:
Multi-panel also works... :)

image

Would be good to remove the X axis label and ticks from all H12 tracks.

alimanfoo:
Hi @jonbrenas, took the liberty to push a small change to the plot_h12_h1x notebook adding a couple of examples to exercise the new h12 functions.

jonbrenas:
Great! Thank you @alimanfoo.

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,305 @@ def plot_h12_gwss(
...
+            sample_query_options=sample_query_options,
+            cohort_size=cohort_size,
+            min_cohort_size=None,
+        )
+
+        if isinstance(window_size, int):
+            window_size = {k: window_size for k in cohort_queries.keys()}
+        elif isinstance(window_size, dict):
+            if np.sort(window_size.keys()) != np.sort(cohort_queries.keys()):
+                raise ValueError("Cohorts and window_sizes should have the same keys.")
```
alimanfoo, line 560: Just tried this out and it's raising an error, see below. Also in cases like this probably better to test for `Mapping` rather than `dict`, as we could accept anything which supports the `Mapping` interface.

Perhaps this...

```suggestion
        elif isinstance(window_size, Mapping):
            if set(window_size.keys()) != set(cohort_queries.keys()):
                raise ValueError("Cohorts and window_sizes should have the same keys.")
```

```
    558 elif isinstance(window_size, dict):
--> 559     if np.sort(window_size.keys()) != np.sort(cohort_queries.keys()):
    560         raise ValueError("Cohorts and window_sizes should have the same keys.")
    562 # Compute H12.

File ~/.cache/pypoetry/virtualenvs/malariagen-data-XzDLJC19-py3.10/lib/python3.10/site-packages/numpy/_core/fromnumeric.py:1077, in sort(a, axis, kind, order, stable)
   1075 else:
   1076     a = asanyarray(a).copy(order="K")
-> 1077 a.sort(axis=axis, kind=kind, order=order, stable=stable)
   1078 return a

AxisError: axis -1 is out of bounds for array of dimension 0
```

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,305 @@ def plot_h12_gwss(
...
+            output_backend=output_backend,
+        )
+
+        # Combine plots into a single figure.
+        fig = bokeh.layouts.gridplot(
+            [fig1, fig2],
+            ncols=1,
+            toolbar_location="above",
+            merge_tools=True,
+            sizing_mode=sizing_mode,
```
alimanfoo, line 709:
```suggestion
            sizing_mode=sizing_mode,
            toolbar_options=dict(active_inspect=None),
```

Apply solution from #633 to avoid warnings.

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,305 @@ def plot_h12_gwss(
...
+
+        figs.append(fig2)
+
+        # Combine plots into a single figure.
+        fig = bokeh.layouts.gridplot(
+            figs,
+            ncols=1,
+            toolbar_location="above",
+            merge_tools=True,
+            sizing_mode=sizing_mode,
```
alimanfoo, line 807:
```suggestion
            sizing_mode=sizing_mode,
            toolbar_options=dict(active_inspect=None),
```

Apply solution from #633 to avoid warnings.

tests/anoph/test_h12.py
```diff
@@ -148,6 +148,13 @@ def check_h12_gwss(*, api, h12_params):
     fig = api.plot_h12_gwss(**h12_params, show=False)
     assert isinstance(fig, bokeh.models.GridPlot)
 
+    h12_params.update({"cohorts": {"all": "year > 0"}})
+
+    fig = api.plot_h12_gwss_multi_overlay(**h12_params, show=False)
+    assert isinstance(fig, bokeh.models.GridPlot)
+    fig = api.plot_h12_gwss_multi_panel(**h12_params, show=False)
+    assert isinstance(fig, bokeh.models.GridPlot)
```
alimanfoo, line 156: Maybe good to test with more than one cohort. 

Also would be good to test both with a single window size (int) and with multiple window sizes (dict).

alimanfoo:
Hi @jonbrenas, just pushed a commit with an addition to the example notebook to exercise the functions using a dict value for the `cohorts` and `window_size` parameters.

malariagen_data/anoph/gplt_params.py
```diff
@@ -83,6 +83,14 @@
     "A bokeh figure (only returned if show=False).",
 ]
 
+def_figure: TypeAlias = Annotated[
```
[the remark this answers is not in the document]
jonbrenas: I think it is a good idea. It is a small change but that will affect a lot of files that have little to do with this PR so I am going to create a new issue (and the associated PR if I have time before I leave for Scotland).

jonbrenas:
Thanks @alimanfoo. There should now be tests with the dict version of `window_size` and the tests for the multi versions (which are now their own tests instead of being tacked on at the end of the mono version) now use 2 cohorts.

malariagen_data/anoph/h12.py
```diff
@@ -514,6 +514,307 @@ def plot_h12_gwss(
...
+                height=track_height,
+                show=False,
+                output_backend=output_backend,
+            )
+            if i > 0:
+                figs.append(self.plot_h12_gwss_track(x_range=figs[0].x_range, **params))
+            else:
+                figs.append(self.plot_h12_gwss_track(**params))
+
+            figs[-1].xaxis.visible = False
```
alimanfoo, line 787:
```suggestion
            if i > 0:
                track = self.plot_h12_gwss_track(x_range=figs[0].x_range, **params)
            else:
                track = self.plot_h12_gwss_track(**params)
            track.xaxis.visible = False
            figs.append(track)
```

Minor suggestion to make the code more obvious, particularly that all tracks should have xaxis hidden.

alimanfoo (commented):
All looks good, thanks @jonbrenas. Only one very small suggestion for code readability.

Just need to add the new functions to the Ag3 and Af1 API docs, then I think we're good to go :)

alimanfoo:
Hi @jonbrenas, as you're on holiday I took the liberty to apply a small code suggestion and add the new functions to the API docs. I'll go ahead and merge here if CI passes.

alimanfoo (approved):

alimanfoo:
Thanks @jonbrenas, great to see this go in, very nice feature! 🎉

sanjaynagi:
This looks great

PR #599, opened 2024-09-13
Handle missing CNV DRCs in similar way to missing CNV HMM

leehart:
Re: issue #551 

- Skip sets with missing CNV DRCs rather than FileNotFoundError.
- Raise ValueError when no given sets have DRCs.


alimanfoo (approved):
LGTM :)

Issue #601, opened 2024-09-16
Problems running large PCAs


alimanfoo:
I'm having problems running a PCA across the whole of the Ag3 resource. I'm running on a dask cluster with ~100 workers, all goes fine through the initial allele counts stage, but then the cluster grinds to a halt I think during the computation of the biallelic diplotypes.

I eventually got an error message about jobs being retried and failing multiple times, and have an associated stack trace for one of the workers where the job failed, here is an extract:

```
File "/opt/coiled/env/lib/python3.12/site-packages/allel/abc.py", line 40, in __getattr__
return getattr(self.values, item)
^^^^^^^^^^^
File "/opt/coiled/env/lib/python3.12/site-packages/allel/abc.py", line 26, in values
return self._values
^^^^^^^^^^^^
File "/opt/coiled/env/lib/python3.12/site-packages/allel/abc.py", line 40, in __getattr__
return getattr(self.values, item)
^^^^^^^^^^^
File "/opt/coiled/env/lib/python3.12/site-packages/allel/abc.py", line 26, in values
return self._values
^^^^^^^^^^^^
File "/opt/coiled/env/lib/python3.12/site-packages/allel/abc.py", line 40, in __getattr__
return getattr(self.values, item)
^^^^^^^^^^^
File "/opt/coiled/env/lib/python3.12/site-packages/allel/abc.py", line 26, in values
return self._values
^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded
distributed.worker - INFO - Stopping worker at tls://10.0.0.70:40125. Reason: check-worker-ttl-1726482783.8158462
distributed.worker - ERROR - failed during get data with tls://10.0.0.70:40125 -> tls://10.0.0.73:44005
```

Full trace: [gsu-vector-team-ca83e1d6-worker-30873129b3-instance.txt](https://github.com/user-attachments/files/17013444/gsu-vector-team-ca83e1d6-worker-30873129b3-instance.txt)

My guess is that there is a recursion error triggered when objects are being pickled and unpickled for transfer between dask workers. This is associated with scikit-allel array wrapper objects, which could be e.g. an instance of `AlleleCountsArray` or `GenotypeArray`. Confirming that this does occur, this code:

```python
import allel
import pickle

ac = allel.AlleleCountsArray(
    [[0, 10], [5, 5]]
)
pickle.loads(pickle.dumps(ac))
```

...gerates this trace:

```
---------------------------------------------------------------------------
[...]

PR #603, opened 2024-09-16
Upgrade scikit-allel to resolve issues running large PCAs on a distributed cluster

alimanfoo:
This PR upgrades the minimum version of scikit-allel and updates the poetry lock file.

Scikit-allel 1.3.12 includes a fix to pickling and unpickling certain data structures, which hopefully will address the problems encountered in #601 when running large PCAs. However, I won't mark #601 as resolved until that is confirmed.

See also https://github.com/cggh/scikit-allel/issues/424.


pyproject.toml
```diff
@@ -17,30 +17,22 @@ authors = [
...
 
 [tool.poetry.dependencies]
-# ensure compatibility with numba
 python = ">=3.10,<3.13"
-# ensure compatibility with modules compiled with numpy v1.x
 numpy = "*"
-# ensure compatibility with modules compiled with numpy v1.x
-numba = "*"
-# force llvmlite above broken version
+numba = ">=0.60.0"
```
alimanfoo, line 22: This seems to be needed otherwise poetry locks an ancient version of numba, I don't understand why.

alimanfoo:
cc @cclarkson, @leehart. If CI passes I'll merge and make a quick micro release, then test some large PCAs on datalab.

Issue #604, opened 2024-09-17
Build karyotyping into malariagen_data 


sanjaynagi:
By reading the targets from https://github.com/rrlove/compkaryo/tree/master/compkaryo we can very simply karyotype samples based on tagging SNPs. 

It would be handy to have a built-in method, so I don't have to keep sharing my notebooks with everyone :P

sanjaynagi:
This would only work for Ag3, not af1 - I think this mean it needs to be stored in ag3.py ?

PR #606, opened 2024-09-17
add support for karyotyping

sanjaynagi:
Addresses #604 

In this PR, I add support for karyotyping. I have added the functionality to ag3.py, as it is specific to gambiae complex - not for funestus. 

This works by reading in the tag SNPs directly from the compkaryo github (https://github.com/rrlove/compkaryo/tree/master/compkaryo/targets), subsetting to those positions, and calculating the mean genotype, returning a dataframe. Currently, the raw mean genotype is returned, but perhaps we should round it to 0,1,2 which might be less confusing for end users (the tag SNPs are not perfect). 

I use biallelic snp calls only, because the compkaryo tag snps do not specify specific alleles and assume biallelism, which is a problem, because what was biallelic in their data could be multiallelic in yours. Still, they do work fine from what I can see. 

Users may have to be advised that some tag SNPs are applicable only to certain species in the gambiae complex. For example, I have seen people try and report the frequency of tag snps of an inversion which isn't even segregating in their species. In theory we could prevent this within the function, but I'm not sure if that's necessary.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -351,9 +351,9 @@ class Ag3(AnophelesDataResource):
         return contig, df_tag_snps["pos"].to_numpy()
 
     def karyotype(
-        self, inversion: str, sample_sets=None, sample_query=None
+        self, inversion, sample_sets=None, sample_query=None
     ) -> pd.DataFrame:
-        contig, tag_snps_pos = self.load_inversion_tags(self, inversion)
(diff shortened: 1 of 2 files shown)


sanjaynagi:
FFS, I forgot about the fork thing and CI. is there no way around that? @ahernank @alimanfoo

sanjaynagi:
Closing this PR - will redo another from the main malariagen_data repo (not my fork), so that the CI will run properly

PR #608, opened 2024-09-17
Add Ag inversion karyotyping support 

### LINKED ISSUE #604: Build karyotyping into malariagen_data

sanjaynagi:
* Resolves #604.

In this PR, I add support for karyotyping. I have added the functionality to ag3.py, as it is specific to gambiae complex - not for funestus.

This works by reading in the tag SNPs, subsetting to those positions, and calculating the mean genotype, returning a dataframe. Currently, the raw mean genotype is returned, but perhaps we should round it to 0,1,2 which might be less confusing for end users (the tag SNPs are not perfect).

The tag SNPs were obtained from https://github.com/rrlove/compkaryo/tree/master/compkaryo/targets but have been augmented with information about the alleles. Tag SNP data are bundled into the malariagen_data package as they are relatively small.

Users may have to be advised that some tag SNPs are applicable only to certain species in the gambiae complex. For example, I have seen people try and report the frequency of tag snps of an inversion which isn't even segregating in their species. In theory we could prevent this within the function, but I'm not sure if that's necessary.


alimanfoo:
Cool, thanks @sanjaynagi!

Just wondering about the biallelic SNPs, as it could be that SNPs which were biallelic in the dataset that Becca used are no longer biallelic in a different sample set. It could also get worse the more samples are included, i.e., the more samples you include in the analysis, the more SNPs you lose as they are no longer biallelic. This could lead to a situation where the same sample could get different results depending on which other samples it was analysed together with. Do you think this might be an issue?

IIRC the original analysis was done with Ag1000G phase 2 data, so we might be able to find the specific alternate alleles if we had to.

alimanfoo:
Also... the `biallelic_snp_calls`() function will recode some alleles to 0. E.g., if a SNP only has the 1 and 2 alleles in the main callset, then these get recoded as 0 and 1 in the dataset returned by `biallelic_snp_calls()`. Again possibly not a major issue as could be uncommon, but thought I'd mention.

sanjaynagi:
> Cool, thanks @sanjaynagi!
> 
> Just wondering about the biallelic SNPs, as it could be that SNPs which were biallelic in the dataset that Becca used are no longer biallelic in a different sample set. It could also get worse the more samples are included, i.e., the more samples you include in the analysis, the more SNPs you lose as they are no longer biallelic. This could lead to a situation where the same sample could get different results depending on which other samples it was analysed together with. Do you think this might be an issue?
> 
> IIRC the original analysis was done with Ag1000G phase 2 data, so we might be able to find the specific alternate alleles if we had to.

Indeed, its not ideal. In my experience, the biallelism problem from compkaryo will introduce some error to the mean genotype estimates, but not enough to actually really confuse karyotyping- i.e, if we round the mean genotypes to 0,1 or 2, I still think samples will be karyotyped correctly. Its a bit anecdotal and I don't have data to show that. Anyway, for this reason I decided to ignore this issue for now. 

Although I agree it wouldn't be too difficult to find alleles - I would lean towards ignoring the issue for now, with a view to potentially finding specific tag SNP alleles soon (and perhaps doing funestus inversions at the same time?).

sanjaynagi:
> Also... the `biallelic_snp_calls`() function will recode some alleles to 0. E.g., if a SNP only has the 1 and 2 alleles in the main callset, then these get recoded as 0 and 1 in the dataset returned by `biallelic_snp_calls()`. Again possibly not a major issue as could be uncommon, but thought I'd mention.

ahhh, cool. Not really sure here what is the best approach (apart from doing it properly :laughing: )

sanjaynagi:
@alimanfoo here is a notebook of me doing that https://colab.research.google.com/drive/1u9OYOtLdQoU8SnJLBCZILC_AhtlOv8Mk?usp=sharing

I get its a little hacky - is this sufficient do you reckon?

and where would you like to store the tag snps + alleles ?

sanjaynagi:
I spoke with Alistair, we instead used phase 2 genotypes to very easily find the right ALT, here is the notebook -
 https://colab.research.google.com/drive/12gUgv7kzIkSDImLJOy84OujyPatZcOyO?usp=sharing 

Here is the csv file of tag snps + alts -  [karyotype_tag_snps.csv](https://github.com/user-attachments/files/17316811/karyotype_tag_snps.csv)
 
The question is, where do we store this file (and in what format?). Its 1080 rows (<20kb) in total. Perhaps we could just store it in the malariagen_data repo? otherwise GCS seeing as that is where we store data? @alimanfoo 

Once we store it somewhere, Ill finish the PR.

sanjaynagi:
Cool, ive added the functionality to check for exact alleles now. The only thing thats left to do I think, is put the tag snp data somewhere, and link to it properly.

malariagen_data/ag3.py
```diff
@@ -341,3 +347,100 @@ def _results_cache_add_analysis_params(self, params):
         super()._results_cache_add_analysis_params(params)
         # override parent class to add AIM analysis
         params["aim_analysis"] = self._aim_analysis
+
+    def load_inversion_tags(self, inversion: str):
+        # needs to be modified depending on where we are hosting
+        path = "malariagen_data/resources/karyotype_tag_snps.csv"
+        df_tag_snps = pd.read_csv(path, sep=",")
```
alimanfoo, line 354:
```suggestion
        import importlib.resources
        from . import resources
        with importlib.resources.path(resources, "karyotype_tag_snps.csv") as path:
            df_tag_snps = pd.read_csv(path, sep=",")
```

I think this is the Pythonic way to access a file that is included in the package.

Also you'll probably need to add a `MANIFEST.in` file at the root of the repository which has the line:

```
include malariagen_data/resources/*
```

...to make sure the files will get included when the package is packaged.

malariagen_data/ag3.py
```diff
@@ -341,3 +347,100 @@ def _results_cache_add_analysis_params(self, params):
...
+        return df_tag_snps.query(f"inversion == '{inversion}'").reset_index()
+
+    def karyotype_tags_n_alt(self, gt, alts, inversion_alts):
```
alimanfoo, line 357: This looks like it could be a function rather than a method, because `self` is never accessed. Suggest to move this to a standalone function. Also suggest to rename as `_karyotype_tags_n_alt` to indicate it's private, i.e., not intended for public API.

malariagen_data/ag3.py
```diff
@@ -341,3 +347,100 @@ def _results_cache_add_analysis_params(self, params):
...
+        sample_sets : list of str, optional
+            Sample sets to include in the analysis. If None (default), use all sample sets.
+        sample_query : str, optional
+            Query to filter samples. If None (default), use all samples.
+        """
+
+        inversion_list = ["2La", "2Rb", "2Rc_gam", "2Rc_col", "2Rd", "2Rj"]
+        assert (
+            inversion in inversion_list
+        ), f"Inversion {inversion} not found - please select one of {inversion_list}"
```
alimanfoo, line 400: You could define a type for this, e.g.:

```
inversion_param: TypeAlias = Literal["2La", "2Rb", "2Rc_gam", "2Rc_col", "2Rd", "2Rj"]
```

...and then declare this as the type of the `inversion` param, in which case the validation will be done automatically by the `check_types` wrapper.

malariagen_data/ag3.py
```diff
@@ -341,3 +347,100 @@ def _results_cache_add_analysis_params(self, params):
...
+        )
+        geno = allel.GenotypeDaskArray(ds_snps["call_genotype"].data)
+        pos = allel.SortedIndex(ds_snps["variant_position"].values)
+        samples = ds_snps["sample_id"].values
+        alts = ds_snps["variant_allele"].values.astype(str)
+
+        # subset to position of inversion tags
+        mask = pos.locate_intersection(inversion_pos)[0]
+        alts = alts[mask]
+        geno = geno.compress(mask, axis=0)
```
alimanfoo, line 423:
```suggestion
        geno = geno.compress(mask, axis=0).compute()
```

For simplicity, suggest to compute this into memory here.

malariagen_data/ag3.py
```diff
@@ -341,3 +347,100 @@ def _results_cache_add_analysis_params(self, params):
...
+
+        with self._spinner("Inferring karyotype from tag SNPs"):
+            gn_alt = self.karyotype_tags_n_alt(
+                gt=geno.values, alts=alts, inversion_alts=inversion_alts
+            )
+            is_called = geno.is_called()
+
+            # calculate mean genotype for each sample whilst masking missing calls
+            av_gts = np.mean(np.ma.MaskedArray(gn_alt, mask=~is_called), axis=0).data
```
alimanfoo, line 432: Not sure why `.data` is needed here?

malariagen_data/ag3.py
```diff
@@ -341,3 +347,100 @@ def _results_cache_add_analysis_params(self, params):
...
+            total_sites = np.sum(is_called, axis=0)
+
+            df = pd.DataFrame(
+                {
+                    "sample_id": samples,
+                    "inversion": inversion,
+                    "mean_tag_snp_genotype": av_gts.round().astype(
+                        int
+                    ),  # round the genotypes then convert to int
```
alimanfoo, line 441: Maybe also worth including the unrounded result, so people can check if results are clear or perhaps a bit more borderline?

malariagen_data/ag3.py
```diff
@@ -341,3 +347,100 @@ def _results_cache_add_analysis_params(self, params):
...
+
+            # calculate mean genotype for each sample whilst masking missing calls
+            av_gts = np.mean(np.ma.MaskedArray(gn_alt, mask=~is_called), axis=0).data
+            total_sites = np.sum(is_called, axis=0)
+
+            df = pd.DataFrame(
+                {
+                    "sample_id": samples,
+                    "inversion": inversion,
+                    "mean_tag_snp_genotype": av_gts.round().astype(
```
alimanfoo, line 439: Maybe this could just be called "karyotype"?

tests/test_ag3.py
```diff
@@ -851,3 +851,12 @@ def test_xpehh_gwss():
...
     assert_allclose(xpehh[:, 2][100], 0.4817561326426265)
+
+
+def test_karyotyping():
+    ag3 = setup_ag3(cohorts_analysis="20230516")
+
+    df = ag3.karyotype(inversion="2La", sample_sets="AG1000G-FR", sample_query=None)
+
+    assert isinstance(df, pd.DataFrame)
+    assert len(df) > 0
```
alimanfoo, line 862: Suggest to also check expected columns are present and expected dtypes of columns. Also expected range or set of values.

alimanfoo (commented):
Thanks @sanjaynagi, looking good, a few suggestions.

malariagen_data/ag3.py
sanjaynagi: Thanks, this is good to know. Currently AnoExpress reads the files it needs from the github repo urls - this is much better!!

malariagen_data/ag3.py
sanjaynagi: haha, never even noticed that. this is the last surviving line of code from compKaryo

malariagen_data/ag3.py
sanjaynagi: good shout

malariagen_data/ag3.py
sanjaynagi: Cool. Is it appropriate to put karyo_params.py in the anoph/ directory given that the karyo functions are in ag3? Or should I just define this inversion param in ag3.py 

also, I assume I was right not to put the code in anoph/ originally - i did that as it was ag3 only and not af1.

alimanfoo: Yeah this is Ag specific so I think it makes sense to put everything in the ag3.py module.

malariagen_data/ag3.py
```diff
@@ -341,3 +350,78 @@ def _results_cache_add_analysis_params(self, params):
...
+
+        Parameters
+        ----------
+        inversion : str
+            Name of inversion to infer karyotype for. One of ["2La", "2Rb", "2Rc_gam", "2Rc_col", "2Rd", "2Rj"].
+        sample_sets : list of str, optional
+            Sample sets to include in the analysis. If None (default), use all sample sets.
+        sample_query : str, optional
+            Query to filter samples. If None (default), use all samples.
+        """
```
alimanfoo, line 381: Would you mind migrating this to use the `@doc(...)` decorator? Should mean you only then need to provide the summary, and the parameter descriptions will get pulled in automatically.

malariagen_data/ag3.py
```diff
@@ -75,6 +81,9 @@ def _setup_aim_palettes():
 }
 
 
+inversion_param: TypeAlias = Literal["2La", "2Rb", "2Rc_gam", "2Rc_col", "2Rd", "2Rj"]
```
alimanfoo, line 84:
```suggestion
inversion_param: TypeAlias = Annotated[
    Literal["2La", "2Rb", "2Rc_gam", "2Rc_col", "2Rd", "2Rj"],
    "Name of inversion to infer karyotype for.",
]
```

malariagen_data/ag3.py
```diff
@@ -341,3 +350,78 @@ def _results_cache_add_analysis_params(self, params):
         super()._results_cache_add_analysis_params(params)
         # override parent class to add AIM analysis
         params["aim_analysis"] = self._aim_analysis
+
+    def load_inversion_tags(self, inversion: str):
```
alimanfoo, line 354: Can we give this on a docstring too, using the `@doc(...)` annotation? And `@check_types`.

malariagen_data/ag3.py
```diff
@@ -341,3 +350,78 @@ def _results_cache_add_analysis_params(self, params):
...
+
+            # calculate mean genotype for each sample whilst masking missing calls
+            av_gts = np.mean(np.ma.MaskedArray(gn_alt, mask=~is_called), axis=0)
+            total_sites = np.sum(is_called, axis=0)
+
+            df = pd.DataFrame(
+                {
+                    "sample_id": samples,
+                    "inversion": inversion,
+                    "mean_tag_snp_genotype": av_gts,
```
alimanfoo, line 420:
```suggestion
                    "karyotype_mean": av_gts,
```

Maybe a little more intuitive?

alimanfoo:
2La works well...

image

alimanfoo:
2Rb...

image

sanjaynagi:
Woohoooo (thanks for testing)

malariagen_data/ag3.py
```diff
@@ -341,3 +354,73 @@ def _results_cache_add_analysis_params(self, params):
...
+            av_gts = np.mean(np.ma.MaskedArray(gn_alt, mask=~is_called), axis=0)
+            total_sites = np.sum(is_called, axis=0)
+
+            df = pd.DataFrame(
+                {
+                    "sample_id": samples,
+                    "inversion": inversion,
+                    "karyotype_mean": av_gts,
+                    # round the genotypes then convert to int
+                    "karyotype": av_gts.round().astype(int),
```
alimanfoo, line 420:
```suggestion
                    f"karyotype_{inversion}_mean": av_gts,
                    # round the genotypes then convert to int
                    f"karyotype_{inversion}": av_gts.round().astype(int),
```

Just a thought, would make it easier if someone then wants to join together results for multiple inversions into a single dataframe, save them having to rename the "karyotype" columns in each.

alimanfoo:
2Rc_gam isn't segregating in the sets I tested.

2Rc_col looks OK though...

image

tests/test_ag3.py
```diff
@@ -877,3 +877,21 @@ def test_xpehh_gwss():
...
+    ag3 = setup_ag3(cohorts_analysis="20230516")
+
+    df = ag3.karyotype(inversion="2La", sample_sets="AG1000G-GH", sample_query=None)
+
+    assert isinstance(df, pd.DataFrame)
+    expected_cols = [
+        "sample_id",
+        "inversion",
+        "karyotype_mean",
+        "karyotype",
```
sanjaynagi, line 892:
```suggestion
        "karyotype_2La_mean",
        "karyotype_2La",
```

tests/test_ag3.py
```diff
@@ -877,3 +877,21 @@ def test_xpehh_gwss():
...
+    expected_cols = [
+        "sample_id",
+        "inversion",
+        "karyotype_2La_mean",
+        "karyotype_2La",
+        "total_tag_snps",
+    ]
+    assert set(df.columns) == set(expected_cols)
+    assert all(df["karyotype"].isin([0, 1, 2]))
+    assert all(df["karyotype_mean"].between(0, 2))
```
sanjaynagi, line 897:
```suggestion
    assert all(df["karyotype_2La"].isin([0, 1, 2]))
    assert all(df["karyotype_2La_mean"].between(0, 2))
```

alimanfoo:
Thanks @sanjaynagi. Are you cool if I push the notebook I made to test the karyotype calling? Might be nice to have it in the notebooks folder.

malariagen_data/ag3.py
```diff
@@ -341,3 +354,72 @@ def _results_cache_add_analysis_params(self, params):
...
+
+            df = pd.DataFrame(
+                {
+                    "sample_id": samples,
+                    "inversion": inversion,
+                    f"karyotype_{inversion}_mean": av_gts,
+                    # round the genotypes then convert to int
+                    f"karyotype_{inversion}": av_gts.round().astype(int),
+                    "total_tag_snps": total_sites,
+                }
```
alimanfoo, line 422:
```suggestion
                },
                dtype={f"karyotype_{inversion}": "category"},
```

I *think* this would work - by making this column explicitly categorical then downstream libraries like plotly will handle it as categorical. Otherwise plotly sees this as continuous and plots a colorbar.

sanjaynagi: FYI committed but not tested

malariagen_data/ag3.py
```diff
@@ -341,3 +354,73 @@ def _results_cache_add_analysis_params(self, params):
...
+                {
+                    "sample_id": samples,
+                    "inversion": inversion,
+                    f"karyotype_{inversion}_mean": av_gts,
+                    # round the genotypes then convert to int
+                    f"karyotype_{inversion}": av_gts.round().astype(int),
+                    "total_tag_snps": total_sites,
+                },
+                dtype={f"karyotype_{inversion}": "category"},
+            )
```
alimanfoo, line 424: Sorry my bad, this is erroring in my notebook.

Maybe we should do...

```suggestion
            kt_dtype = CategoricalDtype(categories=[0, 1, 2], ordered=True)
            df = pd.DataFrame(
                {
                    "sample_id": samples,
                    "inversion": inversion,
                    f"karyotype_{inversion}_mean": av_gts,
                    # round the genotypes then convert to int
                    f"karyotype_{inversion}": av_gts.round().astype(int),
                    "total_tag_snps": total_sites,
                },
            )
            df[f"karyotype_{inversion}"] = df[f"karyotype_{inversion}"].astype(kt_dtype)
```

alimanfoo: ...which would also need a:

```
from pandas import CategoricalDtype
```

...somewhere.

alimanfoo (approved):
I'm happy :)

alimanfoo:
All looks good to me. Happy to be merged if CI passes?

alimanfoo:
Awesome, done!

PR #616, opened 2024-09-20
Add parameters to the Anopheles pca function to allow better handling of outliers

### LINKED ISSUE #389: Run PCA as normal but exclude some samples from computing the projection
In some cases we find samples which probably have some cryptic relatedness, and these drive one or more principal components. It would be useful to be able to compute a PCA excluding some specific samples according to sample ID, but then calculate the projected coordinates including all samples.

A particular use case for this is taxon analysis, where you want to find the principal components which are important for taxon identification, but ultimately want to see where all samples lie in the projection.

alimanfoo:
This PR adds two new parameters to the Anopheles `pca()` function to help with situations where you have PCA outliers that need to be excluded:

* `exclude_samples` - This parameter can be a list of samples that will be excluded completely from the PCA analysis.
* `fit_exclude_samples` - This parameter can be a list of samples that will be excluded during the fitting stage of the PCA analysis, but will be included in the projection stage and therefore in the resulting output dataframe.

The `fit_exclude_samples` parameter is particularly useful where you have samples that are outliers but you still want to see where they fall within real geographical or taxonomic structure.

Note that both of these parameters are only applied after the loading of the input data (biallelic diplotypes) has been computed. This input data will be cached if a `results_cache` has been set, meaning that making changes to either of these parameters then rerunning the function should be relatively quick.

Resolves #389.

### DIFF
--- a/malariagen_data/anoph/pca.py
+++ b/malariagen_data/anoph/pca.py
@@ -72,13 +72,14 @@ class AnophelesPca(
         min_cohort_size: Optional[base_params.min_cohort_size] = None,
         max_cohort_size: Optional[base_params.max_cohort_size] = None,
         exclude_samples: Optional[base_params.samples] = None,
+        fit_exclude_samples: Optional[base_params.samples] = None,
         random_seed: base_params.random_seed = 42,
         inline_array: base_params.inline_array = base_params.inline_array_default,
         chunks: base_params.chunks = base_params.chunks_default,
     ) -> Tuple[pca_params.df_pca, pca_params.evr]:
         # Change this name if you ever change the behaviour of this function, to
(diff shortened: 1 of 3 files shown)


leehart:
Looks great but failed `test_pca_fit_exclude_samples[ag3_sim]`

alimanfoo:
> Looks great but failed `test_pca_fit_exclude_samples[ag3_sim]`

Thanks Lee. [That failure](https://github.com/malariagen/malariagen-data-python/actions/runs/10957716402/job/30426420884?pr=616) is tricky, it's one of those ones that only pops up sometimes, and I don't fully understand why. I've just pushed a commit which tries to work around, will see how it runs.

PR #617, opened 2024-09-20
Support fractions for max_missing_an and min_minor_ac parameters when accessing biallelic SNPs

### LINKED ISSUE #572: Allow fractional value for max_missing_an and min_minor_ac, e.g., 0.01
E.g. with regards to `pca()`

Re: PR https://github.com/malariagen/malariagen-data-python/pull/569

> This does also highlight a general weakness with setting `max_missing_an=0` as default for all datasets. In real datasets, as the number of samples gets larger, so the chance that you'll find variants with no missingness at all get smaller. 
> 
> Rather it would be better to set a fractional value, i.e., to allow something like `max_missing_an=0.01` which would mean keep all variants with at most 1% missing genotype calls.

alimanfoo:
* Resolves #572

### DIFF
--- a/malariagen_data/anoph/base_params.py
+++ b/malariagen_data/anoph/base_params.py
@@ -271,19 +271,21 @@ thin_offset: TypeAlias = Annotated[
 ]
 
 min_minor_ac: TypeAlias = Annotated[
-    int,
+    Union[int, float],
     """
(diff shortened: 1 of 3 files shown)


leehart:
Thanks @alimanfoo ! 👍

Issue #618, opened 2024-09-20
Random error when performing tests - empty range in randrange(3, 3)


jonbrenas:
In order to test `plot_pca_coords_3d()`, at least 3 components are needed when computing the PCA, so the number of components is chosen as `random.randint(3, min(n_samples, n_snps))`. Which is fine because there should be more than 3 samples and 3 snps but n_snps is actually set as `random.randint(1, n_snps_available)` which can be 3 or less.

jonbrenas:
The same bad random value is used elsewhere in the same file. I am going to update the rest of the code in the same way.

PR #619, opened 2024-09-20
Random error when performing tests

jonbrenas:
Addresses #618 .


leehart (approved):
Thanks @jonbrenas .

I'm not sure about the `assert`. I suppose the idea is that it will fail if either the number of samples or the number of SNPs is 3 or fewer. I expect there would be a `ValueError` raised in `randint` if `n_snps_available` was fewer than 4, so I expect `n_snps` will be at least 4 at the `assert` and so it might boil down to `n_samples`. Perhaps we might want a way to either ensure enough samples and SNPs at this point, or otherwise raise a clear error when that hasn't happened. But perhaps that's a different problem for a different day!

leehart:
By the way, I don't quite understand the commit that "Drops a few unused columns", which doesn't come through as a change, presumably due to having an identical change being already merged in from master? Was that commit related to the same issue, or an artefact of something else perhaps?

jonbrenas:
Thanks @leehart .

The `assert` should never be `False` unless the test is really poorly setup. `n_samples` is the sum of the sizes of the two randomly chosen sample sets and `n_snps_available` is the number of biallelic SNPs in the randomly chosen chromosome arm using all these samples. Both values should be significantly higher than 4.

The commits have been part of all my PRs for a while. I am not sure why. I hope they disappear once one of these PRs is merged into master.

leehart:
Thanks @jonbrenas 
CI tests failed on unrelated PCA tests. I'll try re-running.
> =========================== short test summary info ============================
> FAILED tests/anoph/test_pca.py::test_pca_exclude_samples[af1_sim] - ValueError: empty range in randrange(3, 3)
> =========== 1 failed, 379 passed, 322 warnings in 424.39s (0:07:04) ============

jonbrenas:
It actually fails on exactly the test that I modified because it can sometimes randomly fail on unrelated PRs (but this PR has to pass the previous tests for it to be approved, obviously, which it now does).

leehart:
Haha, I see, of course. Thanks @jonbrenas . OK if I delete the branch from GitHub now it's been merged in?

jonbrenas:
Done!

PR #623, opened 2024-09-23
Improve performance when accessing biallelic SNP calls

alimanfoo:
This PR improves performance of the Anopheles `biallelic_snp_calls()` function by around an order of magnitude.


malariagen_data/anoph/snp_data.py
```diff
@@ -1611,7 +1613,7 @@ def biallelic_snp_calls(
 
         with self._spinner("Prepare biallelic SNP calls"):
             # Subset to biallelic sites.
-            ds_bi = ds.isel(variants=loc_bi)
+            ds_bi = dask_compress_dataset(ds, indexer=loc_bi, dim="variants")
```
alimanfoo, line 1616: This is the main optimisation. Our implementation of `dask_compress_dataset()` is much faster than calling `ds.isel()`.

malariagen_data/anoph/snp_data.py
```diff
@@ -1673,20 +1683,21 @@ def biallelic_snp_calls(
...
                     if isinstance(min_minor_ac, float):
                         ac_minor_frac = ac_minor / an
                         loc_minor = ac_minor_frac >= min_minor_ac
                     else:
                         loc_minor = ac_minor >= min_minor_ac
                     loc_out &= loc_minor
 
-                ds_out = ds_out.isel(variants=loc_out)
+                # Apply selection from conditions.
+                ds_out = dask_compress_dataset(ds_out, indexer=loc_out, dim="variants")
```
alimanfoo, line 1695: This is the main optimisation. Our implementation of `dask_compress_dataset()` is much faster than calling `ds.isel()`.

Issue #208, opened 2022-04-06
Interactive map weirdness


alimanfoo:
Trying to use the `Ag3.plot_frequencies_interactive_map()` function I'm getting weirdness. For some reason I can't seem to get the `center` parameter to work as expected, when zoomed in a bit more. Suspect the addition and changes of markers are affecting this somehow. E.g., sometimes the map will jump to a new center when the selection of variant is changed.

leehart:
I think we need a reproducible example, if possible.
For what it's worth, I haven't been able to reproduce any weirdness on Colab yet, using v7.3.0 with just the `center` and `zoom` params, but importantly: I haven't tried adding and changing markers yet, which sounds like the trigger.

Some basic notes on the params for this issue:

Our docs:
```
center : tuple of int, optional
    Location to center the map.
zoom : int, optional
    Initial zoom level.
```
(American English spelling of "center".)


Our defaults:
```
center=(-2, 20),
zoom=3,
```

Our pass-through:
```
freq_map = ipyleaflet.Map(center=center, zoom=zoom)
```


The `ipyleaflet` docs from https://ipyleaflet.readthedocs.io/en/latest/map_and_basemaps/map.html#ipyleaflet.Map.center
```
center
    The current center of the map.
    Type
        list, default [0, 0]

zoom
    The current zoom value of the map.
    Type
        float, default 12
```

leehart:
I think I've found a potential source/explanation for the weirdness, although I haven't verified it yet.
The code for creating the interactive map widget does not maintain the state all of the function's parameters, including `center` and `zoom`.

I'll try changing this:
```
controls = ipywidgets.interactive(
    self.plot_frequencies_map_markers,
    m=ipywidgets.fixed(freq_map),
    ds=ipywidgets.fixed(ds),
    variant=ipywidgets.Dropdown(options=variants, description="Variant: "),
    taxon=ipywidgets.Dropdown(options=taxa, description="Taxon: "),
    period=ipywidgets.Dropdown(options=periods, description="Period: "),
    clear=ipywidgets.fixed(True),
)
```
To this:
```
controls = ipywidgets.interactive(
    self.plot_frequencies_map_markers,
    m=ipywidgets.fixed(freq_map),
    ds=ipywidgets.fixed(ds),
    center=ipywidgets.fixed(center),
    zoom=ipywidgets.fixed(zoom),
    title=ipywidgets.fixed(title),
    epilogue=ipywidgets.fixed(epilogue),
[...]

PR #624, opened 2024-09-23
Set auto_pan False for marker popup in plot_frequencies_map_markers()

leehart:
Re: issue #208


leehart:
E.g. allows
Screenshot 2024-09-23 at 12 57 52
instead of 
Screenshot 2024-09-23 at 12 56 07

alimanfoo (approved):
LGTM 🌴

PR #626, opened 2024-09-24
Further work on scalability for large biallelic genotype data computations

alimanfoo:
Following #623 I'm now finding new problems on a distributed cluster, specifically the scheduler being killed. This PR has further work towards achieving scalability on accessing and computing over biallelic SNP data.


alimanfoo:
Tested on a distributed cluster, seems to work better.

PR #630, opened 2024-09-27
Haplotype frequency analysis methods

### LINKED ISSUE #367: Haplotype frequency time series plot
For a given genome region, identify all the distinct haplotypes, and create a time series plot of haplotype frequencies, similar to time series plots of SNP allele frequencies.

Also possibly allow a stacked view as well as individual lines.

Motivation is to see which haplotypes at a locus under selection are common, how that changes over time, and compare between different cohorts.

jonbrenas:
This PR adds two new functions for computing haplotype frequencies within a genome region of interest, `haplotype_frequencies()` and `haplotype_frequencies_advanced()`. The outputs from these functions can then be passed into generic functions for visualising frequency data as a heatmap, time series or map.

Resolves #367.


alimanfoo:
Hi @jonbrenas, just to say this looks like it's moving in a good direction :)

In the `malariagen_data.anoph.h12` module there is a function `haplotype_frequencies()` which you could use to calculate haplotype frequencies here, if you moved `haplotype_frequencies()` to a common location like `malariagen_data.util` module.

jonbrenas:
Thanks @alimanfoo. I had used `haplotype_frequencies()` in my first prototype but, for the time series plot, counts and nobs are also needed. I considered moving the function and modifying slightly the functions using `haplotype_frequencies()` in `malariagen_data.anoph.h12` (which is probably only `garud_h12`) but I wanted to limit the amount of side modifications in this PR to a minimum. I guess `haplotype_frequencies()` is much faster thanks to numba so I'll transition it now.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,393 @@
...
+        hap_track: dict[np.int64, float] = {}
+        for coh, loc_coh in cohorts_iterator:
+            hap_track = {k: 0 for k in hap_track.keys()}
+            n_samples = np.count_nonzero(loc_coh)
+            assert n_samples >= min_cohort_size
+            gt_coh = allel.GenotypeDaskArray(da.compress(loc_coh, gt, axis=1))
+            gt_hap = gt_coh.to_haplotypes().compute()
+            f, _, _ = haplotype_frequencies(gt_hap)
+            hap_track.update(f)
+            freq_cols["frq_" + coh] = list(hap_track.values())
```
alimanfoo, line 107: I am a little concerned about ensuring that the frequency values are the same for each cohort, i.e., the values correspond to the same haplotypes. Also I think this could end up bleeding values between different cohorts.  

Am I right that you are trying to use `hap_track` to keep a record of what haplotypes you have seen so far?

Consider what would happen here if you have two cohorts, where some haplotypes are present in both cohorts, and each cohort has haplotypes which are not observed in the other cohort.

I wonder if it would be safer to do this in two passes. In the first pass, you compute the results of calling `haplotype_frequencies()` for each cohort. Then you find the set of all haplotype keys across all cohorts, and use this to generate a consistent array of frequencies for each cohort, guaranteeing the same order of values for each cohort, and filling with 0 in case the haplotype is not observed.

alimanfoo: If you did that you could also add a "label" column to the output dataframe, which could contain the hash for each haplotype. Not super user friendly, but would at least allow us to refer to the different haplotypes in some way?

jonbrenas: Not sure I get what you mean. I have little time so what I say might be gibberish. 

For the first part, `haplotype_frequencies()` returns a dict where the keys are the hash for each haplotype and the value is the frequency. These keys are then used by `hap_track` so it should already do what I think you asked with one pass.  
For the second part, "label" is already used for the index (because `plot_frequencies_heatmap()` requires it, I think). We could keep the old index (containing the hash) in a column "hash" if you want.

alimanfoo: > For the first part, `haplotype_frequencies()` returns a dict where the keys are the hash for each haplotype and the value is the frequency. These keys are then used by `hap_track` so it should already do what I think you asked with one pass.

Right, after you have iterated through all cohorts then `hap_track` will have seen all haplotypes present in any cohort, and so the keys will contain hashes of all haplotypes present.

But if you are outputting the frequencies as you go, within that first pass, then the number of values you find will vary between cohorts. This is because you find more haplotypes as you visit more cohorts.

Also, `hap_track` currently accumulates data from all cohorts. When you visit a new cohort, any haplotypes previously found will have their frequencies updated. But any haplotypes found in previous cohorts but not present in the current cohort will still be present within `hap_track`. If you then take all values from `hap_track` you could get a mangle of data from the current cohort and data from cohorts visited in previous iterations.

jonbrenas: Sorry @alimanfoo, I thought I had commented further yesterday but my comments seems to have got lost (I probably forgot to click on the big green button). 

> Right, after you have iterated through all cohorts then hap_track will have seen all haplotypes present in any cohort, and so the keys will contain hashes of all haplotypes present.
But if you are outputting the frequencies as you go, within that first pass, then the number of values you find will vary between cohorts. This is because you find more haplotypes as you visit more cohorts.

That's right. This is the reason why there is a step before creating the DataFrame that adds 0 to the end of the lists that are shorter than the total number of haplotypes (which is the length of the last list generated). This could be a problem if `update` doesn't treat the dict entries as a queue and reorders them (according to whatever order it likes). I tried this:
```
d = {"B": 0, "C":1}
d.update({"A": 2, "B":3}
d.values()
```
and it correctly returns `[3,1,2]` so the order seems to be indeed kept. So the new haplotypes are always added at the end of `hap_track`. I should probably document my code better (that is, at least a little) to make my reasoning clearer.

> Also, hap_track currently accumulates data from all cohorts. When you visit a new cohort, any haplotypes previously found will have their frequencies updated. But any haplotypes found in previous cohorts but not present in the current cohort will still be present within hap_track. If you then take all values from hap_track you could get a mangle of data from the current cohort and data from cohorts visited in previous iterations.

The first step of every cohort run is to reset all values of `hap_track` to 0 so I don't think there is a risk of mangling. I was scared at some point that the content of `hap_track` would be copied by reference instead of by value in `freq_coh` but given that they end up different, I don't think that's the case. 

There is one situation that I am not quite sure how to handle though. Resetting all values to 0 for each cohort run makes sense for `freq` and `count` but, if a haplotype is not observed, having 0 as the `nobs` value might be wrong. What do you think? Should the lengthening step set all values in `nobs` to be equal or is 0 for haplotypes that are not observed OK. As far as I can see, it only matters for the CIs ... and I just convinced myself that it is wrong to have `nobs == 0` so I will update that when I come back from my leave.

leehart:
@jonbrenas to add examples here and notebook examples.

jonbrenas:
Here are a few plots.

Ag3, heatmap:
Screenshot 2024-11-06 at 10 10 37
 Ag3, taxon/area facets:
Screenshot 2024-11-06 at 10 10 47
AF1, taxon/area facets:
Screenshot 2024-11-06 at 10 10 58

Should we add a parameter to the advanced version to query haplotypes (most likely to keep only ones that have a max_af > f)?

malariagen_data/anoph/hap_freq.py
alimanfoo, line 1: Nit, suggest rename file to `hap_frq.py` for naming consistency with `snp_frq.py` and `frq_params.py`.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
+from typing import Optional  # , Dict, Union, Callable, List
```
alimanfoo, line 1:
```suggestion
from typing import Optional
```

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+import numpy as np
+import xarray as xr
+import allel
+import dask.array as da
+from numpydoc_decorator import doc  # type: ignore
+
+from ..util import check_types, haplotype_frequencies
+from .hap_data import AnophelesHapData
+from .sample_metadata import locate_cohorts
+from . import base_params, frq_params  # , map_params, plotly_params
```
alimanfoo, line 13:
```suggestion
from . import base_params, frq_params
```

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+        self,
+        **kwargs,
+    ):
+        # N.B., this class is designed to work cooperatively, and
+        # so it's important that any remaining parameters are passed
+        # to the superclass constructor.
+        super().__init__(**kwargs)
+
+        # Set up cache variables.
+        self._cache_annotator = None
```
alimanfoo, line 29: This doesn't look like it's used at all? If not, please remove.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+        notes="""
+            Cohorts with fewer samples than `min_cohort_size` will be excluded from
+            output data frame.
+        """,
+    )
+    def haplotypes_frequencies(
+        self,
+        region: base_params.region,
+        cohorts: base_params.cohorts,
+        sample_query: Optional[base_params.sample_query] = None,
```
alimanfoo, line 48: Add `sample_query_options` parameter and pass through internally where appropriate.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+            coh: loc_coh
+            for coh, loc_coh in coh_dict.items()
+            if np.count_nonzero(loc_coh) >= min_cohort_size
+        }
+
+        # Early check for no cohorts.
+        if len(coh_dict) == 0:
+            raise ValueError(
+                "No cohorts available for the given sample selection parameters and minimum cohort size."
+            )
```
alimanfoo, line 73: Feels like all of this could reasonably be moved inside the `locate_cohorts()` function, in which case might also allow removal of code duplicated elsewhere.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+            coh_dict.items(), desc="Compute allele frequencies"
+        )
+        for coh, loc_coh in cohorts_iterator:
+            # We reset all frequencies to 0 for each cohort
+            hap_dict = {k: 0 for k in f_all.keys()}
+
+            n_samples = np.count_nonzero(loc_coh)
+            assert n_samples >= min_cohort_size
+            gt_coh = allel.GenotypeDaskArray(da.compress(loc_coh, gt, axis=1))
+            gt_hap = gt_coh.to_haplotypes().compute()
```
alimanfoo, line 109: At this point the genotypes are already computed, so this could be simplified. E.g.:

```suggestion
            gt_coh = gt.compress(loc_coh, axis=1)
            gt_hap = gt_coh.to_haplotypes()
```

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+            counts and frequency calculations.
+        """,
+    )
+    def haplotypes_frequencies_advanced(
+        self,
+        region: base_params.region,
+        area_by: frq_params.area_by,
+        period_by: frq_params.period_by,
+        sample_sets: Optional[base_params.sample_sets] = None,
+        sample_query: Optional[base_params.sample_query] = None,
```
alimanfoo, line 154: Add `sample_query_options` parameter and pass through internally.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+        df_cohorts = _build_cohorts_from_sample_grouping(
+            group_samples_by_cohort=group_samples_by_cohort,
+            min_cohort_size=min_cohort_size,
+        )
+
+        # Early check for no cohorts.
+        if len(df_cohorts) == 0:
+            raise ValueError(
+                "No cohorts available for the given sample selection parameters and minimum cohort size."
+            )
```
alimanfoo, line 185: Feels like this could be moved inside `_build_cohorts_from_sample_grouping()` to reduce code duplication.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+            # We reset all frequencies, counts to 0 for each cohort, nobs is set to the number of haplotypes
+            n_samples = cohort.size
+            hap_freq = {k: 0 for k in f_all.keys()}
+            hap_count = {k: 0 for k in f_all.keys()}
+            hap_nob = {k: 2 * n_samples for k in f_all.keys()}
+            assert n_samples >= min_cohort_size
+            sample_indices = group_samples_by_cohort.indices[cohort_key]
+            loc_coh = [i in sample_indices for i in range(0, gt.shape[1])]
+            gt_coh = allel.GenotypeDaskArray(da.compress(loc_coh, gt, axis=1))
+            gt_hap = gt_coh.to_haplotypes().compute()
```
alimanfoo, line 231: See similar comment above, data are already computed so can simplify.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+            cohort_key = cohort.taxon, cohort.area, cohort.period
+            cohort_key_str = cohort.taxon + "_" + cohort.area + "_" + str(cohort.period)
+            # We reset all frequencies, counts to 0 for each cohort, nobs is set to the number of haplotypes
+            n_samples = cohort.size
+            hap_freq = {k: 0 for k in f_all.keys()}
+            hap_count = {k: 0 for k in f_all.keys()}
+            hap_nob = {k: 2 * n_samples for k in f_all.keys()}
+            assert n_samples >= min_cohort_size
+            sample_indices = group_samples_by_cohort.indices[cohort_key]
+            loc_coh = [i in sample_indices for i in range(0, gt.shape[1])]
```
alimanfoo, line 229: Possibly simplification, if you have the sample indices you could use those directly via `take()`, rather than building a boolean mask and using `compress()`.

malariagen_data/anoph/hap_freq.py
```diff
@@ -0,0 +1,391 @@
...
+            df_haps_sorted.to_numpy()[:, 2 * len(df_cohorts) : -2],
+        )
+
+        # Add confidence intervals.
+        _add_frequency_ci(ds=ds_out, ci_method=ci_method)
+
+        return ds_out
+
+
+def _prep_samples_for_cohort_grouping(*, df_samples, area_by, period_by):
```
alimanfoo, line 288: This and all other functions below in the module are duplicates of functions already in the `snp_frq` module. Suggest to import from there rather than copying here.

malariagen_data/anopheles.py
```diff
@@ -102,7 +103,7 @@ class AnophelesDataResource(
     AnophelesPca,
     AnophelesIgv,
     AnophelesAimData,
-    AnophelesHapData,
+    AnophelesHapFrequencyAnalysis,
```
alimanfoo, line 106: Suggest to leave the `AnophelesHapData` class explicitly here within the parents, and add `AnophelesHapData` further back, e.g., together with other analysis classes.

alimanfoo (commented):
Thanks so much @jonbrenas, this is looking good. Some suggestions for code simplification and to remove duplication.

Also please add the new public methods `haplotype_frequencies()` and `haplotype_frequencies_advanced()` to the Ag3 and Af1 API docs.

alimanfoo:
Took the liberty to edit PR title and description so this PR looks good in the release notes.

malariagen_data/anopheles.py
jonbrenas: Thanks @alimanfoo. It would make sense but when `AnophelesHapData` is added to the list, this error pops up:

```
E   TypeError: Cannot create a consistent method resolution
E   order (MRO) for bases AnophelesSnpData, AnophelesHapData, AnophelesSampleMetadata, AnophelesCnvData, AnophelesGenomeFeaturesData, AnophelesGenomeSequenceData, AnophelesBase
```

I have tried to figure out why but I have not been able to yet.

malariagen_data/anoph/hap_freq.py
jonbrenas: Thank @alimanfoo. Given that these functions are not haplotype or SNP specific, wouldn't it make more sense to move them to `util` as well?

malariagen_data/anopheles.py
jonbrenas: Turns out `AnophelesHapData` really wants to be before `AnophelesSnpData` and `AnophelesCnvData`.

malariagen_data/anoph/hap_freq.py
alimanfoo: Yep, seems reasonable to me.

malariagen_data/anopheles.py
jonbrenas: Sort of. The only reason (as far as I can see) the order of these 3 would be important would be if the superclasses are used differently in `AnophelesHapData` and I don't see where they would be. I am not sure it is worth investigating too much given that it works this way, though.

alimanfoo (commented):
Hi @jonbrenas, this is looking good.

Small suggestion, the functions that have been refactored into the util module, suggest to remove the leading underscore from the function names. A leading underscore by convention indicates a function is intended to be private to a single module.

jonbrenas:
Thanks @alimanfoo, makes a lot of sense. I think I made the changes we requested.

alimanfoo (approved):
Looks good, thanks @jonbrenas 🌼

alimanfoo:
Just to note that this PR and #665 have some overlap and so expect there will be conflicts to resolve after one is merged before the second can be merged.

It might be slightly easier to merge this one first, and then deal with any conflicts in #665, but happy to go in whichever order you think best.

PR #632, opened 2024-09-30
Raise error if colab VM not in US

### LINKED ISSUE #620: Colab location check is not working
I just got a colab VM in Taiwan and there was no warning about being outside the US.

alimanfoo:
* Raises an error rather than a warning if the colab VM is not in the US
* Resolves #620

### DIFF
--- a/malariagen_data/util.py
+++ b/malariagen_data/util.py
@@ -1062,11 +1062,18 @@ def get_gcp_region(details):
                     return "us-west4"
                 elif region == "Texas":
                     return "us-south1"
+                else:
+                    return "us-other"
             elif country == "ZA":
                 if region == "Gauteng":
                     return "africa-south1"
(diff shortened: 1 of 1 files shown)


malariagen_data/util.py
```diff
@@ -1062,32 +1062,38 @@ def get_gcp_region(details):
                     return "us-west4"
                 elif region == "Texas":
                     return "us-south1"
+                else:
+                    return "us-other"
             elif country == "ZA":
                 if region == "Gauteng":
                     return "africa-south1"
             # Add other regions later if needed.
+            return "other"
```
alimanfoo, line 1071: Need to make sure we return a string here, otherwise the location check will be skipped. This was why the warnings were not being raised (#620).

leehart (approved):
Thanks @alimanfoo 👍

Issue #609, opened 2024-09-19
Various code warnings raised - umbrella issue


leehart:
This is an umbrella issue for code warnings being raised during current releases:

- [x] `UserWarning` "found multiple competing values for 'toolbar.active_inspect' property; using the latest value" - raised by `anoph/fst.py` - See issue https://github.com/malariagen/malariagen-data-python/issues/521
  - also raised by `anoph/g123.py`, `anoph/h12.py`, `anoph/h1x.py`
- [x] `RuntimeWarning` "invalid value encountered in divide" - raised by `allel/stats/fst.py` - See issue https://github.com/malariagen/malariagen-data-python/issues/598
- [x] `DeprecationWarning` "`in1d` is deprecated. Use `np.isin` instead." - raised by `anoph/g123.py` 
- ~[ ]  `DeprecationWarning` "  __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments." - raised by `dask/array/core.py`~ [Not seen recently.]

Kami-prog:
download
https://bit.ly/3XQLHlH
If you don't have the c compliator, install it.(gcc or clang)

leehart:
It looks like more work needs to be done to prevent the Bokeh toolbar warnings, e.g. https://github.com/malariagen/malariagen-data-python/actions/runs/10113341942/job/27969500641

I'll investigate these and see if they can be remedied in the same way:
```
anoph/cnv_data.py:747: UserWarning:
  
  found multiple competing values for 'toolbar.active_drag' property; using the latest value

anoph/cnv_data.py:747: UserWarning:
  
  found multiple competing values for 'toolbar.active_scroll' property; using the latest value

anoph/h1x.py:309: UserWarning:
  
  found multiple competing values for 'toolbar.active_inspect' property; using the latest value

anoph/h1x.py:309: UserWarning:
  
  found multiple competing values for 'toolbar.active_scroll' property; using the latest value

anoph/snp_data.py:1300: UserWarning:
  
  found multiple competing values for 'toolbar.active_drag' property; using the latest value

anoph/snp_data.py:1300: UserWarning:
  
  found multiple competing values for 'toolbar.active_scroll' property; using the latest value
```

Issue #521, opened 2024-05-09
UserWarnings when using bokeh.layouts.gridplot for GWSS


jonbrenas:
I have solved this locally (because I thought it was messing up my plots, turns out it wasn't) and I was going to update the code.

I remembered, however, that we have a lot of very similar functions doing GWSS plotting that need to be modified in the exact same way to remove these UserWarnings and I was wondering if it wouldn't be a good idea to factor in that code to reduce the errors that my crop up with duplicated code.

alimanfoo:
Thanks @jonbrenas - can you paste a snippet of code which shows how this can be resolved for a single function? Might help then to decide if/how we factor out some common code to do this for all functions.

jonbrenas:
Here is my code for windowed Fsts:

```
def moving_window_fst(contig, window_size, cohort_queries, sample_sets=None, sizing_mode = 'stretch_width', shape_size=5, y_range = (0,1), y_ticker = [0,1], min_cohort_size = 15, title = "Windowed Fst"):
  if len(cohort_queries.keys()) > 2:
    pair_list = list(combinations(cohort_queries.keys(), 2))
    lpl = len(pair_list)
  elif len(cohort_queries.keys()) == 2:
    pair_list = [(list(cohort_queries.keys())[0], list(cohort_queries.keys())[1])]
    lpl = 5
  else:
    print('At least 2 cohort queries are needed')
    return None
  colors= bokeh.palettes.Category10[lpl]
  fst_l = []
  for p in pair_list:
    x, fst = ag3.fst_gwss(
            contig=contig,
            window_size=window_size,
            cohort1_query=cohort_queries[p[0]],
            cohort2_query=cohort_queries[p[1]],
            min_cohort_size=min_cohort_size
    )
    fst_l.append(fst)
  #print(len(fst_l))
  x_min = x[0]
  x_max = x[-1]
  x_range = bokeh.models.Range1d(x_min, x_max, bounds="auto")
  xwheel_zoom = bokeh.models.WheelZoomTool(
      dimensions="width", maintain_focus=False
  )
  xpan_tool = bokeh.models.PanTool(
      dimensions="width"
  )

  fig2, hov_tool = plot_genes(
      region=contig,
      sizing_mode=sizing_mode,
      height = 90,
      x_range=x_range,
      active_scroll=xwheel_zoom,
      active_drag=xpan_tool,
      show=False,
  )

  fig1 = bokeh.plotting.figure(
      title=title,
[...]

PR #633, opened 2024-10-01
Add toolbar_options to grumbling gridplots

leehart:
Re: issues #609 and #521

### DIFF
--- a/malariagen_data/anoph/fst.py
+++ b/malariagen_data/anoph/fst.py
@@ -325,6 +325,7 @@ class AnophelesFstAnalysis(
             toolbar_location="above",
             merge_tools=True,
             sizing_mode=sizing_mode,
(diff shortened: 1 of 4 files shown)


leehart:
Hi @jonbrenas . Judging from local tests and the CI checks here, it appears to prevent the warnings as I mentioned in the issue #521.  What are the issues you're referring to with scroll and zoom? I don't expect this PR to address anything to do with that. Can we deal with that in a separate PR or is it inextricably linked to these particular warnings? Sounds like I've missed something!

jonbrenas (approved):
Thanks @leehart . I think I am the one who missed something. I think I remember a point where a lot of different `toolbar` properties were causing warnings but they seem to have been solved elsewhere. This looks perfect.

PR #634, opened 2024-10-01
Random error when performing tests

jonbrenas:
I expected it to be added at the end of the previous PR on the same branch but that will do.

There were a few more instances of the same bad `random.randint` calls that could create an error.


leehart (approved):
Thanks @jonbrenas

Issue #563, opened 2024-07-03
SNP browser to replace Panoptes


EricRLucas:
I tried to access the Panoptes web brower today, and found that it is being decommissioned. 

I gazed upon its ruin, and I wept. 

I gazed upon it some more, and I did wail, and beat my chest, and gnash my teeth. 

It would be good to have some way in malariagen_data to put in a bunch of snp position and get information on snp effect, allele frequencies, and a plot with the genomic context for the snps. Something along the lines of the colab below, but less crap. 

To the extent that there may be gems within with sewage, you are welcome to reuse that notebook. Also happy to help if needed. 

https://colab.research.google.com/drive/1e8AI2PjZgIWLJyciz4h6TbYF4B_PBe59?usp=drive_link

leehart:
Ref: https://github.com/cggh/panoptes

alimanfoo:
Thanks @EricRLucas, this is the best issue that's ever been raised on GitHub 😆

sanjaynagi:
actually needed something like this the other day - would you be interested in adding it? @EricRLucas

vigg-lstm:
@sanjaynagi  I'm happy to help, but certainly couldn't do it myself as I don't know enough about the workings of malariagen data. In the meantime, feel free to just make a copy of the colab I posted in the original issue. Should hopefully work out of the box

EricRLucas:
Oops, looks like I was logged in with the vigg-lstm account when writing the above

leehart:
@alimanfoo Maybe I should make a proof-of-concept PR based on @EricRLucas's Colab notebook, then we can all help to turn it into something beautiful, to all eyes. By the hammer of Hephaestus and the caduceus of Hermes...

leehart:
A local prototype is alive:
<img width="583" alt="Screenshot 2024-10-01 at 21 53 15" src="https://github.com/user-attachments/assets/065ee858-f324-4618-8380-d8db06311a46">

EricRLucas:
Very niice, thank you @leehart !  Is it possible for the function to also return the dataframe from the notebook? Or to have a separate function to do this (and perhaps even display it with the cells highlighted as in the notebook?)

leehart:
> have a separate function to do this (and perhaps even display it with the cells highlighted as in the notebook?)

@EricRLucas Yep, that's already in the works. I'll show that tomorrow. I'll also try to get the code into a good enough state for a PR soon.

leehart:
[...]

PR #636, opened 2024-10-10
Add SNP browser

leehart:
Re: issue #563

Hi @EricRLucas , @sanjaynagi , @alimanfoo

Here's a draft PR that incorporates the code from @EricRLucas's Colab notebook, just to provide a proof-of-concept and a straw-man prototype.

The general idea, to quote the issue is to give us:
> some way in malariagen_data to put in a bunch of snp position and get information on snp effect, allele frequencies, and a plot with the genomic context for the snps

However, it's clear that we still need to define the precise requirements and specify the behaviour in more detail. I will need some guidance on that, since I'm not very familiar with the specific use cases, etc.

As it stands, which is very much work in progress and open to comments and discussion, this implementation would provide a few new functions:
1) `snp_browser`, which takes a list of SNP positions and a contig name, and returns a Bokeh plot of the genes on the contig (using `plot_genes`) along with the given SNP positions coloured by their SNP effect, e.g. red for non-synonymous. 
2) `snp_freqs`, which also takes a list of SNP positions and a contig name, but returns a DataFrame containing the SNP frequencies for every allele at every position in the given contig, along with the SNP effect, impact, transcript, etc.
3) `snp_freqs_styled`, which returns a styled DataFrame (Styler object) that includes SNP frequencies coloured according to their value, and SNP effects coloured according to their type. 

These functions can be called in the following way, to effectively replicate the original Colab demo:
```python
import malariagen_data
ag3 = malariagen_data.Ag3()
list_of_snps = [28439294,28439830,28441004,28441038,28441062,28441261,28441637,28442886,28443076,28443204,28443615,28443671,28444974,28445027,28447012,28447524,28448227,28448524,28448698,28450330,28450649,28451169,28454362,28454790,28455838,28456044,28458165,28458776,28458790,28458809,28459999,28460127,28460712,28462387,28462494,28462954,28465913,28466872,28467008,28467686,28467860,28468704,28469065,28471707,28473673,28474453,28493647,28493982,28494698,28495340,28495917]
contig='2R'
```
```python
ag3.snp_browser(
  list_of_snps=list_of_snps,
  contig=contig
)
```
```python
import pandas as pd
pd.set_option('display.max_columns', None)
pd.set_option('display.max_rows', None)
ag3.snp_freqs_styled(
  list_of_snps=list_of_snps,
  contig=contig
)
```

There is the question of whether this functionally makes sense being embedded into the package, as opposed to providing a notebook that demonstrates how to achieve this using existing functionality. My personal reading is that there is evidently a call for us to integrate this functionality, in some form, if only for convenience, but there is also a desire to do it in the right way.

The naming of the functions and their arguments given here, even the placement of the code within the `AnophelesSnpFrequencyAnalysis` class (`anoph/snp_frq.py` module), is completely open for debate, just like everything else! There will be some technical restraints, of course, but generally the code will need to be harmonious with the rest of the package. 

I'm very conscious that this is basically just a copy-and-paste, so far, just to get the ball rolling, so I expect there is still lots of room for rewriting, refactoring, renaming and rejigging, to improve the code and make it fit in better with its surroundings. For example, the code will need to be adapted to also cater for other species groups, such as funestus (Af1).

Please feel free to make a PR on this PR, or submit your own new PRs, or anything else to help move this in the right direction, even if it's towards the bin! I'm also more than happy to update this PR accordingly, in response to any comments or suggestions.

### DIFF
--- a/malariagen_data/anoph/snp_frq.py
+++ b/malariagen_data/anoph/snp_frq.py
@@ -1277,7 +1277,7 @@ class AnophelesSnpFrequencyAnalysis(
         array_of_snps = np.array(list_of_snps)
 
         # Get the SNP frequencies as a DataFrame.
-        snp_freqs = self._any_snp_allele_freq(region, array_of_snps)
+        snp_freqs = self._any_snp_allele_freq(region, array_of_snps, sample_sets=sample_sets, sample_query=sample_query, site_mask=site_mask)
 
         # Return a copy of the DataFrame to insulate against changes.
         return snp_freqs.copy()
--- a/malariagen_data/anoph/snp_frq.py
+++ b/malariagen_data/anoph/snp_frq.py
@@ -1449,12 +1449,8 @@ class AnophelesSnpFrequencyAnalysis(
         if cohorts is None:
             # Create cohorts full all samples, then each species, then each country x species combo
             # FIXME: this needs to be made agnostic to species groups, e.g. funestus.
-            coh_dict1 = {
-                "all": np.full(num_samples, True),
-                "gambiae": df_samples.taxon == "gambiae",
-                "coluzzii": df_samples.taxon == "coluzzii",
-                "arabiensis": df_samples.taxon == "arabiensis",
-            }
+            coh_dict1 = {t : df_samples.taxon == t for t in np.unique(df_samples.taxon)}
+            coh_dict1['all'] = np.full(num_samples, True)
             coh_dict2 = {
                 c: df_samples.country == c for c in np.unique(df_samples.country)
             }
--- a/malariagen_data/anoph/snp_frq.py
+++ b/malariagen_data/anoph/snp_frq.py
@@ -3,18 +3,21 @@ import warnings
 from textwrap import dedent
 
 import allel  # type: ignore
-import numpy as np
-import pandas as pd
-from numpydoc_decorator import doc  # type: ignore
-import xarray as xr
+from functools import reduce
 import numba  # type: ignore
+import numpy as np
+from numpydoc_decorator import doc  # type: ignore
+import pandas as pd
+from pandas.io.formats.style import Styler
 import plotly.express as px  # type: ignore
+import re
+import xarray as xr
 
 from .. import veff
 from ..util import check_types, pandas_apply
 from .snp_data import AnophelesSnpData
 from .sample_metadata import locate_cohorts
-from . import base_params, frq_params, map_params, plotly_params
+from . import base_params, frq_params, gplt_params, map_params, plotly_params
 
 
 AA_CHANGE_QUERY = (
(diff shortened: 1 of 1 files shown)


malariagen_data/anoph/snp_frq.py
```diff
@@ -1252,6 +1255,398 @@ def snp_genotype_allele_counts(
...
+        # Build cohort dictionary, maps cohort labels to boolean indexers.
+        if cohorts is None:
+            # Create cohorts full all samples, then each species, then each country x species combo
+            # FIXME: this needs to be made agnostic to species groups, e.g. funestus.
+            coh_dict1 = {
+                "all": np.full(num_samples, True),
+                "gambiae": df_samples.taxon == "gambiae",
+                "coluzzii": df_samples.taxon == "coluzzii",
+                "arabiensis": df_samples.taxon == "arabiensis",
+            }
```
sanjaynagi, line 1457:
```suggestion
            coh_dict1 = {t : df_samples.taxon == t for t in np.unique(df_samples.taxon)}
            coh_dict1['all'] = np.full(num_samples, True)
```

malariagen_data/anoph/snp_frq.py
```diff
@@ -1252,6 +1255,398 @@ def snp_genotype_allele_counts(
...
+        """,
+        parameters=dict(
+            list_of_snps="List of SNP positions.",
+        ),
+    )
+    def snp_freqs(
+        self,
+        *,
+        list_of_snps: List[int],
+        contig: base_params.contig,
```
sanjaynagi, line 1271:
```suggestion
    def snp_freqs(
        self,
        *,
        list_of_snps: List[int],
        contig: base_params.contig,
        sample_sets=None,
        sample_query=None,
        site_mask=None
```

malariagen_data/anoph/snp_frq.py
```diff
@@ -1252,6 +1255,398 @@ def snp_genotype_allele_counts(
...
+    ) -> pd.DataFrame:
+        # Compose the region using the min and max SNP positions.
+        region = f"{contig}:{min(list_of_snps)}-{max(list_of_snps)}"
+
+        # Convert the list of SNP positions into a Numpy array.
+        array_of_snps = np.array(list_of_snps)
+
+        # Get the SNP frequencies as a DataFrame.
+        snp_freqs = self._any_snp_allele_freq(region, array_of_snps)
```
sanjaynagi, line 1280:
```suggestion
        snp_freqs = self._any_snp_allele_freq(region, array_of_snps, sample_sets=sample_sets, sample_query=sample_query, site_mask=site_mask)
```

malariagen_data/anoph/snp_frq.py
sanjaynagi: just exposing those parameters... wanted to see what speed is like for smaller cohorts

malariagen_data/anoph/snp_frq.py
sanjaynagi: this can easily be generalised using the metadata

sanjaynagi:
Thanks Lee! thats great. FWIW I definitely think functionality like this would be useful in malariagen_data. 

Some small suggestions above, and some thoughts below, though it'll be good to hear what other people think. 

1. The use of both contig and a list of SNPs is a little unusual for malariagen_data, and would make it difficult to look at snps on different contigs in one go. Ideally the function will be able to handle that, and will therefore loop through each contig, creating and concatenating the dataframes. I suggest the input should be a list/array of strings, like ['2RL:4564435', 'X:15000345'].
2. I would be inclined to keep these functions in a separate script (snp_explorer.py?) but lets see what @alimanfoo thinks 
3. Not sure whether there is a need for the styled version? unless malariagen_data produces other styled dataframes... i dont recall that 
4. I agree with you that the names of the functions should be clearer. Perhaps ag3.snp_browser() could be `ag3.plot_snp_location()` . The snp_freq() function needs to be clearly distinct from the current function ag3.snp_allele_frequencies() which takes a transcript as input. perhaps ag3.any_snp_allele_frequencies()? Open to ideas


**Actually(!!!),** considering that this function is basically ag3.snp_allele_frequencies() that can handle snp positions as input rather than transcripts, I wonder if we should just try and modify snp_allele_frequencies() instead to allow transcript or snp position ('2R:3454525') input. Otherwise, we will be duplicating a lot of code.

malariagen_data/anoph/snp_frq.py
vigg-lstm: Yes. Though I think it would still be nice to include the "all" cohort.

vigg-lstm:
Thanks Lee, and thank you Sanjay for your thoughts. Here are my musings, taking each of Sanjay's points in turn:

> 1. The use of both contig and a list of SNPs is a little unusual for malariagen_data, and would make it difficult to look at snps on different contigs in one go. Ideally the function will be able to handle that, and will therefore loop through each contig, creating and concatenating the dataframes. I suggest the input should be a list/array of strings, like ['2RL:4564435', 'X:15000345'].

That's fine for the function that returns a dataframe. For the one that returns a chromosomal map, it becomes problematic if more than one chromosome is supplied. Unless you can make it output one map for every relevant contig? In that case, you'd need to add a bit that groups the supplied SNPs by contig, and then draws the map separately for each one. 

> 2. I would be inclined to keep these functions in a separate script (snp_explorer.py?) but lets see what @alimanfoo thinks

No opinion. 

> 3. Not sure whether there is a need for the styled version? unless malariagen_data produces other styled dataframes... i dont recall that

Won't die on this hill if it's out of keeping with other malariagen_data function. But I think the colour is useful. It makes it a lot easier to get a quick visual impression of SNP frequency. Maybe it could be an optional argument (doesn't even need to be a default option).

> 4. I agree with you that the names of the functions should be clearer. Perhaps ag3.snp_browser() could be ag3.plot_snp_location() . The snp_freq() function needs to be clearly distinct from the current function ag3.snp_allele_frequencies() which takes a transcript as input. perhaps ag3.any_snp_allele_frequencies()? Open to ideas

No opinion

> Actually(!!!), considering that this function is basically ag3.snp_allele_frequencies() that can handle snp positions as input rather than transcripts, I wonder if we should just try and modify snp_allele_frequencies() instead to allow transcript or snp position ('2R:3454525') input. Otherwise, we will be duplicating a lot of code.

Fine with me

Only one suggestion, though I realise that this would require changing ag3.plot_genes() and so maybe requires its own issue: it would be good for the gene plot to show introns and exons. I know that we already colour the SNPs according to whether they are coding or non-coding, but it would still be useful to get a quick oversight of where the introns and exons are relative to the SNPs. 

Thanks again for doing this.

malariagen_data/anoph/snp_frq.py
sanjaynagi: > Yes. Though I think it would still be nice to include the "all" cohort.

its there :)

alimanfoo:
> Only one suggestion, though I realise that this would require changing ag3.plot_genes() and so maybe requires its own issue: it would be good for the gene plot to show introns and exons. I know that we already colour the SNPs according to whether they are coding or non-coding, but it would still be useful to get a quick oversight of where the introns and exons are relative to the SNPs.

There is a `plot_transcript()` function already which plots introns and exons for a single transcript.

Harder to do for multiple genes as each gene can have multiple transcripts.

alimanfoo:
What about having a function called something like `plot_snp_allele_frequencies()` which takes a specific `transcript` as a parameter, and calls `snp_allele_frequencies()` internally to generate the data on effects and frequencies to visualise. It then plots the transcript as a track...

image

...and above that plots another track with a circle marker for each alternate allele where the X coordinate is the position, the Y coordinate is the allele frequency, and the colour is related to the effect? Hover tooltips on each marker then give more information about the specific effect (e.g., what amino acid substitution, etc.).

I.e., visualise allele frequencies and effects in a single plot?

EricRLucas:
> Btw a challenge for creating a plot or output that shows SNPs with effects across an arbitrary genome region is that sometimes genes overlap. I.e., at a given position in the reference genome, there may be more than one overlapping gene. In this case the effect of a SNP at such a position will depend on which gene you want to consider.
> 
> Also genes have multiple transcripts and effects will depend on which transcript you choose.

The colab notebook dealt with this in the dataframe by making the SNP effect take the form of a semi-colon-delimited list of effects, one for every transcript (the transcript column is also a semi-colon-delimited list of the same length, and so by comparing the two, you can see which effects apply to which transcript). For the plot, it used the "strongest" effect to determine colour (only three effect types were considered: non-synonymous, synonymous or non-coding). So if a SNP was non-synonymous in only one transcript, it would be classed as non-synonymous.

EricRLucas:
> What about having a function called something like `plot_snp_allele_frequencies()` which takes a specific `transcript` as a parameter, and calls `snp_allele_frequencies()` internally to generate the data on effects and frequencies to visualise. It then plots the transcript as a track...
> 
> image
> 
> ...and above that plots another track with a circle marker for each alternate allele where the X coordinate is the position, the Y coordinate is the allele frequency, and the colour is related to the effect? Hover tooltips on each marker then give more information about the specific effect (e.g., what amino acid substitution, etc.).
> 
> I.e., visualise allele frequencies and effects in a single plot?

That would be a useful function, but would only work for transcripts, and you would need to know what transcript your SNPs are in. My use case is that I will have a bunch of SNPs that I know nothing about, I want to plug them into a function to know things like what's around them in the genome, what genes (if any) are nearby, what frequencies are those SNPs at and, if inside a gene, what are the SNP effects.

alimanfoo:
> The colab notebook dealt with this in the dataframe by making the SNP effect take the form of a semi-colon-delimited list of effects, one for every transcript (the transcript column is also a semi-colon-delimited list of the same length, and so by comparing the two, you can see which effects apply to which transcript). For the plot, it used the "strongest" effect to determine colour (only three effect types were considered: non-synonymous, synonymous or non-coding). So if a SNP was non-synonymous in only one transcript, it would be classed as non-synonymous.

That's good to know.

So is this a situation where you have a list of SNPs that you suspect are associated with insecticide resistance, and you're hunting for potentially causal variants (hence wanting to know what the strongest effect is)?

alimanfoo:
> How did Panoptes get around this? Did it use canonical transcripts for each gene?

Yeah it was driven from SNPEFF annotations, and IIRC I generated those in a way that only output annotations for a canonical transcript for each gene.

alimanfoo:
SNP effects are hard 😅

EricRLucas:
> > That would be a useful function, but would only work for transcripts, and you would need to know what transcript your SNPs are in. My use case is that I will have a bunch of SNPs that I know nothing about, I want to plug them into a function to know things like what's around them in the genome, what genes (if any) are nearby, what frequencies are those SNPs at and, if inside a gene, what are the SNP effects. But maybe I'm an exception in needing something like this. Happy to use this plot transcript function and just use my notebook for my particular use case.
> 
> It's good to understand your use case :)
> 
> Where is this list of SNPs coming from? A GWAS?

Yes, a GWAS is the most likely scenario. Or for example recently I was looking for SNPs with different frequencies in gambiae and coluzzii and wanted to know their local frequencies in a few different countries, then I wanted to know how close those SNPs were to introns in case I wanted to design cDNA primers for them.

EricRLucas:
> > The colab notebook dealt with this in the dataframe by making the SNP effect take the form of a semi-colon-delimited list of effects, one for every transcript (the transcript column is also a semi-colon-delimited list of the same length, and so by comparing the two, you can see which effects apply to which transcript). For the plot, it used the "strongest" effect to determine colour (only three effect types were considered: non-synonymous, synonymous or non-coding). So if a SNP was non-synonymous in only one transcript, it would be classed as non-synonymous.
> 
> That's good to know.
> 
> So is this a situation where you have a list of SNPs that you suspect are associated with insecticide resistance, and you're hunting for potentially causal variants (hence wanting to know what the strongest effect is)?

Yes

alimanfoo:
Thanks @EricRLucas, very useful to know.

alimanfoo:
Just to add another consideration here, this package calculates SNP effects on demand, and has it's own code for doing so. It's fast enough for one gene but have never tested it with lots of SNPs and genes at the same time.

alimanfoo:
I wonder if a good first step here would be to add a `plot_transcripts()` function, which has an API similar to `plot_genes()` but plots the full model (i.e., exons, introns and UTRs) for each gene within the region given. I.e., same visualisation as the current `plot_transcript()` function but for all transcripts in any genome region, rather than a single transcript, similar to most standard genome browsers, e.g.:

image
 
Possibly could also have an option `canonical=True` to just show a canonical transcript for each gene.

EricRLucas:
Sounds good. IMO, would be good for canonical = True to be the default value

alimanfoo:
Trying to also think about how we could generalise `snp_allele_frequencies()` to handle an arbitrary genome region rather than a single transcript. Should be possible, but need to give it a bit more thought.

If we could get there, then might be doable to also support passing in a specific set of positions rather than a region.

alimanfoo:
> Trying to also think about how we could generalise `snp_allele_frequencies()` to handle an arbitrary genome region rather than a single transcript. Should be possible, but need to give it a bit more thought.

Almost feels like something that should be precomputed. Computing effects on the fly for larger regions is gonna be a bit slow.

PR #637, opened 2024-10-10
Replace np.in1d with np.isin in g123.py

leehart:
Re: issue #609

Avoids `DeprecationWarning` raised by `anoph/g123.py`: "`in1d` is deprecated. Use `np.isin` instead."


leehart:
Will try to resolve CI coverage test failures.

leehart:
Thanks @jonbrenas

Issue #597, opened 2024-09-10
Disable progress via an environment variable


alimanfoo:
Sometimes we would like to run without showing progress outputs, e.g., when running a notebook that we then want to render to a Jupyter book for a training course. This can be done via code, but then we don't want the notebook to contain this code, because if a user launches the notebook in colab, then they will also have progress turned off. 

Propose to support hiding progress via an environment variable.

PR #640, opened 2024-10-15
Use optional env var for show_progress in Ag3, Af1

leehart:
Re: issue #597 

I'm still wondering if we can put this logic in the common base class `AnophelesDataResource` instead.


malariagen_data/af1.py
```diff
@@ -91,6 +92,18 @@ def __init__(
...
+            # Get the env var, if it exists.
+            show_progress_env = os.getenv("MGEN_SHOW_PROGRESS")
+
+            # If the env var does not exist, then use the class default.
+            # Otherwise, convert the env var value to a boolean and use that.
+            if show_progress_env is None:
+                show_progress = True
+            else:
+                show_progress = show_progress_env.lower() in ("true", "1", "yes", "on")
+
```
alimanfoo, line 106: The logic looks good but could move this up to `AnophelesBase.__init__`.

leehart: Confusingly, `AnophelesBase.__init__` currently defaults `show_progress` to `False`, whereas `Ag3` and `Af1` currently default to `True`. I expect this is just an historical artefact, so I'll set the default as `True`.

malariagen_data/ag3.py
```diff
@@ -150,6 +150,8 @@ def __init__(
         tqdm_class=None,
         **storage_options,  # used by fsspec via init_filesystem()
     ):
+        print("Ag3 show_progress", show_progress)
+
```
alimanfoo, line 154:
```suggestion
```

alimanfoo (commented):
Looks good, just a print statement to remove.

alimanfoo (approved):
All good, thanks!

PR #641, opened 2024-10-15
Improve neighbour-joining tree performance

### LINKED ISSUE #627: Neighbour-joining tree performance
Running the `plot_njt()` function on larger datasets can take upwards of 20 minutes or more specifically in the stage after the pairwise distances have been computed and during the construction of the neighbour-joining tree..

Perhaps there is an opportunity to accelerate or parallelise the neighbour-joining tree construction. Currently we use the implementation from biotite, which was faster than either BioPython or scikit-bio, but perhaps there is room for further improvement.

We could also at least cache the results of the tree computation so we don't need to rerun it.

alimanfoo:
* Use [anjl](https://alimanfoo.github.io/anjl/) instead of biotite to build neighbour-joining trees, which is capable of building trees for 20,000+ samples in around 5 minutes. Resolves #627.
* Cache the results of tree construction to avoid repeated computation. 
* Use anjl for tree plotting.
* Also work towards #366. 
* Along the way, fix issues encountered when plotting and data includes missing values.

---

* [x] Create a new module malariagen_data.anoph.distance.
* [x] Add a class AnophelesDistanceAnalysis.
* [x] Move the public method biallelic_pairwise_distances() and associated private methods from the AnophelesDataResource class to the new AnophelesDistanceAnalysis class.
* [x] Add a public method njt() to the new class which returns an array Z and which also saves results to the results cache.
* [x] Add a private method _njt() which is called from within njt() to run the actual neighbour-joining if the result is not cached, calling biallelic_pairwise_distances() internally to obtain a distance matrix then anjl.rapid_nj() to compute a tree.
* [x] Add a public method plot_njt() to the new class which calls njt() first to obtain a tree then calls anjl.plot() to create a plot.
* [x] Remove the plot_njt() method from the AnophelesDataResource class.
* [x] Add AnophelesDistanceAnalysis as a parent class of AnophelesDataResource.
* [x] Add tests to get coverage of the new AnophelesDistanceAnalysis class.


leehart (approved):
Awesome. Thanks @alimanfoo 👍

Issue #635, opened 2024-10-07
Add `sample_query_options` arg to functions without `sample_query` arg, e.g. h1x_gwss


leehart:
How best to handle cases like this?
E.g. from `anoph/h1x.py`
```python
    def h1x_gwss(
        self,
        contig: base_params.contig,
        window_size: h12_params.window_size,
        cohort1_query: base_params.sample_query,
        cohort2_query: base_params.sample_query,
        analysis: hap_params.analysis = base_params.DEFAULT,
        sample_sets: Optional[base_params.sample_sets] = None,
        cohort_size: Optional[base_params.cohort_size] = h12_params.cohort_size_default,
        min_cohort_size: Optional[
            base_params.min_cohort_size
        ] = h12_params.min_cohort_size_default,
        max_cohort_size: Optional[
            base_params.max_cohort_size
        ] = h12_params.max_cohort_size_default,
        random_seed: base_params.random_seed = 42,
        chunks: base_params.chunks = base_params.native_chunks,
        inline_array: base_params.inline_array = base_params.inline_array_default,
    ) -> Tuple[np.ndarray, np.ndarray]:
```

```python
    def _h1x_gwss(
        self,
        contig,
        analysis,
        window_size,
        sample_sets,
        cohort1_query,
        cohort2_query,
        cohort_size,
        min_cohort_size,
        max_cohort_size,
        random_seed,
        chunks,
        inline_array,
    ):
        # Access haplotype datasets for each cohort.
        ds1 = self.haplotypes(
            region=contig,
            analysis=analysis,
            sample_query=cohort1_query,
            sample_sets=sample_sets,
            cohort_size=cohort_size,
            min_cohort_size=min_cohort_size,
            max_cohort_size=max_cohort_size,
            random_seed=random_seed,
            chunks=chunks,
            inline_array=inline_array,
        )
        ds2 = self.haplotypes(
            region=contig,
            analysis=analysis,
            sample_query=cohort2_query,
            sample_sets=sample_sets,
            cohort_size=cohort_size,
            min_cohort_size=min_cohort_size,
            max_cohort_size=max_cohort_size,
            random_seed=random_seed,
            chunks=chunks,
            inline_array=inline_array,
        )
```

E.g. we could add support for
```python
        cohort1_query_options: base_params.sample_query_options,
        cohort2_query_options: base_params.sample_query_options,
```
Or just one shared set of options, e.g.
```python
        sample_query_options: base_params.sample_query_options,
```

[...]

PR #645, opened 2024-10-21
Add sample_query_options to more functions

leehart:
Re: issue #635


leehart:
I'm starting to suspect changing these functions to kwargs-only was unnecessary, and it risks being incompatible with existing calls, although it does seem in keeping with the approach/style elsewhere in the package, and I suppose function calls can be updated... 🤔

alimanfoo:
> I'm starting to suspect changing these functions to kwargs-only was unnecessary, and it risks being incompatible with existing calls, although it does seem in keeping with the approach/style elsewhere in the package, and I suppose function calls can be updated... 🤔

The approach I've taken so far is that private functions should be kwargs-only wherever possible, to help avoid bugs introduced by changing the order of parameters in a function call. But public functions can accept positional args.

If we changed public functions to kwargs-only we'd likely break some user code. I was tempted to do it a while back, but then thought better of it when I realised I do tend to use positional args when doing data analysis.

leehart:
Thanks @alimanfoo . I'll amend this PR accordingly.

alimanfoo (approved):
LGTM, thanks @leehart!

PR #647, opened 2024-10-22
Add support for dynamic neighbour-joining

### LINKED ISSUE #646: Dynamic neighbour-joining
A substantially faster and more scalable neighbour-joining algorithm is available in anjl 1.2.0. We can upgrade the dependency and add support for the dynamic algorithm, setting it as the default.

alimanfoo:
A new version of [anjl](https://alimanfoo.github.io/anjl/) is available which provides an implementation of the dynamic neighbour-joining algorithm. The dynamic algorithm is much faster and more scalable than other algorithms.

image

This PR adds support for the dynamic algorithm and makes it the default.

* Resolves #646


alimanfoo:
The codecov/project check is failing because of changes introduced in a previous PR (#597). I guess codecov has got itself a bit behind somehow. Suggest we ignore and merge here regardless.

PR #649, opened 2024-10-31
sort gff in `genome_features()`

### LINKED ISSUE #558: Advanced diplotype clustering orders genes in the CNV section by label rather than by genomic position
If the `cnv_region` contains multiple genes, the CNV heatmap rows will be ordered by the gene label rather than their genomic position. This can make it a bit confusing to try to understand the structure of the CNVs. Would be better to order by genomic position.

sanjaynagi:
To resolve #558 . 

Currently the funestus GFF is not properly sorted, and so when doing advanced diplotype clustering, the genes are displayed in the wrong order, making it difficult to infer patterns and spans of CNVs. 

Now we sort the GFF by contig, start in the genome_features() function.


leehart (approved):
Thanks @sanjaynagi

Issue #598, opened 2024-09-10
Warning while running tests/anoph/test_fst.py for af1: RuntimeWarning: invalid value encountered in divide


jonbrenas:
While running tests/anoph/test_fst.py for af1, a RuntimeWarning is raised. This is due to the fact the denominator in the Hudson Fst (i.e., the divergence between the two cohorts) is 0 (i.e., the two cohorts are identical for the whole window). This is actually the correct behaviour so I am not sure what to do about this. My instinct would be to catch the Warning so that it stops showing and maybe raise another one informing the user that the divergence was 0 in at least some of the windows (because the Warning is actually raised by allel, the exact number of such windows isn't immediately available) and that it was chosen that the Fst would be 0 in these windows.

jonbrenas:
That's the warning, for the record:

```
tests/anoph/test_fst.py::test_fst_gwss[af1_sim]
  /Users/jb52/Library/Caches/pypoetry/virtualenvs/malariagen-data-IrlFm8eo-py3.10/lib/python3.10/site-packages/allel/stats/fst.py:619: RuntimeWarning:
  
  invalid value encountered in divide
```

alimanfoo:
Example of using `np.errstate()` to suppress warnings within a specific section of code ... https://github.com/malariagen/malariagen-data-python/blob/b4a3cc96c5339b3fc9937080e8da2fb2813b1fd0/malariagen_data/anoph/snp_frq.py#L522

PR #650, opened 2024-11-05
Remove the transient Warning during pytests

jonbrenas:
Addresses #598.

Not sure it works all the time yet but so far, so good.


leehart (approved):
Thanks @jonbrenas

leehart:
Merging in soon unless objections.

PR #653, opened 2024-11-14
Add gene_labels and labels to plot_genes()

### LINKED ISSUE #407: Label genes
It would be useful to be able to label some genes of interest in a genes plot. 

The bokeh [LabelSet](https://docs.bokeh.org/en/latest/docs/reference/models/annotations.html#bokeh.models.LabelSet) feature looks potentially useful here, see also [bokeh labels docs](https://docs.bokeh.org/en/latest/docs/user_guide/basic/annotations.html#labels).

leehart:
Resolves #407.


alimanfoo:
Looking good...

image

alimanfoo:
> @alimanfoo How does this padding look to you? Screenshot 2024-11-22 at 12 56 18
> 
> Ideally I'd like a more robust mechanism for accommodating these labels, because this is currently tailored to the default height of the plot, i.e. 120, so specifying a different height, e.g. `height=240`, results in a skewed layout, e.g. Screenshot 2024-11-22 at 13 02 44

Thanks @leehart, I think it's probably ok for now, generally unlikely to need to vary the height of this track by much.

alimanfoo:
Hi @leehart, could we rename the `labels` param to `gene_labelset`? That way when the parameter appears in functions like `plot_h12_gwss()` it will be more obvious that the parameter will apply to the genes track.

leehart:
Looks like codecov/patch wants some test coverage for the new `gene_labels` param. Will do.

alimanfoo:
> Looks like codecov/patch wants some test coverage for the new `gene_labels` param. Will do.

Cool, thanks @leehart. Happy to be merged once codecov/patch passes.

Issue #182, opened 2022-03-21
Add taxon and area parameters to plot_frequencies_time_series()


alimanfoo:
Allow to control which taxa and areas to show, without having to recompute the frequencies.

leehart:
Some notes:

Here `ds` is an input param, an xarray.Dataset of variant frequencies:
```
cohort_vars = [v for v in ds if v.startswith("cohort_")]
df_cohorts = ds[cohort_vars].to_dataframe()
```

DataFrames for each cohort are concatenated:
```
dfs = []
  for cohort_index, cohort in enumerate(df_cohorts.itertuples()):
      ds_cohort = ds.isel(cohorts=cohort_index)
      df = pd.DataFrame(
          {
              "taxon": cohort.taxon,
              "area": cohort.area,
              "date": cohort.period_start,
              "period": str(
                  cohort.period
              ),  # use string representation for hover label
              "sample_size": cohort.size,
              "variant": variant_labels,
              "count": ds_cohort["event_count"].values,
              "nobs": ds_cohort["event_nobs"].values,
              "frequency": ds_cohort["event_frequency"].values,
              "frequency_ci_low": ds_cohort["event_frequency_ci_low"].values,
              "frequency_ci_upp": ds_cohort["event_frequency_ci_upp"].values,
          }
      )
      dfs.append(df)
  df_events = pd.concat(dfs, axis=0).reset_index(drop=True)
```


A query is applied to remove events with no observations:
```
df_events = df_events.query("nobs > 0")
```

I suppose we could exclude cohorts that don't match the specified `taxon` or `area` parameters from that concatenation, but it looks like that would still require us to compute the frequencies.

We need to make sure we use the appropriate data in calculations, e.g.:
```
frq = df_events["frequency"]
frq_ci_low = df_events["frequency_ci_low"]
frq_ci_upp = df_events["frequency_ci_upp"]
df_events["frequency_error"] = frq_ci_upp - frq
df_events["frequency_error_minus"] = frq - frq_ci_low
```

But maybe the idea is to simply toggle different taxon and areas just on the plot itself, somehow, without any refreshing or recomputing? (I don't understand yet.)

PR #657, opened 2024-11-18
Add taxon and area params to plot_frequencies_time_series()

leehart:
Re: issue #182

### DIFF
--- a/malariagen_data/anoph/plotly_params.py
+++ b/malariagen_data/anoph/plotly_params.py
@@ -215,3 +215,13 @@ scope: TypeAlias = Annotated[
     ],
     "Scope of the map.",
 ]
+
+taxon: TypeAlias = Annotated[
+    Optional[str],
+    "The taxon to restrict the dataset to.",
+]
+
+area: TypeAlias = Annotated[
+    Optional[str],
+    "The area to restrict the dataset to.",
+]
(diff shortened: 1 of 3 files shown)


leehart:
Do we want to add tests for these new param? Note: existing tests only run `fig = api.plot_frequencies_time_series(ds, show=False)`.

Do we want to support lists of taxa or areas? We could upgrade this functionality to accept lists of strings (of taxa or areas) as well as single strings.

malariagen_data/anoph/plotly_params.py
```diff
@@ -215,3 +215,13 @@
...
+
+taxon: TypeAlias = Annotated[
+    Optional[str],
+    "The taxon to restrict the dataset to.",
+]
+
+area: TypeAlias = Annotated[
+    Optional[str],
+    "The area to restrict the dataset to.",
+]
```
alimanfoo, line 227: Maybe these params should go in the `frq_params` module to be together with other parameters related to frequency calculation and visualisation?

alimanfoo (commented):
Thanks @leehart, looks nice.

I like your suggestion to add support for multiple values, i.e., allow `taxon` and `area` to also be a list or tuple of strings. This could be useful, e.g., when you have good data for `gambiae` and `coluzzii` but limited data for `arabiensis` and so you want to show only `gambiae` and `coluzzii`.

alimanfoo:
> Do we want to add tests for these new param? Note: existing tests only run `fig = api.plot_frequencies_time_series(ds, show=False)`. [Examples of usage were added to the existing notebook, which might suffice.]

Having the notebook examples is nice. We might also need to add some unit tests to maintain coverage.

alimanfoo (approved):
Thanks @leehart, looks good 😊

PR #659, opened 2024-11-21
Add Pf8

eselimnl:
This PR adds a new `Pf8` API class for accessing the Pf8 data resource. 

**Pf8():** The Pf8() class is built upon the `Pf7()` template, and should not disrupt the functionality of `Pf7() `(verified via test_pf7_integration.py).

**Sanger S3 Configuration:** Updates have been made in malariagen_data/util.py to support data access on the Sanger S3 storage. For privacy reasons, the bucket name is not included in the PR (such as in the tests/test_pf8_integration.py).

**Variant/altlen Field Handling:** We observed that the variant/altlen field contains a single value rather than one per ALT, despite the number of ALT being set to 6. The cause of this behavior is unclear, but it may relate to the comment “special case (for altlen, number depends on ALT)” in [scikit-allel](https://scikit-allel.readthedocs.io/en/stable/_modules/allel/io/vcf_read.html#vcf_to_zarr). Since it is possible to calculate this value from  `REF` and `ALT`, we are not concerned about it now. Therefore, this field is excluded from `data_vars` after reading the Pf8 zarr file.

This PR was previously created from a [fork](https://github.com/malariagen/malariagen-data-python/pull/658), so comments made there will also be followed here.


ahernank:
Just FYI, tests are failing here as the github-actions service account doesn't have permissions for the pf8 bucket.
I've assigned the permissions now, so hopefully they should run smoothly now.

eselimnl:
That's extremely helpful, thanks @ahernank

malariagen_data/util.py
```diff
@@ -459,6 +459,16 @@ def init_filesystem(url, **kwargs):
...
+    elif "s3://" in url:
+        config = {
+            "signature_version": "s3",
+            "s3": {"addressing_style": "virtual"},
+        }
+
+        # Create an S3FileSystem with custom endpoint if specified.
+        kwargs["anon"] = kwargs.get("anon", True)  # Default to anonymous access.
+        kwargs["endpoint_url"] = "https://cog.sanger.ac.uk"
+        kwargs["config_kwargs"] = kwargs.get("config_kwargs", config)
```
alimanfoo, line 471:
```suggestion
        kwargs.setdefault("config_kwargs", config)
```

malariagen_data/util.py
```diff
@@ -459,6 +459,16 @@ def init_filesystem(url, **kwargs):
...
             # Chained URL.
             kwargs["gcs"] = dict(token=credentials)
+    elif "s3://" in url:
+        config = {
+            "signature_version": "s3",
+            "s3": {"addressing_style": "virtual"},
+        }
+
+        # Create an S3FileSystem with custom endpoint if specified.
+        kwargs["anon"] = kwargs.get("anon", True)  # Default to anonymous access.
```
alimanfoo, line 469:
```suggestion
        kwargs.setdefault("anon", True)  # Default to anonymous access.
```

malariagen_data/util.py
```diff
@@ -459,6 +459,16 @@ def init_filesystem(url, **kwargs):
...
+        kwargs["endpoint_url"] = "https://cog.sanger.ac.uk"
```
alimanfoo, line 470:
```suggestion
        kwargs.setdwfault("endpoint_url", "https://cog.sanger.ac.uk")
```

malariagen_data/plasmodium.py
```diff
@@ -189,6 +189,9 @@ def variant_calls(self, extended=False, inline_array=True, chunks="native"):
         # Add extended data
         if extended:
             data_vars = self._add_extended_data(root, inline_array, chunks, data_vars)
+            data_vars.pop(
+                "variant_altlen", None
+            )  # variant_altlen is under investigation in Pf8.zarr
```
alimanfoo, line 194: Perhaps it would be possible to handle this by removing the variable from the Pf8 config file, rather than here? In which case no changes to tests for pf7 or pv4 would be needed?

eselimnl: Thanks, changed accordingly.

eselimnl:
> Thanks @eselimnl, a few small suggestions.
> 
> Btw currently the way the S3 URLs are handled will mean that chained URLs like `simplecache::s3://...` won't work. Perhaps that is fine for now, up to you.

Thanks @alimanfoo, I have incorporated your suggestions. As you expected, I receive botocore.exceptions.NoCredentialsError when I tried `simplecache::s3://...`, but I couldn't figure out why this is the case.

eselimnl:
Sorry @alimanfoo, debugging the chained URL issue took a while, but `simplecache::s3://...` works fine now.

alimanfoo:
> Sorry @alimanfoo, debugging the chained URL issue took a while, but `simplecache::s3://...` works fine now.

My apologies, I should've explained a bit better how chained URLs are handled in fsspec, at least as I understand it. In looking at this I've also realised the way this is done for GCS could be clearer. I've taken the liberty to push a commit here which consistifies handling of storage options for both GCS and S3 including the the case where URLs are chained. Would you mind taking a look and testing to see if this works for your S3 bucket?

alimanfoo:
Hold that thought, looks like I've broken some tests, investigating...

alimanfoo:
Bug fixed, ready for testing with S3.

alimanfoo:
> Many thanks for sorting this out, works perfectly fine with S3 in my testing environment.

Cool, thanks @eselimnl. Happy for this to be merged?

eselimnl:
Yes, I am happy for this to be merged

alimanfoo:
Thanks @eselimnl, merged! 🎉

Would you like us to make a release?

eselimnl:
Exciting! Yes please!

ahernank:
Just to note here that the GCS and S3 buckets for Pf8 are not currently accessible (i.e. not public/not part of the authentication process).

eselimnl:
I noticed an oversight in the testing environment: S3 access requires the `s3fs` library to be installed. While this works fine if done on the user side, I realize I could have included it in the package as well. Would you like me to create a new PR to address this?

PR #670, opened 2024-11-27
Ensure that missing CNVs are shown in a different colour in `plot_diplotype_clustering_advanced`

### LINKED ISSUE #559: Display normal copy number (2) differently from missing CNV call (NaN) in advanced diplotype clustering
Currently missing CNV HMM data due to high coverage variance is rendered in white in the CNV heatmap. Normal copy number of 2 is also rendered as white. This can be confusing and allow mistaken interpretation of the data.

E.g., here is a plot with the default value for `cnv_max_coverage_variance`:

image

Here's the same plot but with `cnv_max_coverage_variance` set to None, showing that lots of samples which appeared white previously were due to high coverage variance:

image

jonbrenas:
Resolves #559.

I followed bokeh's lead (and @sanjaynagi's hint) and replaced `NaN`s  with `-1`. I have been trying to use `zhoverformat` to have the `z` value show as `NaN` instead of `-1` but I couldn't find a way to make it work.

Here is a picture of what it looks like:
Screenshot 2024-11-27 at 08 19 47

### DIFF
--- a/malariagen_data/anoph/cnv_params.py
+++ b/malariagen_data/anoph/cnv_params.py
@@ -20,3 +20,6 @@ coverage_calls_analysis: TypeAlias = Annotated[
     `coverage_calls_analysis_ids` property for available values.
     """,
 ]
+
+# This is grey followed by PuOr5
+colorscale_default = ["#cccccc", "#5e3c99", "#b2abd2", "#f7f7f7", "#fdb863", "#e66101"]
--- a/malariagen_data/anoph/cnv_data.py
+++ b/malariagen_data/anoph/cnv_data.py
@@ -821,7 +821,6 @@ class AnophelesCnvData(
         debug = self._log.debug
 
         import bokeh.models as bkmod
-        import bokeh.palettes as bkpal
         import bokeh.plotting as bkplt
 
         region_prepped: Region = parse_single_region(self, region)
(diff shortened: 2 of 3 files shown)


alimanfoo:
> Here is a picture of what it looks like

Looks good!

malariagen_data/anoph/cnv_params.py
```diff
@@ -20,3 +20,6 @@
     `coverage_calls_analysis_ids` property for available values.
     """,
 ]
+
+# This is grey followed by PuOr5
+colorscale_default = ["#cccccc", "#5e3c99", "#b2abd2", "#f7f7f7", "#fdb863", "#e66101"]
```
alimanfoo, line 25: Nice to have this defined in a common location. Could we also now use this within the plot_cnv_hmm_heatmap() function so we know the same colours for CNV states are being used throughout?

jonbrenas:
Thanks @alimanfoo . I have set the palette for `plot_cnv_hmm_heatmap()` to be the same. However, it is hard-coded and I think it would be better to offer users the option to choose their own palette (as is the case for `plot_diplotype_clustering_advanced`) so I will add `palette` to the parameters.

alimanfoo:
> Thanks @alimanfoo . I have set the palette for `plot_cnv_hmm_heatmap()` to be the same. However, it is hard-coded and I think it would be better to offer users the option to choose their own palette (as is the case for `plot_diplotype_clustering_advanced`) so I will add `palette` to the parameters.

Thanks @jonbrenas, good call.

alimanfoo (approved):
Looks good, thanks!

leehart:
@jonbrenas Pylint is telling me that it's "dangerous" to have a mutable list as a default value for an argument, so we might want to remedy this at some point.
https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/dangerous-default-value.html

PR #673, opened 2024-11-28
Replicate anjl params to avoid global import causing delays

### LINKED ISSUE #656: Slow import
Importing `malariagen_data` is a little slow at the moment, likely because `anjl` is imported internally which triggers jit compilation of `anjl` functions. Perhapse we could move to only importing `anjl` when needed within the neighbour-joining functions.

leehart:
* Resolves #656.


alimanfoo (approved):
🌻

Issue #682, opened 2024-12-03
Faster CI


alimanfoo:
As part of #366 we have been moving all Anopheles tests to use simulated data, rather than accessing data from GCS. The action which run these new unit tests ([tests.yml](https://github.com/malariagen/malariagen-data-python/blob/master/.github/workflows/tests.yml)) are relatively fast, e.g., running in ~6 minutes. They also have the benefit that they don't need to access GCS data and so can be run from a PR from a fork.

The action which runs the remaining legacy tests that access GCS data are defined in [legacy_tests.yml](https://github.com/malariagen/malariagen-data-python/blob/master/.github/workflows/legacy_tests.yml). These tests have been gradually removed during the refactoring work, and at some point we might be able to completely remove this action. Until that time, this action currently takes ~13m and should reduce in time with further refactoring ongoing (e.g., via #665).

The slowest action currently is the [notebooks.yml](https://github.com/malariagen/malariagen-data-python/blob/master/.github/workflows/notebooks.yml) action which runs all Jupyter notebooks in the "notebooks" folder. This takes around 41m and accesses data from GCS so cannot be run in a PR from a fork.

It would be great in general if checks that run in PRs run faster and can be run in a PR from a fork.

Here's a suggested approach:

* Move the `notebooks.yml` action to only run as part of the [release.yml](https://github.com/malariagen/malariagen-data-python/blob/master/.github/workflows/release.yml) action. I.e., we only run the notebooks when trying to make a release. Any notebook failures will block a release being uploaded to PyPI, and so we can then fix any problems and retag the release.

* Move the `legacy_tests.yml` action to only run as part of the `release.yml` action, for the same reason as above.

* At a later date, when all of the Anopheles unit tests that access GCS data have been removed via the refactoring work, maybe remove the `legacy_tests.yml` action completely. Although possibly some Plasmodium tests will remain that still access GCS data. In which case rename `legacy_tests.yml` to `integration_tests.yml` to help better clarify the purpose.

alimanfoo:
We could also consider running the notebooks and integration tests on merge to master. I.e., these checks only run after a PR is merged, not during commits to an open PR. This would catch any failures earlier than if they were only run on a release. However, this would rely on someone getting notifications of the failures on master.

alimanfoo:
@leehart what do you think? Should the notebooks and integration checks notebooks run...

* Option 1 - on merge to master
* Option 2 - within the release action (preventing upload to PyPI if they fail)
[...]

PR #686, opened 2024-12-04
Refactor unit and integration tests and corresponding github actions

alimanfoo:
This PR aims to make a clearer distinction between unit and integration tests, where unit tests run against local or simulated data, and integration tests run against real remote data in GCS.

* Reorganise the tests to bring together all tests which run against data in GCS into a common `tests/integration` directory, both for Anopheles and Plasmodium.

* The `legacy_tests.yml` action is renamed to `integration_tests.yml` and explicitly runs only those tests within the `tests/integration` directory.

* The `tests.yml` and `coverage.yml` actions are updated to run all tests **except** those in the `tests/integration` directory.

* Previously also we had a mixture of using poetry or pip to install the package and its dependencies in different actions. For simplicity this PR uses pip in all actions running tests or notebooks.

* Ensure that mypy runs with the package and all its dependencies installed, for more comprehensive type checking, and fix or ignore additional typing errors that come up.

This lays some groundwork for #682 but doesn't resolve it.


tests/test_plasmodium.py
```diff
@@ -186,7 +186,7 @@ def test_setup_returns_config_correctly(self):
         self.assertEqual(
             plasmodium.CONF,
             {
-                "default_url": "gs://test_plasmodium_release/",
+                "default_url": "s3://test_plasmodium_release/",
```
alimanfoo, line 189: This change ensures the tests don't try to access GCS within this test module (using a GCS URL triggers an attempt to authenticate, which then requires secrets, which then won't work on a PR from a fork).

leehart (approved):
Thanks @alimanfoo

eselimnl (approved):
Thanks for including s3fs @alimanfoo!

PR #688, opened 2024-12-06
Palette option in `plot_cnv_hmm_heatmap`

### LINKED ISSUE #687: Colors are hard-coded in `plot_cnv_hmm_heatmap`
Currently, the palette is hard-coded to the default value (that we also use in `gene_cnv_frequencies` ) but we could give users the option of using a different palette (as we do in `gene_cnv_frequencies`).

jonbrenas:
Resolves #687.


alimanfoo (approved):
Thanks @jonbrenas, good call to expose this parameter here as well.

Issue #664, opened 2024-11-25
`plot_frequencies_heatmap` is not in the correct file


jonbrenas:
`plot_frequencies_heatmap` is in snp_frq.py when it is used to plot all kinds of frequencies heat maps. It could probably be moved to util.py (if there are parasite functions that want to use a similar function) or anopheles.py.

This issue replaces the second part of #661. 

Here is @leehart comment on this issue from #661: 

>I'm less comfortable with migrating plot_frequencies_heatmap to util.py or anopheles.py. I'm wary about files like util.py becoming a general purpose catch-all place for all kinds of stuff, like that stereotypical one kitchen drawer! And if we're talking about moving stuff out of anopheles.py, which sounds like a pretty high-level place, something seems wrong about moving other stuff into it, especially plotting functions, within the same breath. But I would need to refresh my memory on the scope of those particular modules. In principle, I would go with grouping like-with-like, but you know those sort of decisions can become subjective or aesthetic, rather than perfectly rational.
>
>Where do we usually put functions (e.g. plotting functions) that are used by both parasite and vector? Is it really util.py?

jonbrenas:
I see what you mean @leehart and I agree. I don't really think there are many (if any) functions that are used by both parasite and vector. I think a good alternative would be to create a new file, e.g. `frq_functions.py` that contains the functions that are shared between `hap_frq.py`, `snp_frq.py` and (future) `cnv_frq.py`. Would that make sense @leehart, @alimanfoo ?

jonbrenas:
Worth mentioning, #630 (the PR that creates `hap_freq.py`) moves a lot of functions (but not the plotting functions) that are shared by all 3 `_advanced` versions of the frequency analyses to `utils.py`. I think that it would make more sense to move them to `frq_functions.py` as well if we choose to go this way.

kebarr:
Closing as fixed in [PR 690](https://github.com/malariagen/malariagen-data-python/pull/690)

PR #690, opened 2024-12-06
Creating a file for the code shared by all `_frequencies` functions

jonbrenas:
Addresses #664 .

I moved the functions to enable the computation of frequencies and the plotting functions (which required to define a new `AnophelesFrequency` class). I still want to move the tests of the plotting functions to their own file instead of `test_snp_frq.py`.


jonbrenas:
The tests for the plotting functions are still only on SNP frequencies. Is that good enough?

tests/anoph/test_frq.py
```diff
@@ -0,0 +1,98 @@
+import pytest
+import plotly.graph_objects as go  # type: ignore
+
+import random
+
+
+@pytest.mark.skip
+def test_plot_frequencies_heatmap(api, frq_df):
```
alimanfoo, line 8: Suggest to rename "check_..." instead of "test_...". That way the function won't be seen as a test by pytest, and so won't need to be marked for skipping. Also it will follow the convention used elsewhere in malariagen_data, where utility functions for use in testing are named "check_...".

Same comment applies to all functions in this module.

alimanfoo (commented):
Hi @jonbrenas, thanks so much. A few comments...

alimanfoo:
> @alimanfoo, I might have found a solution that works for what I wanted to do (namely, import shared testing functions for plotting frequencies into the various `test_*_frq.py`s) by telling `pytest` to ignore them when it reaches their file (it means that the tests now return the number of "passed", "skipped", "failed" tests (as well as warnings). I checked that when the tests are actually run when called in the relevant file, they would show a failure if they failed (i.e., they are not skipped when explicitly called).

Thanks @jonbrenas. As suggested in the code review above, it might be even better to name these utility functions like "check_..." instead of "test_..." then no need for skipping. 

If the importing of functions between test modules seems to work then I guess we're OK! I still don't understand the mechanics of how pytest imports test modules. But if you've confirmed it works then no objection.

jonbrenas:
Thanks @alimanfoo. This all makes a lot of sense.

tests/anoph/conftest.py
```diff
@@ -1013,6 +1013,14 @@ def contigs(self) -> Tuple[str, ...]:
     def random_contig(self):
         return choice(self.contigs)
 
+    def random_transcript_id(self):
+        df_transcripts = self.genome_features.query("type == 'mRNA'")
+        transcript_ids = [
+            t.split(";")[0].split("=")[1] for t in df_transcripts.loc[:, "attributes"]
+        ]
```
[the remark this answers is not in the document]
leehart: Noting here that the code in this PR is now:
```python
    def random_transcript_id(self):
        df_transcripts = self.genome_features.query("type == 'mRNA'")
        transcript_ids = [
            gff3_parse_attributes(t)["ID"] for t in df_transcripts.loc[:, "attributes"]
        ]
        transcript_id = choice(transcript_ids)
        return transcript_id
```

tests/anoph/test_frq.py
leehart: Noting here that the code in this PR Is now:
```python
def check_plot_frequencies_heatmap(api, frq_df):
    fig = api.plot_frequencies_heatmap(frq_df, show=False, max_len=None)
    assert isinstance(fig, go.Figure)

    # Test max_len behaviour.
    with pytest.raises(ValueError):
        api.plot_frequencies_heatmap(frq_df, show=False, max_len=len(frq_df) - 1)

    # Test index parameter - if None, should use dataframe index.
    fig = api.plot_frequencies_heatmap(frq_df, show=False, index=None, max_len=None)

    if "contig" in list(frq_df.columns):
        # Not unique.
        with pytest.raises(ValueError):
            api.plot_frequencies_heatmap(
                frq_df, show=False, index="contig", max_len=None
            )
```

leehart (approved):
Thanks @jonbrenas . This looks good to go, unless you know of any further changes required?

Issue #391, opened 2023-05-03
allele_frequencies_advanced parameter flexibility


alimanfoo:
* Add a "taxon_by" parameter to allow varying which column is used for taxon stratification.
* Allow any value in the "period_by" column to allow for more flexible temporal grouping, particularly useful if want a span of multiple years.

PR #694, opened 2024-12-10
Add taxon_by param to snp_allele_frequencies_advanced(). Add example …

leehart:
Re: issue #391

### DIFF
--- a/malariagen_data/util.py
+++ b/malariagen_data/util.py
@@ -1612,20 +1612,22 @@ def _karyotype_tags_n_alt(gt, alts, inversion_alts):
     return inv_n_alt
 
 
-def prep_samples_for_cohort_grouping(*, df_samples, area_by, period_by):
+def prep_samples_for_cohort_grouping(
+    *, df_samples, area_by, period_by, taxon_by="taxon"
+):
     # Take a copy, as we will modify the dataframe.
     df_samples = df_samples.copy()
 
     # Fix "intermediate" or "unassigned" taxon values - we only want to build
     # cohorts with clean taxon calls, so we set other values to None.
     loc_intermediate_taxon = (
-        df_samples["taxon"].str.startswith("intermediate").fillna(False)
+        df_samples[taxon_by].str.startswith("intermediate").fillna(False)
     )
-    df_samples.loc[loc_intermediate_taxon, "taxon"] = None
+    df_samples.loc[loc_intermediate_taxon, taxon_by] = None
     loc_unassigned_taxon = (
-        df_samples["taxon"].str.startswith("unassigned").fillna(False)
+        df_samples[taxon_by].str.startswith("unassigned").fillna(False)
     )
-    df_samples.loc[loc_unassigned_taxon, "taxon"] = None
+    df_samples.loc[loc_unassigned_taxon, taxon_by] = None
 
     # Add period column.
     if period_by == "year":
@@ -1647,7 +1649,9 @@ def prep_samples_for_cohort_grouping(*, df_samples, area_by, period_by):
     return df_samples
 
 
-def build_cohorts_from_sample_grouping(*, group_samples_by_cohort, min_cohort_size):
+def build_cohorts_from_sample_grouping(
+    *, group_samples_by_cohort, min_cohort_size, taxon_by="taxon"
+):
     # Build cohorts dataframe.
     df_cohorts = group_samples_by_cohort.agg(
         size=("sample_id", len),
@@ -1669,7 +1673,7 @@ def build_cohorts_from_sample_grouping(*, group_samples_by_cohort, min_cohort_si
     # Create a label that is similar to the cohort metadata,
     # although this won't be perfect.
     df_cohorts["label"] = df_cohorts.apply(
-        lambda v: f"{v.area}_{v.taxon[:4]}_{v.period}", axis="columns"
+        lambda v: f"{v.area}_{v[taxon_by][:4]}_{v.period}", axis="columns"
     )
 
     # Apply minimum cohort size.
--- a/malariagen_data/anoph/frq_params.py
+++ b/malariagen_data/anoph/frq_params.py
@@ -80,3 +80,10 @@ areas: TypeAlias = Annotated[
     Optional[Union[str, List[str], Tuple[str, ...]]],
     "The area or areas to restrict the dataset to.",
 ]
+
+taxon_by: TypeAlias = Annotated[
+    str,
+    "The column to use for taxon stratification.",
+]
+
+taxon_by_default: taxon_by = "taxon"
--- a/malariagen_data/anoph/snp_frq.py
+++ b/malariagen_data/anoph/snp_frq.py
@@ -452,6 +452,7 @@ class AnophelesSnpFrequencyAnalysis(
         ci_method: Optional[frq_params.ci_method] = frq_params.ci_method_default,
         chunks: base_params.chunks = base_params.native_chunks,
         inline_array: base_params.inline_array = base_params.inline_array_default,
+        taxon_by: frq_params.taxon_by = frq_params.taxon_by_default,
     ) -> xr.Dataset:
         # Load sample metadata.
         df_samples = self.sample_metadata(
@@ -465,15 +466,17 @@ class AnophelesSnpFrequencyAnalysis(
             df_samples=df_samples,
             area_by=area_by,
             period_by=period_by,
+            taxon_by=taxon_by,
         )
 
         # Group samples to make cohorts.
-        group_samples_by_cohort = df_samples.groupby(["taxon", "area", "period"])
+        group_samples_by_cohort = df_samples.groupby([taxon_by, "area", "period"])
 
         # Build cohorts dataframe.
         df_cohorts = build_cohorts_from_sample_grouping(
             group_samples_by_cohort=group_samples_by_cohort,
             min_cohort_size=min_cohort_size,
+            taxon_by=taxon_by,
         )
 
         # Early check for no cohorts.
@@ -529,7 +532,8 @@ class AnophelesSnpFrequencyAnalysis(
             desc="Compute SNP allele frequencies",
         )
         for cohort_index, cohort in cohorts_iterator:
-            cohort_key = cohort.taxon, cohort.area, cohort.period
+            cohort_taxon = getattr(cohort, taxon_by)
+            cohort_key = cohort_taxon, cohort.area, cohort.period
             sample_indices = group_samples_by_cohort.indices[cohort_key]
 
             cohort_ac, cohort_an = _cohort_alt_allele_counts_melt(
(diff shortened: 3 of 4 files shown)


leehart:
Hi @alimanfoo . Do we know how we want to allow any value in the `period_by` column? Looking at the current code, it's not clear to me what we want to do here.

`util.py` currently has:
```python
def prep_samples_for_cohort_grouping(
    *, df_samples, area_by, period_by, taxon_by="taxon"
):
    # Take a copy, as we will modify the dataframe.
    df_samples = df_samples.copy()

    # Fix "intermediate" or "unassigned" taxon values - we only want to build
    # cohorts with clean taxon calls, so we set other values to None.
    loc_intermediate_taxon = (
        df_samples[taxon_by].str.startswith("intermediate").fillna(False)
    )
    df_samples.loc[loc_intermediate_taxon, taxon_by] = None
    loc_unassigned_taxon = (
        df_samples[taxon_by].str.startswith("unassigned").fillna(False)
    )
    df_samples.loc[loc_unassigned_taxon, taxon_by] = None

    # Add period column.
    if period_by == "year":
        make_period = _make_sample_period_year
    elif period_by == "quarter":
        make_period = _make_sample_period_quarter
    elif period_by == "month":
        make_period = _make_sample_period_month
    else:  # pragma: no cover
        raise ValueError(
            f"Value for period_by parameter must be one of 'year', 'quarter', 'month'; found {period_by!r}."
        )
    sample_period = df_samples.apply(make_period, axis="columns")
    df_samples["period"] = sample_period

    # Add area column for consistent output.
    df_samples["area"] = df_samples[area_by]

    return df_samples
```
with
```python
def _make_sample_period_month(row):
    year = row.year
    month = row.month
    if year > 0 and month > 0:
        return pd.Period(freq="M", year=year, month=month)
    else:
        return pd.NaT


def _make_sample_period_quarter(row):
    year = row.year
    month = row.month
    if year > 0 and month > 0:
        return pd.Period(freq="Q", year=year, month=month)
    else:
        return pd.NaT


def _make_sample_period_year(row):
    year = row.year
    if year > 0:
        return pd.Period(freq="Y", year=year)
    else:
        return pd.NaT
```

leehart:
@cclarkson @ahernank @jonbrenas Any thoughts or knowledge regarding the above question for this issue?

jonbrenas:
Your guess is probably as good as mine. Maybe allow to use a column that already exists as "period" or a dictionary of queries? It might be worth asking for details at the next tools meeting.

leehart:
Restricting this PR to just adding the "taxon_by" param.

I've split the original issue into two sub-issues, in order to try to address the `period_by` feature request separately later, when we have a clearer idea of a path forward for that.

leehart:
To do: add to other `_allele_frequencies_advanced` functions.

malariagen_data/util.py
```diff
@@ -1626,7 +1649,9 @@ def prep_samples_for_cohort_grouping(*, df_samples, area_by, period_by):
     return df_samples
 
 
-def build_cohorts_from_sample_grouping(*, group_samples_by_cohort, min_cohort_size):
+def build_cohorts_from_sample_grouping(
+    *, group_samples_by_cohort, min_cohort_size, taxon_by="taxon"
```
alimanfoo, line 1653:
```suggestion
    *, group_samples_by_cohort, min_cohort_size, taxon_by
```

Shouldn't need a default value here.

alimanfoo (commented):
Thanks @leehart. Would be good to add this parameter to the other ..._frequencies_advanced() functions.

leehart:
To do: investigate and resolve potential merge conflict issue around `_karyotype_tags_n_alt()`.

malariagen_data/util.py
leehart: This was needed while the other `..._frequencies_advanced()` functions didn't support the parameter, e.g. calls such as this, in `haplotypes_frequencies_advanced()` and `_gene_cnv_frequencies_advanced()`, where `taxon_by` is not otherwise specified:
```python
        df_cohorts = build_cohorts_from_sample_grouping(
            group_samples_by_cohort=group_samples_by_cohort,
        )
```
...which would result in a problem around this expectation of a `taxon_by ` value in `build_cohorts_from_sample_grouping()`:
```python
    df_cohorts["label"] = df_cohorts.apply(
        lambda v: f"{v.area}_{v[taxon_by][:4]}_{v.period}", axis="columns"
    )
```
...but it looks like this becomes  a non-issue when the other functions gain support, e.g.
```python
        df_cohorts = build_cohorts_from_sample_grouping(
            group_samples_by_cohort=group_samples_by_cohort,
            taxon_by=taxon_by,
        )
```

leehart:
Now that I have some clarity on the sibling issue https://github.com/malariagen/malariagen-data-python/issues/723 I'm tempted to include it under this PR, and then update the PR title accordingly.

leehart:
To do: increase test coverage, in light of codecov test failures.

alimanfoo (commented):
Looking good, just a couple of minor suggestions re parameter checking.

malariagen_data/anoph/cnv_frq.py
```diff
@@ -445,7 +445,15 @@ def gene_cnv_frequencies_advanced(
...
         inline_array: base_params.inline_array = base_params.inline_array_default,
+        taxon_by: Optional[frq_params.taxon_by] = frq_params.taxon_by_default,
     ) -> xr.Dataset:
+        # Check that the taxon_by default hasn't been subverted, e.g. via `taxon_by=None`.
+        # This also satisfies type-checking, e.g. with `getattr`.
+        if taxon_by is None:
+            raise ValueError(
+                f"`taxon_by` cannot be set to `None`. The default would be {frq_params.taxon_by_default!r}."
+            )
+
```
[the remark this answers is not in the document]
leehart: Many thanks @alimanfoo . This was a pleasant surprise!

Somehow I was convinced that `taxon_by` needed to be optional and then had to deal with type-checking issues.

malariagen_data/anoph/frq_base.py
```diff
@@ -14,42 +14,67 @@
...
+
+    # Copy the specified area_by column to a new "area" column.
     df_samples["area"] = df_samples[area_by]
 
     return df_samples
 
 
-def build_cohorts_from_sample_grouping(*, group_samples_by_cohort, min_cohort_size):
+def build_cohorts_from_sample_grouping(
+    *, group_samples_by_cohort, min_cohort_size, taxon_by
```
jonbrenas, line 76: I would have expected that `taxon_by` would be entirely dealt with within `prep_samples_for_cohort_grouping` (as opposed to being propagated to `build_cohorts_from_sample_grouping`). Is there a reason why it is treated differently?

malariagen_data/anoph/frq_base.py
```diff
@@ -71,7 +96,7 @@ def build_cohorts_from_sample_grouping(*, group_samples_by_cohort, min_cohort_si
     # Create a label that is similar to the cohort metadata,
     # although this won't be perfect.
     df_cohorts["label"] = df_cohorts.apply(
-        lambda v: f"{v.area}_{v.taxon[:4]}_{v.period}", axis="columns"
+        lambda v: f"{v.area}_{v[taxon_by][:4]}_{v.period}", axis="columns"
```
jonbrenas, line 99: I know we have always shortened "gambiae" to "gamb" and so on but the first application that comes to my mind is Kwi or Antonio splitting coluzzii in Cameroon into "coluzzii A" and "coluzzii B" ... They would both be called "colu" here.

malariagen_data/anoph/frq_base.py
```diff
@@ -14,42 +14,67 @@
...
 
 
-def prep_samples_for_cohort_grouping(*, df_samples, area_by, period_by):
+def prep_samples_for_cohort_grouping(*, df_samples, area_by, period_by, taxon_by):
     # Take a copy, as we will modify the dataframe.
     df_samples = df_samples.copy()
 
     # Fix "intermediate" or "unassigned" taxon values - we only want to build
     # cohorts with clean taxon calls, so we set other values to None.
     loc_intermediate_taxon = (
```
jonbrenas, line 23: I know this is unchanged but I am not sure why we would prohibit anyone from using an "intermediate" or "unassigned" cohort by default. In almost all cases, there will not be enough samples to form a cohort (and if there ever are enough "intermediate" or "unassigned" samples to form a cohort using the default `cohort_min_size`, we should probably have a good think about what it means) but, in particular if someone manually labeled some samples as "intermediate" in a different column not named "taxon" for the specific purpose of studying them, they will be surprised for them not to show up, I think.

jonbrenas (commented):
None of my comments are required changes and I think @leehart addressed all the concerns expressed by @alimanfoo (but I will let @alimanfoo confirm that it is so).

malariagen_data/anoph/frq_base.py
leehart: Thanks @jonbrenas , that's a good point. It also occurs to me that this won't be robust to other arbitrary strings either, e.g. those that would result in a cohorts label with a space in it, so it probably requires some more thought, e.g. replacing spaces in the string provided or other string sanitisation. Obviously this was originally only designed with a certain set of strings in mind. I wonder if we need two different ways to handle this, depending on the scenario. 🤔 
That comment in the code saying "this won't be perfect" is shining through!

malariagen_data/anoph/frq_base.py
leehart: Thanks @jonbrenas. I'm not sure what you mean by "dealt with", but I can say what I see, in case it elucidates:

`prep_samples_for_cohort_grouping` returns a DataFrame of samples (for SNP, CNV, HAP) in such a way that the DataFrame can then be grouped using `df_samples.groupby([taxon_by, "area", "period"])`, which was more simply `df_samples.groupby(["taxon", "area", "period"])` before this PR. This function needs `taxon_by` in order to deal with "intermediate" or "unassigned" taxon values, the same way that the "taxon" column name was used before the change.

The reason `taxon_by` is being passed to `build_cohorts_from_sample_grouping` is because the `taxon_by` variable, which identifies a column name, needs to be used to get the appropriate taxon value (as a string), which is being used directly in the labelling of the cohorts, via:
```python
    df_cohorts["label"] = df_cohorts.apply(
        lambda v: f"{v.area}_{v[taxon_by][:4]}_{v.period}", axis="columns"
    )
```

Before this PR, it was the simply value of the standard `taxon` column that was being used for this labelling:
```python
    df_cohorts["label"] = df_cohorts.apply(
        lambda v: f"{v.area}_{v.taxon[:4]}_{v.period}", axis="columns"
    )
```

Does that explain, or have I missed something?

jonbrenas: Thanks @leehart. I don't exactly remember what I meant or how the code works exactly but I think that `prep_samples_for_cohort_grouping` basically creates new columns in the metadata (currently "area" and "period") that contain the values that are going to be used for cohorts definition. My question was why is "taxon_by" treated differently? Is it just that the "taxon" column already exists? I know that creating a new column, say "taxon_cohort", with the values in the column identified by "taxon_by" would only duplicate data that is already present but I think that's what we do with "area_by" as well.

jonbrenas: Thanks for the clarification @leehart. 

I think the "clean" solution to avoid conflicts would be to create "private" columns instead of "area", ... e.g., "_area" and to have a check in `add_extra_metadata` that none of the new columns has a name starting with '_' (and add an error if they do). And then propagate the changes in column names to downstream dependencies if there are any.

But that's a lot of work for little to no benefit, so I think the way the code is now is more than satisfactory.

jonbrenas (approved):
LGTM.

malariagen_data/anoph/frq_base.py
leehart: Thanks @jonbrenas . Yes, I like the idea of having "private" columns for internal use and enforcing that. Your Markdown got scrambled, but I got the message!
But I also still like the idea that we wouldn't need the private columns in cases where the user-defined columns are simply being copied, where we can maintain the original column names. That might be beyond the scope of this issue, since you've approved these changes, but I'll open a new issue to stick a pin in it, all the same.

PR #696, opened 2024-12-11
Add @jonbrenas to the list of authors

jonbrenas:
During my appraisal, @alimanfoo and @cclarkson agreed that I could add my name to the list of authors of the `malariagen_data` package.


alimanfoo (approved):
Most certainly, thanks @jonbrenas!

Issue #698, opened 2024-12-11
Moving `karyotype` to anoph


jonbrenas:
Part of #689.

For the record, I think in the future:
> The function is moved to anoph, same reason as above [#697] for future-proofing, even if for a while the function is hidden in funestus or even returns a `NotImplementedError: the tags for the inversion {inversion} have not yet been generated`.

Issue #700, opened 2024-12-11
Add tests on simulated data for `karyotype`


jonbrenas:
Part of #689 .

For the record:
> It [the `karyotype function] has tests on simulated data to be compliant with the rest of the package.

jonbrenas:
As @alimanfoo mentioned in #702, creating the tests "could be tricky because the simulated data is not guaranteed to generate data at the tag SNP positions."

Looking quickly at the code simulating the SNP data, it looks like for each chromosome, a "size" is chosen randomly (between 50 000 and 100 000 for Ag3 and between 80 000 and 120 000 for Af1) and the positions are then assigned (starting at 1) meaning that the positions are all < 120 000, i.e., none of the tags is ever going to have simulated data. 

I see a few possible solutions:

1.  Generate enough simulated SNPs to cover all the regions containing targets (i.e., use a minimum size > the highest value in the tags). That would blow up the size of the simulated data which sounds sub-optimal.
2. Generate extra data for exactly the tags. This would cause the generation of more simulated data but not at the same scale.
3. Use a similar method to the one used for the AIMs, i.e., generate the targets on the fly instead of using the ones from the file. This would require a bit of recoding of `karyotype.py` as the path to the targets is hard-coded to a path in the package (i.e., it cannot be simulated) instead of a path in the data storage (i.e., it can be simulated).

I think 3) would make the most sense but differing opinions are welcome.

Anything that I missed? Any better idea?

leehart:
Just to note that, probably as expected, CI coverage tests are now failing for `karyotype.py`
- e.g. https://app.codecov.io/gh/malariagen/malariagen-data-python/commit/0197957621c6ce6af0fa1dcbb3e658db20519ee7

I gather we might want to increase the priority of this issue in order to get a green tick for test coverage?

PR #702, opened 2024-12-11
Moving `karyotype` to anoph

jonbrenas:
Addresses #698.

The move is done and all tests pass locally but more tests need to be added (for funestus, for example).


jonbrenas:
I added tests to `test_ag3.py` and `test_af1.py` to check that errors were indeed raised when expected. I am not really satisfied yet, though.

malariagen_data/anoph/karyotype.py
```diff
@@ -0,0 +1,112 @@
+import pandas as pd  # type: ignore
+from pandas import CategoricalDtype
+import numpy as np  # type: ignore
+import allel  # type: ignore
+
+from numpydoc_decorator import doc
+from ..util import check_types, _karyotype_tags_n_alt
```
alimanfoo, line 7: If `_karyotype_tags_n_alt` is only used in this module, perhaps it should be moved here.

malariagen_data/anoph/karyotype.py
```diff
@@ -0,0 +1,112 @@
...
+from . import base_params
+from typing import Optional
+
+from .snp_data import AnophelesSnpData
+from .karyotype_params import inversion_param
+
+
+class AnophelesKaryotypeData(AnophelesSnpData):
```
alimanfoo, line 15:
```suggestion
class AnophelesKaryotypeAnalysis(AnophelesSnpData):
```

Nit, suggest slight naming change, then will hopefully also make sense if you later on add in here some functions for calculating inversion allele frequencies.

malariagen_data/anoph/karyotype.py
```diff
@@ -0,0 +1,112 @@
...
+        inversion_tag_path: Optional[str] = None,
+        **kwargs,
+    ):
+        # N.B., this class is designed to work cooperatively, and
+        # so it's important that any remaining parameters are passed
+        # to the superclass constructor.
+        super().__init__(**kwargs)
+
+        # If provided, this analysis version will override the
+        # default value provided in the release configuration.
```
alimanfoo, line 27:
```suggestion
```

Cruft.

malariagen_data/anoph/karyotype.py
```diff
@@ -0,0 +1,112 @@
...
+        self._inversion_tag_path = inversion_tag_path
+
+    @check_types
+    @doc(
+        summary="Load tag SNPs for a given inversion in Ag.",
```
alimanfoo, line 32:
```suggestion
        summary="Load tag SNPs for a given inversion.",
```

malariagen_data/anoph/karyotype.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+    def load_inversion_tags(self, inversion: inversion_param) -> pd.DataFrame:
+        # needs to be modified depending on where we are hosting
+        import importlib.resources
+        from .. import resources
+
+        if not self._inversion_tag_path:
+            raise FileNotFoundError(
+                "The file containing the inversion tags is missing."
+            )
```
alimanfoo, line 42: `FileNotFoundError` isn't quite the right exception class here.

```suggestion
        if self._inversion_tag_path is None:
            raise NotImplementedError(
                "No inversion tags are available for this data resource."
            )
```

malariagen_data/anoph/karyotype.py
```diff
@@ -0,0 +1,112 @@
...
+        else:
+            with importlib.resources.path(resources, self._inversion_tag_path) as path:
+                df_tag_snps = pd.read_csv(path, sep=",")
+            return df_tag_snps.query(f"inversion == '{inversion}'").reset_index()
+
+    @check_types
+    @doc(
+        summary="Infer karyotype from tag SNPs for a given inversion in Ag.",
```
alimanfoo, line 50:
```suggestion
        summary="Infer karyotype from tag SNPs for a given inversion.",
```

malariagen_data/anoph/karyotype_params.py
```diff
@@ -0,0 +1,10 @@
+"""Parameter definitions for karyotype analysis functions."""
+
+from typing import Literal
+
+from typing_extensions import Annotated, TypeAlias
+
+inversion_param: TypeAlias = Annotated[
+    Literal["2La", "2Rb", "2Rc_gam", "2Rc_col", "2Rd", "2Rj"],
+    "Name of inversion to infer karyotype for.",
+]
```
alimanfoo, line 10: These values will only be valid for Ag. Af would have different possible values. Suggest broadening here to just a string.
```suggestion
inversion_param: TypeAlias = Annotated[
    str,
    "Name of inversion to infer karyotype for.",
]
```

tests/integration/test_af1.py
```diff
@@ -75,3 +75,24 @@ def test_locate_region(region_raw):
...
+            af1.karyotype(
+                inversion=inversion, sample_sets="AG1000G-GH", sample_query=None
+            )
+    else:
+        with pytest.raises(FileNotFoundError):
+            af1.karyotype(
+                inversion=inversion,
+                sample_sets="1229-VO-GH-DADZIE-VMF00095",
+                sample_query=None,
+            )
```
alimanfoo, line 98: All `inversion` parameter values should fail with `NotImplementedError` I think.

tests/integration/test_ag3.py
```diff
@@ -159,19 +159,30 @@ def test_xpehh_gwss():
...
-        "inversion",
-        "karyotype_2La_mean",
-        "karyotype_2La",
-        "total_tag_snps",
-    ]
-    assert set(df.columns) == set(expected_cols)
-    assert all(df["karyotype_2La"].isin([0, 1, 2]))
-    assert all(df["karyotype_2La_mean"].between(0, 2))
+    if inversion == "X_x":
+        with pytest.raises(TypeError):
```
alimanfoo, line 170: Should be a `ValueError` I think.

alimanfoo (commented):
Hi @jonbrenas, thanks for this, a few suggestions...

malariagen_data/anoph/karyotype.py
jonbrenas: Yes, it makes sense.

malariagen_data/anoph/karyotype.py
jonbrenas: I agree.

malariagen_data/anoph/karyotype.py
jonbrenas: I have been of two minds about this one: on one hand, it is true that we have not generated the tags for `Af1` and one could argue that `NotImplementedError` is a more appropriate error for work that is still to be done; on the other hand, the code itself would (I think) work if the file actually existed and it is thus less an issue of missing code and more an issue of a missing input. Also, I can imagine a situation where someone created tags for an inversion and would want to try to use their local file (it is not currently possible, the path that is used is hard-coded in both `Ag3` and `Af1`) in which case the error would come from the path being incorrect (though, an error message referring to the actual path inputed would be more helpful if we want to offer this option).

jonbrenas:
Thanks @alimanfoo. I made most of the changes you recommended. As for the tests, I created another issue #700 to deal with it independently of this as I agree that it is not trivial.

malariagen_data/anoph/karyotype.py
leehart: Yes, this is an interesting case, and I have sympathies for both sides. However, I would advise against raising `FileNotFoundError` in cases where the code has not actually looked for a file and found it missing, such as in this case. In this case, we are merely expecting the file to be not found because the variable has not been set, but we haven't technically even looked for it. I would personally prefer to only raise FileNotFoundError in situations like this if the variable has been set to None as a direct consequence of there being no file found, rather than this case where it has been set to None as a direct consequence of the variable not being changed from its default, if I'm reading it correctly. Therefore, I expect that FileNotFoundError would be potentially misleading.

In this situation, where a particular value is begging a halt and an exception to be raised, I would first think of raising a ValueError, since we want to escape from an invalid state, from which we cannot continue on the same path, unless the value is changed to something else before rerunning. However, I can appreciate that this technical fact is not going to be the most useful error for the end user, since they might not know why the value hasn't been set to something valid. I suppose this might be explained in the actual error message. Relatedly, I reckon raising a RuntimeError would be too severe and inappropriate, since we might expect this scenario to happen occasionally.

Raising a `NotImplementedError` would make sense to me in a situation where this variable is set to None as a direct consequence of the functionality not yet being supported for that particular species group, which seems to be the case here, if I reading it correctly. However, what I'm still chewing on is the error message itself. Personally I would want to be explicit as well as helpful in the message, and not make too many assumptions. So I might say something about the fact that the variable has not been set together with the hint that this might mean that we do not yet have inversion tags available for this data resource. If you consider that the exception report will probably show the surrounding code and reveal the precise point of failure around `self._inversion_tag_path is None` then I seem to end up at an identical conclusion. I probably wouldn't object to raising a ValueError with a similarly helpful error message, instead of NotImplementedError, but this seems more like a "lack of support" scenario rather than an "invalid input" scenario.

leehart:
Thanks @jonbrenas . I reckon I'd be happy to approve this once the `FileNotFoundError` exception has been changed to `NotImplementedError` as suggested above. I've attempted to explain my own reasoning, but please feel free to debate it more if you sense it's the wrong direction.

jonbrenas:
Thanks @leehart. I agree with your and Alistair's reasoning.

leehart (approved):
Thanks @jonbrenas

PR #703, opened 2024-12-12
Allow multiple snp_transcripts in `plot_diplotype_clustering_advanced()`

### LINKED ISSUE #600: Allow multiple snp_transcripts in `plot_diplotype_clustering_advanced()`
For gene clusters like Cyp6p/aa, it can be handy to look at amino acid mutations in multiple adjacent genes at the same time, just like we do for CNVs. 

This could be done by providing a `snp_transcript_region`, or just providing multiple `snp_transcripts`

jonbrenas:
Resolves #600.

@KellyLBennett and Nana Amoako needed this functionality earlier today so I sped it up a bit. There are currently no tests for `plot_diplotype_clustering_advanced()` so the best I can say is that it worked in the notebook. I will add tests when I have time.


malariagen_data/anoph/dipclust.py
```diff
@@ -561,7 +563,7 @@ def plot_diplotype_clustering_advanced(
         region: base_params.regions,
         heterozygosity: bool = True,
         heterozygosity_colorscale: plotly_params.color_continuous_scale = "Greys",
-        snp_transcript: Optional[base_params.transcript] = None,
+        snp_transcripts: Sequence[base_params.transcript] = [],
```
alimanfoo, line 566: Hi @jonbrenas, this will be a breaking API change. Instead we could leave the parameter name the same but widen the type so it can accept either an individual transcript or a sequence of transcripts.

malariagen_data/anoph/dipclust.py
```diff
@@ -23,14 +23,16 @@
     cnv_params,
 )
 from .snp_frq import AnophelesSnpFrequencyAnalysis
-from .cnv_data import AnophelesCnvData
+from .cnv_frq import AnophelesCnvFrequencyAnalysis
```
alimanfoo, line 26: Just wondering, why is this change here in this PR?

malariagen_data/anoph/dipclust.py
```diff
@@ -551,7 +553,7 @@ def _dipclust_concat_subplots(
         summary="Perform diplotype clustering, annotated with heterozygosity, gene copy number and amino acid variants.",
         parameters=dict(
             heterozygosity="Plot heterozygosity track.",
-            snp_transcript="Plot amino acid variants for this transcript.",
+            snp_transcripts="Plot amino acid variants for these transcripts.",
```
alimanfoo, line 556: Breaking API change could be avoided here, see similar comment below.

jonbrenas:
Here is a picture (the chosen genes were Cyp6aa1 and Cyp6p15p):

Screenshot 2024-12-13 at 18 02 24

We might want to add titles to make things clearer.

malariagen_data/anoph/dipclust.py
jonbrenas: Yes, we had this debate a while back (I don't remember which function it was for) about whether it was OK to break the API to make sure that the parameter names still make sense and (I think) we agreed that it was the case but that it would only be done for a new major version. 

Given that we had a major version released 2 weeks ago, it might make more sense to keep the parameter name unchanged but change its content and do our best to remember to update it for the next major release. Is that the decision?

malariagen_data/anopheles.py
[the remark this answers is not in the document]
jonbrenas: See comments below.

malariagen_data/anoph/dipclust.py
jonbrenas: `gene_cnv` used to be in `anopheles.py` and the code for `plot_diplotype_clustering_advanced` contains a comment saying that it needs to be moved. This has been done and it is now in `cnv_frq.py`. To be able to use it as `self.gene_cnv()`, the class `AnophelesDipClustAnalysis` thus has to inherit the functions of `AnophelesCNVFrequencyAnalysis`. It worked previously because the API classes (`Ag3` and `Af1` inherit this function) but it breaks the listing and coverage tests to have a class call functions that are not part of its inheritance tree. There were no tests until now so it wasn't raising an issue but that was hardly the best solution.

It may have been clearer to create a separate issue and PR to clean up `AnophelesDipClustAnalysis` and create the tests but given that the changes are relatively minor I thought it would be OK to do both at the same time.

sanjaynagi:
awesome.

malariagen_data/anoph/dipclust.py
leehart: As mentioned in other issues, I'm still of the opinion that we should make such breaking changes in order to extend functionality and align parameter names with their semantics. I appreciate that breaking changes are annoying in some situations, but I personally still favour semantic consistency over convenience.

In my view, using code designed for one major version of a package with a different major version of the package should not be expected to always work flawlessly, not without reading the release notes or documentation and being prepared to adapt the code accordingly. I fear the quality of the package will suffer if developers are dissuaded from making breaking changes that are natural consequences of approved developments, which might also be an extra frustration for the developer. In many case the user of the code, after importing a new major version, will only have to add an "s" to their existing parameter. We can even accept single data type values as alternatives to the list. We could even support both the singular parameter and the plural parameters together at the same time, if needs be, and perhaps eventually deprecate the singular one.

At its root, I suspect this might point to a slight short-sightedness or perhaps a suboptimal habit in greenfield development, where singular parameters are set up without some effort or foresight towards future-proofing, which might have favoured supporting one-or-more values (plurals) instead. Obviously this is not always obvious or convenient to do at the time, and we should adapt accordingly to new requirements.

This probably comes down to different competing concerns and philosophies, and I agree that we shouldn't make things unnecessarily difficult for the users, but I would personally endorse:
1. Prioritising semantic consistency (clarity) and long-term maintainability (future-proofing) over strict backwards compatibility, i.e. not avoiding breaking changes when they improve clarity and functionality.
2. Generally expecting major updates to happen frequently during this period, and that any major version update will probably introduce a breaking change, and ensuring that we communicate these expectations and the relevant details to the package users, so they can prepare and adapt their code accordingly.
3. Adopting some agreed and consistent transition strategy. I'm not sure if we should be aliasing old params, allowing both singular and plural forms, providing clear deprecation warnings, etc., but I expect we want to help minimise the inevitable impact of migrating to new major versions, in any case.

leehart:
Thanks @jonbrenas . There seems to be some general disagreement around how to approach the issue of changing parameters and causing breaking changes. I'll try to formulate a diplomatic strategy in order to resolve my PR https://github.com/malariagen/malariagen-data-python/pull/691 , which might hopefully shed some light on agreeable way forward and also apply to this case, as a package-wide policy.

leehart:
@jonbrenas @alimanfoo I'm coming around to the idea that we should, as many other packages do, maintain support for deprecated parameters for a limited time (although I'm not sure how we'll decide when to drop support) but issue a `DeprecationWarning` preparing them for the upgrade path.

This introduces the problem of kicking the can down the road, in terms of when to eventually drop support completely, but alleviates the problem of users' code suddenly not working without any warning or clue how to fix it, assuming they haven't read the upgrade notes or latest docs.

This strategy would particularly apply to situations like this were we naturally want to change the name and nature of a parameter in order to suit some added or extended functionality. It would allow us the freedom to develop the user-facing interface (the public function signatures) towards clarity and semantic consistency without risking too much friction, allowing smoother and friendlier transitions towards API improvements in functionality without risking clarity.

But I'm still not sure what the policy should be for how long we allow deprecated parameters to linger around and clutter up the interface and codebase. I suppose we could either have a time-based approach, where we completely drop features if they have been deprecated in this manner for over, say, 6 months. Or we could have a release-based approach, where we drop deprecated features after having supported them for, say, 2 unrelated version increments. I guess this creates a sort of tsunami of delayed feature drops and a big cull whenever we bite the bullet and actually cut a major release. We could inform users ahead of time when the feature would be dropped, e.g. in the deprecation warning and docs. Another approach might be to somehow monitor the usage of the old parameter, and finally drop it when usage falls below a certain percentage, in favour of the new parameter... but that sounds hard. I suppose we could also decide on some kind of mixed policy, e.g. whichever comes first: a new major release or 6 months of deprecation.

Putting those details aside, I can certainly see how this new approach should (if adopted) reduce the number of major releases that would otherwise be made under a policy of semantic versioning and fearless development, e.g. whenever we slightly tweak the name of one parameter, as well as being generally more user-friendly.

leehart:
To do: add test(s)

malariagen_data/anoph/dipclust.py
leehart: Hi @jonbrenas . Since this issue is high priority, to be pragmatic and diplomatic, I reckon we should go with @alimanfoo 's suggestion in this situation, but it would still be good to try to reach agreement on a package-wide policy for consistency. We would still need to work out the details for that (e.g. param deprecation strategy, perhaps via PR #691) anyway, which would probably delay this enhancement beyond justification. What do you think? Personally I'm still not entirely comfortable with this, maybe because I'm an idealist at heart... I suppose compromise can be as difficult as naming things!

sanjaynagi:
Personally think it should remain as snp_transcript entirely as the
canonical usage is to use with a single transcript

sanjaynagi:
Well actually, I see what you mean with how we use 'sample_sets' for
example. Understandable to want consistency.

malariagen_data/anoph/dipclust.py
jonbrenas: Hi @leehart. I agree with you, I would also be more comfortable with a uniform naming convention, if possible that relies more on an accurate description of the parameter than on our current understanding of the most common use case. The current status of the code follows @alimanfoo's suggestion (I think) in that the parameter is called `snp_transcript` (no 's') and can take either a transcript or a sequence of transcripts.

leehart (approved):
Thanks @jonbrenas . I reckon this is good to go, unless there's anything more think needs doing here.

leehart:
Merging soon unless objections.

Issue #629, opened 2024-09-26
Better defaults for chunks parameters


alimanfoo:
The `chunks` parameter is exposed in many functions throughout the API and controls how chunks of data in the underlying zarr storage are mapped onto chunks of computation in dask.

It is a key parameter affecting computational performance and scalability.

Currently the default value of this parameter is "native" in all functions, meaning that zarr chunks are mapped directly to dask chunks.

In some cases this can lead to poor performance when running functions that require a large scan across a SNP genotype array, and when running on a distributed cluster. For example, running a PCA across all Ag3 samples on chromosome 3RL. The main problem is that the zarr genotype data is chunked relatively small, around 30MB per chunk, and this leads to a large number of tasks in the dask graph. The size and/or complexity of this graph then causes problems for the dask scheduler.

Through experimentation I've found that mapping 10 zarr chunks to 1 dask chunk, with the zarr chunks being joined along the first (variants) dimension, leads to much better dask performance on a distributed cluster for these large genotype scans.

However, this only applies to "call" arrays like SNP genotype data. We do not want to do the same for "variant" arrays such as position or alleles, as this leads to memory problems. 

Suggest that we provide support for setting separate defaults for "variant" and "call" arrays. The "variant" arrays would retain a default "native" everywhere. The "call" arrays would have a different default when called from functions likely to perform a large scan, such as PCA or GWSS.

PR #704, opened 2025-01-08
Better defaults for chunks parameters

jonbrenas:
Addresses #629.

This is just a prototype to check that it is what @alimanfoo had in mind. I have not tested that it actually improves performances yet.

If that's the correct idea, I will extend that to other similar functions.

### DIFF
--- a/malariagen_data/anoph/base_params.py
+++ b/malariagen_data/anoph/base_params.py
@@ -10,6 +10,8 @@ from ..util import (
     single_contig_param_type,
     single_region_param_type,
     chunks_param_type,
+    zarr_chunks_type,
+    dask_chunks_type,
 )
 
 contig: TypeAlias = Annotated[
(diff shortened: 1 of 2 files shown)


leehart:
@alimanfoo

leehart:
I'll try re-running the integration and notebook tests here, to see if it automatically picks up the fix for gcsfs.

leehart:
By the way, did we ever find out whether this was what was in mind? I wonder if there is a way to confirm, e.g. if performance is improved and nothing breaks and it seems to make sense to others, etc.

jonbrenas:
I am not sure that was exactly what @alimanfoo had in mind. I think there is a piece of chunk optimisation in the PCA part of the cohort analysis which is what he meant to do but I have not checked that it does the same thing as this. I'll move this back to draft to force me to remember to check it.

Issue #708, opened 2025-01-15
ImportError: cannot import name 'BaseStore' from 'zarr.storage'


jonbrenas:
It looks like an import may have broken down in Zarr:
<img width="889" alt="Screenshot 2025-01-15 at 15 44 46" src="https://github.com/user-attachments/assets/a1525cc5-0089-4b47-a59a-c0c95e44ae22" />

sanjaynagi:
Just had the same issue, i assume because Zarr-python 3 has just been released with lots of breaking changes and we do not pin to v2 @alimanfoo

leehart:
@cclarkson

PR #709, opened 2025-01-16
fix zarr-python to v2

sanjaynagi:
Addresses #708 

Zarr-python has been updated to V3, with breaking changes, which is causing import errors in malariagen_data. 

fixing to V2 for now.


leehart:
@alimanfoo

ahernank:
Thanks, I am happy to merge this one as Alistair is deep in grant work today!

PR #718, opened 2025-01-29
Update plot_njt() in distance.py - render_mode="svg"

### LINKED ISSUE #717: `plot_njt()` fails to render on some browser/driver combos
I am finding that larger NJTs (multi-country) wouldn't render on my machine (Windows, Firefox). I would get a message saying "WebGL is not supported by your browser", however WebGL is supported and configured correctly. Plots with fewer samples rendered fine.

I could get the larger plots to render by using `render_mode="svg"`. This is the default for `plot_haplotype_networks()`, perhaps a similar issue was found there? 

We could change `plot_njt()` to match `plot_haplotype_networks`?

cclarkson:
changes the default to `render_mode="svg"` in `plot_njt()` to fix WebGL issues.

resolves #717


ahernank (approved):
Thanks @cclarkson!

Issue #727, opened 2025-03-11
Update switcher.json


leehart:
The current documentation only goes up to version 13.0.0, but the latest release (at time of writing) is on v15.0.1.

leehart:
@alimanfoo It looks like the version switcher has been broken by malformed JSON in your PR https://github.com/malariagen/malariagen-data-python/pull/729 
e.g. https://malariagen.github.io/malariagen-data-python/latest/Ag3.html

I'll now attempt to fix with a new PR.

leehart:
@alimanfoo Issue https://github.com/malariagen/malariagen-data-python/issues/578 is also potentially relevant, hence attempts to minimise the number of listed versions.

PR #731, opened 2025-03-13
Fix switcher.json

leehart:
Re: issue #727


leehart:
@alimanfoo @jonbrenas What's going on with these linting errors from CI checks? I'm getting them on every PR now.

leehart:
Just to note that I suspect the linting failures might be due to an upstream bug (or an accidental breaking change during a minor version update) in the Bokeh 3.7.0 release, which might soon be patched through Bokeh 3.7.1
- https://github.com/malariagen/malariagen-data-python/issues/734

ahernank (approved):

PR #735, opened 2025-03-17
WIP Admixture statistics functions

KellyLBennett:
A draft of the function for f3. I have tried wrapping the f3 function in a class but it is not recognised when trying to run the notebook inside VS code.


leehart:
Thanks @KellyLBennett . Shall I have a look at this and try to modify this code to get the class wrapper working?

I notice that the branch is a little behind master, so we should update that too, but I don't want to confuse matters by pushing my own changes to this branch.

Would it help if I open a different draft PR based on this one, so you can see and apply changes to help get the class wrapper working?

KellyLBennett:
Thanks @leehart your help would be appreciated if you have the time, otherwise I plan to meet with @alimanfoo tomorrow.

KellyLBennett:
@leehart @alimanfoo has helped me change my environmental set up so that the function runs in VS code. I will go ahead and add in some testing for the f3 function.

jonbrenas:
Obsolete. @KellyLBennett is working on a more up-to-date PR.

Issue #734, opened 2025-03-14
Linting failures during CI checks


leehart:
Started happening to all PRs recently. Linting errors unrelated to code changes within the PRs themselves. It does not appear to be related to any recent changes to the code in this package, including GitHub workflows, as far as I can yet determine.

E.g.
- https://github.com/malariagen/malariagen-data-python/actions/runs/13833829346/job/38717985960?pr=731
- https://github.com/malariagen/malariagen-data-python/actions/runs/13854247698/job/38767444733?pr=733

```
Run mypy malariagen_data tests --ignore-missing-imports
  mypy malariagen_data tests --ignore-missing-imports
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.1[2](https://github.com/malariagen/malariagen-data-python/actions/runs/13854247698/job/38767444733?pr=733#step:6:2).9/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/[3](https://github.com/malariagen/malariagen-data-python/actions/runs/13854247698/job/38767444733?pr=733#step:6:3).12.9/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.9/x6[4](https://github.com/malariagen/malariagen-data-python/actions/runs/13854247698/job/38767444733?pr=733#step:6:4)
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.9/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.9/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.12.9/x64/lib
malariagen_data/anoph/genome_features.py:207: error: Unexpected keyword argument "bounds" for "Range1d"  [call-arg]
/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/bokeh/models/ranges.pyi:[5](https://github.com/malariagen/malariagen-data-python/actions/runs/13854247698/job/38767444733?pr=733#step:6:5)3: note: "Range1d" defined here
malariagen_data/anoph/genome_features.py:221: error: Unexpected keyword argument "tools" for "figure"  [call-arg]
malariagen_data/anoph/genome_features.py:221: error: Unexpected keyword argument "active_scroll" for "figure"  [call-arg]
[...]

PR #740, opened 2025-03-21
Resolve CI linting failures

leehart:
Re: issue #734


leehart:
Linting fails for a different reason here, using to Bokeh 3.6.3

```
malariagen_data/anopheles.py:2389: error: Property "layout" defined in "Dash" is read-only  [misc]
Found 1 error in 1 file (checked 94 source files)
```

The relevant code in `anopheles.py` currently looks like this:
```python
        app.layout = html.Div(
            [
                html.Div(
                    cytoscape_component,
                    className="nine columns",
                    style={
                        # required to get cytoscape component to show ...
                        # reduce to prevent scroll overflow
                        "height": f"{height - 50}px",
                        "border": "1px solid black",
                    },
                ),
                html.Div(
                    legend_component,
                    className="three columns",
                    style={
                        "height": f"{height - 50}px",
                    },
                ),
                html.Div(id="output"),
            ],
        )
```

jonbrenas (approved):
Seems sensible to me.

Issue #744, opened 2025-03-27
Error building latest docs using Sphinx


leehart:
> Expecting property name enclosed in double quotes

https://github.com/malariagen/malariagen-data-python/actions/runs/14111451807/job/39530956680

```
Run poetry run sphinx-build -b html docs/source docs/build/html
  poetry run sphinx-build -b html docs/source docs/build/html
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.1[2](https://github.com/malariagen/malariagen-data-python/actions/runs/14111451807/job/39530956680#step:7:2).9/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.12.9/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/[3](https://github.com/malariagen/malariagen-data-python/actions/runs/14111451807/job/39530956680#step:7:3).12.9/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.9/x6[4](https://github.com/malariagen/malariagen-data-python/actions/runs/14111451807/job/39530956680#step:7:4)
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.9/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.12.9/x64/lib
Running Sphinx v8.1.3
loading translations [en]... done
making output directory... done
[autosummary] generating autosummary for: Af1.rst, Ag3.rst, Amin1.rst, index.rst
[...]

PR #745, opened 2025-03-28
Fix Sphinx documentation build error (issue #744)

mohamed-laarej:
# Fix Sphinx documentation build error (Issue #744)

## Problem
The documentation build was failing with a JSON parsing error: "Expecting property name enclosed in double quotes: line 12 column 5 (char 323)".

## Root Cause Analysis
After investigation, I identified multiple issues:

1. The switcher.json file had invalid JSON syntax:
   - Trailing commas after URL entries
   - Malformed URLs with extra forward slashes (https:/// instead of https://) 

2. The conf.py file had configuration issues:
   - The order of keys in the switcher dictionary was causing parsing errors

## Solution
This PR implements a two-part fix:

1. Fixed the switcher.json file:
   - Removed trailing commas to ensure valid JSON syntax
   - Corrected URL formats by removing extra forward slashes

2. Modified the conf.py file:
   - Removed the problematic switcher configuration:

```python
html_theme_options = {
    # Removed switcher configuration
    "navbar_center": ["version-switcher", "navbar-nav"],
    "icon_links": [
        {
            "name": "GitHub",
            "url": "https://github.com/malariagen/malariagen-data-python",
            "icon": "fa-brands fa-github",
        }
    ],
}
```
### Verification
I've verified that with this change, the documentation builds successfully without errors.
### Additional Notes
I also tried fixing the JSON file (removing trailing commas and correcting URLs)  and adding back a minimal switcher configuration, but the error persisted. This suggests there might be deeper compatibility issues with the pydata_sphinx_theme version or how it processes the switcher configuration.

### DIFF
--- a/docs/source/_static/switcher.json
+++ b/docs/source/_static/switcher.json
@@ -2,81 +2,81 @@
     {
         "name": "15.0.1",
         "version": "v15.0.1",
-        "url": "https:///malariagen.github.io/malariagen-data-python/v15.0.1/",
+        "url": "https://malariagen.github.io/malariagen-data-python/v15.0.1/",
         "preferred": true
     },
     {
         "name": "14.0.0",
         "version": "v14.0.0",
-        "url": "https:///malariagen.github.io/malariagen-data-python/v14.0.0/",
+        "url": "https://malariagen.github.io/malariagen-data-python/v14.0.0/"
     },
     {
         "name": "13.5.0",
         "version": "v13.5.0",
-        "url": "https:///malariagen.github.io/malariagen-data-python/v13.5.0/",
+        "url": "https://malariagen.github.io/malariagen-data-python/v13.5.0/"
     },
     {
         "name": "13.4.0",
         "version": "v13.4.0",
-        "url": "https:///malariagen.github.io/malariagen-data-python/v13.4.0/",
+        "url": "https://malariagen.github.io/malariagen-data-python/v13.4.0/"
     },
     {
         "name": "13.3.0",
         "version": "v13.3.0",
-        "url": "https:///malariagen.github.io/malariagen-data-python/v13.3.0/",
+        "url": "https://malariagen.github.io/malariagen-data-python/v13.3.0/"
     },
     {
         "name": "13.2.1",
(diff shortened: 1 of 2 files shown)


mohamed-laarej:
Hi @leehart , I hope you’re doing well! Just wanted to check in on this PR I’d love to get your thoughts whenever you have a moment to review. Thanks so much

leehart:
Thanks @mohamed-laarej .

Just to note that the issue with the trailing commas in the switcher JSON will be covered by issue https://github.com/malariagen/malariagen-data-python/issues/727 . I've also created a new issue for the extra slash in the switcher URLs https://github.com/malariagen/malariagen-data-python/issues/746

By the way, this PR will now need conflict resolution after recent related changes to the master branch.

It's not clear to me how the changes in this PR are expected to resolve issue #744 since they are simply removing the configuration for the switcher from the `html_theme_options`, i.e.:
```json
    "switcher": {
        "check_switcher": True,
        "version_match": version,
        "json_url": "https://malariagen.github.io/malariagen-data-python/latest/_static/switcher.json",
    },
```

You mentioned that you suspected that the order of keys in this dictionary was causing problems. It might be worth noting that the switcher was working until PR https://github.com/malariagen/malariagen-data-python/pull/729 was merged in with the training commas a few weeks ago.

leehart:
Many thanks for your work on this @mohamed-laarej .

Closing this PR as I expect the issue will be resolved via #749

mohamed-laarej:
@leehart , thanks so much for your feedback on the PR it’s really helpful! I just wanted to share why I went with removing the switcher config. I tried a few things first, like cleaning up the JSON (removing trailing commas, tweaking URLs), but the build kept failing. After a quick peek at how `pydata_sphinx_theme` handles JSON, I thought maybe the key order in the switcher dict was causing issues, so I tested rearranging it still no dice. Removing the config was the only thing that got the docs to build, so I ran with that.

I totally see now that there’s a better way to handle this, and I’m all ears for how to do it right next time I want to get this stuff nailed down properly. Super excited to keep contributing and learning let me know if there’s anything else I can tackle!

PR #747, opened 2025-03-31
Fix URLs in switcher.json

leehart:
Re: issue #747


leehart:
This should be merged after https://github.com/malariagen/malariagen-data-python/pull/749 
Otherwise the CI will fail based on the malformed JSON in the existing deployed switcher.

leehart:
I suspect this might still not fix the deployed switcher, ironically because the deployed switcher is broken, so we might need to run a manual redeployment.

Issue #737, opened 2025-03-19
Functions in `anoph/igv.py` errors


jonbrenas:
The reason is that the reference files now require authentication which cannot be done (as far as I know) through the `igv_notebook` API. The solution seems to be to move them to S3 (and  change the path in IGV's config). Previously, @alimanfoo has configured `igv` to look for bam files in `vo_agam_output.cog.sanger.ac.uk`. Is that where we want to move the references?

jonbrenas:
The files are now on S3:
`http://vobs-resources.cog.sanger.ac.uk/reference/genome/agamp4/Anopheles-gambiae-PEST_CHROMOSOMES_AgamP4.fa.fai`
`http://vobs-resources.cog.sanger.ac.uk/reference/genome/agamp4/Anopheles-gambiae-PEST_BASEFEATURES_AgamP4.12.gff3.gz`
`http://vobs-resources.cog.sanger.ac.uk/reference/genome/agamp4/Anopheles-gambiae-PEST_CHROMOSOMES_AgamP4.fa`

I have tried to access the files via `https://igv.org/app/` but I get an error:

<img width="433" alt="Image" src="https://github.com/user-attachments/assets/5bc57414-5391-4579-80fd-a82520623381" />

leehart:
Do we know if the IGV app generally supports accessing resources from Sanger's S3 buckets?

I saw this message on their website, which suggests that GCS would work:
<img width="1333" alt="Image" src="https://github.com/user-attachments/assets/03b0e877-52e9-4f71-bda1-76b383d06ef5" />

leehart:
I figure the URL should be able to load from any source, judging from the docs https://igv.org/doc/webapp/#UserGuide/#loading-the-reference-genome

> Unless your FASTA file is quite small, it is recommended that you also load a corresponding index file. Loading a large FASTA file without an index is likely to freeze the app. The FASTA and index files must be specified at the same time. When loading via the URL option the dialog has entry fields for the web link URL to the file and to the corresponding index file. For the other loading options, both the FASTA file and the index file must be selected from the file chooser that pops up.

And these URLs appear to be returning data
- https://vobs-resources.cog.sanger.ac.uk/reference/genome/agamp4/Anopheles-gambiae-PEST_CHROMOSOMES_AgamP4.fa.fai
[...]

PR #752, opened 2025-04-07
IGV access after authentication requirement for master

jonbrenas:
Addresses #737.

Most of the changes were actually to gcs. I added a new parameter (`public_url`) to the AnophelesBase class which is currently initialised with a hard-coded value. It should be set from the config file but I didn't want to modify the contig without some degree of approval first.

The notebook seemed to show that it works for both agam and afun but the tests will show if it is really the case, I guess.


ahernank (commented):
Thanks @jonbrenas! Wonder if this would be a bit clearer if we define the public URL in the same way as the other GCS URLs, e.g.`GCS_DEFAULT_URL = "gs://vo_agam_release_master_us_central1/` in ag3.py (& equivalent in af1), i.e. globally defined & linking directly to the gsutil file path rather than the public URL (unless this is causing issues with IGV?)

jonbrenas:
Thanks @ahernank. I updated the URL and created a variable containing it. My first attempts at using a GCS URL had failed but it looks like it was due to a typo. 

I don't know if we want to do something similar to what was done for `url` (i.e., it can be inferred from the region ... but with only "us_central1" as an option, I don't think it is worth the faff).

There seems to be an unrelated issue with linting, though.

ahernank (approved):

ahernank:
Approving as just a small change moving the assignment is required.

ahernank:
Actually, decided to keep as-is with the caveat this could create confusion as the public ref only is relevant for IGV.

Issue #753, opened 2025-04-07
Broken import of numcodecs


jonbrenas:
When trying to import `malariagen_data` (at least in Colab and during CI in this repo) an error is raised due to `numcodecs`:

<img width="1027" alt="Image" src="https://github.com/user-attachments/assets/96ec1a4a-ee5a-41d6-b673-e5dde04a3630" />

leehart:
Seems related: https://github.com/zarr-developers/zarr-python/issues/2963

I'll try pinning numcodecs<0.16

PR #754, opened 2025-04-07
Pin numcodecs < 0.16

leehart:
Re: issue #753


ahernank (approved):
Thanks @leehart & @jonbrenas!

jonbrenas (approved):

PR #665, opened 2024-11-25
Moving CNV frequencies functions to their own file

jonbrenas:
Addresses #663.

This passed all tests locally ... despite the fact that some parameters in the functions in the new file were not defined at all which leads be to believe that we never created tests for these functions (maybe a notebook was used instead). I will try to remedy this.

PR #755, opened 2025-04-08
Fix color parameter validation in plot_haplotype_network function - Shadow PR

jonbrenas:
Addresses #738.

Shadows #742.

Thanks @mohamed-laarej for the code. I will remove the draft status once I checked I shadowed the PR correctly.


jonbrenas:
Ed and I would need this merged for the Cameroon coluzzii paper. @leehart, do you think you could find the time to review it?

mohamed-laarej:
Hi @leehart  and @jonbrenas ,

Thanks for reviewing this PR. My memory on the specifics is a bit rusty, but I'll do my best to address your points:

- **`pytest-mock`**: As @jonbrenas noted, this was introduced for testing `Dash` components in the haplotype network function. I'm open to alternative, more consistent testing approaches if available.

- **poetry.lock changes**: These updates reflect the new `pytest-mock` dependency and its sub-dependencies, automatically managed by Poetry for reproducible builds.

- **`color="partition"` line:** I agree this looks a bit odd at first glance. From what I remember, the function creates a temporary "partition" column to standardize color handling regardless of how colors are specified (column name, dictionary, etc.). The line passes this standardized column name to the plotting function. You're right that it could be more readable - perhaps using a private column name like `_partition` would make the intent clearer.

I'm ready to make any necessary adjustments. Thanks again for the review!

leehart (changes requested):
Thanks @jonbrenas @mohamed-laarej . Yep, I reckon we should start using "private" column names, so let's use `_partition` here. If there are any other simple changes we can make to try to improve the readability here, that would be good too.

leehart (approved):
Thanks @mohamed-laarej . Happy with this if @jonbrenas is happy.

jonbrenas:
LGTM. Thanks @mohamed-laarej and @leehart .

leehart:
@jonbrenas I suspect adding tests to `tests/integration/test_ag3.py` and `tests/integration/test_af1.py` might be contrary to issue https://github.com/malariagen/malariagen-data-python/issues/366 which has been removing tests from those files in favour of tests under `tests/anoph`, e.g. https://github.com/malariagen/malariagen-data-python/pull/665 which removed `test_gene_cnv` from `tests/test_af1.py` and `tests/test_ag3.py`. (Although those files themselves were also renamed and refactored during https://github.com/malariagen/malariagen-data-python/pull/686 in an attempt to separate "integration" from "unit" tests.)

In any case, I suggest that we merge this work in and potentially address migrating these tests (and other work related to #366) in a subsequent issue/PR. For instance, it may be that these tests do actually belong alongside the other tests that are currently in `tests/integration/test_ag3.py` and `tests/integration/test_af1.py`, especially because these are testing Dash components.

PR #761, opened 2025-04-15
Updates pf8 default url to s3

eselimnl:
Hello, 
Pf8 data objects are currently stored in both Google Cloud Storage (GCS) and Sanger S3. Based on a recent cost analysis, S3 has been identified as the more cost-effective option. As a result, this PR attempts to update Pf8()'s default URL to point to S3.
Cheers, Eyyub

### DIFF
--- a/malariagen_data/pf8_config.json
+++ b/malariagen_data/pf8_config.json
@@ -1,5 +1,5 @@
 {
-    "default_url": "s3://pf8-release/",
+    "default_url": "s3://pf8-release/" ,
     "metadata_path": "metadata/Pf8_samples.txt",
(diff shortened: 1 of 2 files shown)


ahernank (commented):
Thanks @eselimnl, looks good. If we merge & release here, as the S3 bucket is public, all Pf8 data will be available by default through the API. As you know, this is not currently the case, as the GCS bucket is private at the moment. Can I double check this the expected behaviour that you & @podpearson want now?

nw20:
Hi @ahernank - can confirm we are happy to have Pf8 data available now.

ahernank (approved):

Issue #766, opened 2025-04-28
`sample_metadata` and `lookup_terms_of_use_info ` does not work when there is no terms-of-use info


leehart:
The code should be robust to missing terms-of-use columns in the sample set `manifest.tsv`. (`unrestricted_use` is a nullable boolean.)

PR #767, opened 2025-04-28
Handling absence of terms of use info in `sample_metadata`

jonbrenas:
Addresses #766.

I chose to catch the error raised by `lookup_terms_of_use_info` in `sample_metadata` so that, in the absence of terms of use info (e.g., before release of a sample set), the function remains usable. This means that `sample_metadata` no longer raises an error when the terms of use info is missing in the general case but this probably shouldn't be checked by this function. In cases where the check should still take place, `lookup_terms_of_use_info` still raises an error so it shouldn't affect any other function (though I don't think it used anywhere else).

### DIFF
--- a/malariagen_data/anoph/sample_metadata.py
+++ b/malariagen_data/anoph/sample_metadata.py
@@ -162,10 +162,12 @@ class AnophelesSampleMetadata(AnophelesBase):
                 df[column] = study_info[column]
 
             # Add terms-of-use columns.
-            terms_of_use_info = self.lookup_terms_of_use_info(sample_set=sample_set)
-            for column in terms_of_use_info:
-                df[column] = terms_of_use_info[column]
-
+            try:
+                terms_of_use_info = self.lookup_terms_of_use_info(sample_set=sample_set)
+                for column in terms_of_use_info:
+                    df[column] = terms_of_use_info[column]
+            except ValueError:
+                pass
             return df
 
         else:


leehart:
Thanks @jonbrenas. Although this would seem to work around the issue, I'm concerned that this will just fail silently without any message at all. I appreciate that we probably want it to "fail" silently in this case, but I'm not sure this is the right way to go about it, e.g. it might make it tricky to predict or trace if there were other ValueErrors raised, hypothetically, although there aren't in this case, just as a design/engineering principle. 

It's probably worth noting here that the structure of `lookup_terms_of_use_info` was probably made in such a way to be consistent with its sibling function `lookup_study_info`, which also raises a ValueError in an identical way whenever there is missing study info.

Generally I suspect we shouldn't be controlling normal situations (like missing terms-of-use info) using exceptions anyway. Just ignoring the exception like this doesn't seem good for clarity, either in the code or the user interface, imo.

Ultimately I reckon we want to allow the missingness of terms-of-use info (due to the "normal" (non-exceptional) case of having pre-public release sample sets without terms-of-use), rather than throwing an error and then promptly ignoring it. What do you think? Maybe I need to look at this more closely and see if I can suggest a concrete alternative.

jonbrenas:
Thanks @leehart. I think we have several options:

1. Do a quick and dirty fix to a bug and think more deeply about the issue to address it better at a later date. That is the option I went with in this PR. Risk: we will never take the time to correct the issue properly and the code will increasingly be held by pieces of tape.
2. Take our time to solve the problem correctly right now:

2a. Change our philosophy and stop checking that the data is "correct" as often as possible. Risk: our tests might not be enough and some of the data might become "incorrect".
2b. Correct the code base to have two different states of `sample_metadata` "correctness": released to partners and publicly released. Risk: This probably means having two different functions or a function with an additional parameter, both of which are prone to misuse. There might be an issue when trying to access data that is publicly released and data that is not.
2c. Modify the manifest so that it tells us whether to go look for extra columns or not. Risk: It potentially requires changing a lot of templates and files. There might be an issue when trying to access data that is publicly released and data that is not.

I think I like 2c the best as a long term solution.

mohamed-laarej:
Hi @leehart and @jonbrenas,

Following the discussion on exception handling, I was thinking about an alternative to the current PR that might address the immediate crash while being more informative than a silent `pass`.

What if we refine the exception handling to:
1. Catch specific errors (`KeyError`, `ValueError`, `FileNotFoundError`).
2. Issue a `warnings.warn()` when terms-of-use info is missing.
3. Add the standard terms-of-use columns filled with `NaN`.

**Pros:** This fixes the crash, avoids silent failure, provides a consistent DataFrame structure, and is safer than a broad `except`.
**Cons:** It still uses exceptions for potentially normal flow, and the warning might be noise for some.

This could serve as an intermediate fix for #766. It seems better than the silent failure, though I understand the long-term goal might be a deeper refactor to avoid exceptions entirely here (like modifying the lookup logic or manifests).

Would this intermediate approach be acceptable for now, or is the preference to focus directly on the more fundamental redesign?

jonbrenas:
Thanks @mohamed-laarej, I had considered the 'warning' but decided against it because it is basically the worst of both worlds: It treats the pre-release data as "abnormal", it doesn't address the longer term issue and it creates a new intermediate level of "not-quite-right" state of the data that doesn't make sense if it is limited to this one case (i.e., if we want to use warnings, we have to do it consistently across the whole package which is a huge paradigm shift).

I really think that the best long-term solution is to modify the SNP release process to generate a manifest with either a) a column that says that the data is currently pre-released only or b) `NaN` set as the value for all of the terms of use columns, both of which can be tested for in `sample_metadata` and would result in the columns not being added for pre-released data.

As a side-note, `general_metadata` is called twice in `sample_metadata`, once on its own and once inside `sequence_qc_metadata`. It has no real side-effect but it can probably be avoided.

mohamed-laarej:
Thanks @jonbrenas , for the detailed explanation and clarifying the preferred long-term solution! I understand the reasoning for wanting to modify the manifest generation process itself rather than relying on exception handling.

I'd be happy to help contribute towards implementing this. While I'm keen to contribute, I'm currently unsure about the complexity and potential time involved in modifying the manifest generation process itself. To help me understand the scope and how best I can assist, I have a couple of questions:

1.  Could you please point me towards the relevant scripts, repository locations, or parts of the data release pipeline where the `samples.meta.csv` files are typically generated?
2.  Is there a preference between the two methods you suggested for indicating pre-release status in the manifest: adding a new `release_status` column, or ensuring the terms-of-use columns contain `NaN` values?

Getting these pointers will be very helpful in figuring out the next steps for this part, alongside updating the Python code in `_parse_general_metadata` to read the modified manifest.

Thanks again for the guidance!

jonbrenas:
Thanks @mohamed-laarej, most of the changes that are required are outside of the scope of the `malariagen_data` package so I don't think you can really help here.

mohamed-laarej:
Thanks for the clarification @jonbrenas. I understand that the manifest generation changes are outside the scope of this package.

I appreciate you letting me know. I'm still keen to contribute to `malariagen-data-python`, so please feel free to ping me if help is needed with updating the Python code here to handle the modified manifests once they are ready, or if other issues arise where I might be able to assist.

leehart:
Thanks @jonbrenas and @mohamed-laarej.

Just to recap my understanding:

Testing of the Ag3.15 SNP Data Release is currently failing because use of the `sample_metadata` function for "pre" released sample sets, i.e. sets that haven't been through the Public Data Release process, is erroneously expecting the presence of terms-of-use columns (`terms_of_use_expiry_date` and `terms_of_use_url`), which do not yet exist in the `manifest.tsv` file for the release, i.e. `vo_agam_staging/v3.15/manifest.tsv`.

The current solution proposed in this PR is to catch the ValueError raised by `lookup_terms_of_use_info` during `_parse_general_metadata` and simply pass it. I still feel strongly that this isn't the right path.

Re "2a", I'm not sure I understand the connection to what @jonbrenas says is "our philosophy" regarding checking that the data is correct as often as possible. This specific issue seems to relate to false assumptions about the existence of columns in a file, and the discovery of the issue relates to the testing of the Ag3.15 SNP Data Release, which seems routine and reasonably prudent.

Re "2b", I agree that we clearly have two different cases of "correctness" regarding the released data content, i.e. differences between the Partner Data Release (which we call "pre-release") and the subsequent Public Data Release, which necessarily contains additional data such as terms-of-use. We already use the `pre` setting to branch into different code, e.g. in the current `releases` function. I'm not sure I understand the concern regarding different functions or parameters being "prone to misuse" in this context.

Re "2c", I'm not sure I understand the idea to use the manifest to tell us whether to look for extra columns or not. I still expect that we can avoid any fallout of requiring changes to lots of templates and files, if we can amend the code to allow the missingness of terms-of-use info, for pre-release relases, which seems to be the original design and current intention.

We can probably move the question of potential package-wide changes to exception-handing to a different issue. That seems out of scope for this.

Likewise, if this buggy expectation of term-of-use info in the manifest is holding up the pre-release (Partner Data Release) of Ag3.15, then I would personally understand any kind of provisional terms-of-use data being put into the manifest file just to get past (workaround) the routine SNP Data Release checks. Then we can focus on fixing this properly, i.e. relaxed and cleanly, rather than quick and dirty.

For what it's worth, I agree that we don't want a warning to be issued here, since I currently believe it should be a normal situation for the terms-of-use info to be missing in some releases.

Changing the strategy to always expect (or require) terms-of-use columns but to allow missing values in those columns seems viable, but so too does the current strategy of only adding those columns during the Public Data Release. I suppose there might be an argument for consistency, i.e. so that the manifests always have the same columns, but we I expect should be able to handle missing columns just as easily as missing values. I suppose the consistency of columns in the files might make things like data concatenations and merges more straightforward, but I also suspect that switching to this strategy might cause an unnecessary set of alterations elsewhere, e.g. making sure that the pre-release manifest contains the terms-of-use columns and any other Public Data Release related columns, even though they'll just be blank. I'd personally prefer to keep data relevant and minimal.

Modifying the SNP Data Release process to contain a new column in the manifest doesn't sound right to me. It seems like there would be simpler or clearer ways of signifying whether a particular release was pre-release or not.

I suppose I'm OK with using `NaN` values for the terms-of-use data while the release is still in `pre`, but this seems like a fudge just to appease the erroneous API code that currently only works properly when these columns exist, which wasn't the original plan, and additionally might require further API development and testing to account for that.

Functionally, though, there wouldn't seem to be a difference between a) creating a lot of blank data in the manifest file itself, for missing terms-of-use info, and b) amending `lookup_terms_of_use_info` so that it returns blank terms-of-use info whenever those columns don't exist in `sample_sets()` for whatever reason. I haven't reproduced this error locally or looked into it deeply myself yet, so I'm probably missing some important details. For example, when there are missing terms-of-use info, I expect we still need to generate the derived boolean `unrestricted_use`, as done in `_read_sample_sets`, which suggests that we actually want `unrestricted_use` to be null when there are no terms-of-use info (e.g. for pre-release sets), which makes sense to me personally, perhaps only because I wrote that!

jonbrenas:
Thanks @leehart and @mohamed-laarej.

I created a PR (https://github.com/malariagen/vector-data-processing/issues/195) in `vector-data-processing` as the question is more theoretical (how do we handle 'pre-released' data) than technical. 

Another solution could be to check for terms of use columns in `sample_metadata` only for "public releases" (which are hard-coded in the config file).

leehart:
Thanks @jonbrenas . Closing this PR as we devise a better solution.

Issue #768, opened 2025-04-30
Z-score for pairwise Fst can raise an error


jonbrenas:
When computing the Z score in `plot_pairwise_fst`, if `se == 0` we can get an error. It shouldn't happen on real data but it sometimes does on the simulated data leading me to believe that we should catch this error and return a 'NaN' (and that we may not simulate the data well enough).

PR #769, opened 2025-04-30
Handle ZeroDivisionError in plot_pairwise_average_fst

jonbrenas:
Addresses #768. 

Because it is not supposed to happen, it is hard to test that it works.


leehart (approved):
Thanks @jonbrenas

PR #771, opened 2025-05-12
Modernize NumPy Random Functions and Fix Mypy Errors for #756 - Shadow PR

jonbrenas:
Addresses #756.

Shadows #760.

Thanks @mohamed-laarej. I think there are some troubles with `int` casting, though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,17 @@ the documentation via GitHub Actions.
 The version switcher for the documentation can then be updated by
 modifying the `docs/source/_static/switcher.json` file accordingly.
 
+## Citation
+
+If you use the `malariagen_data` package in a publication
+or include any of its functions or code in other materials (_e.g._ training resources),
+please cite: [doi.org/10.5281/zenodo.11173411](doi.org/10.5281/zenodo.11173411)
+
+Some functions may require additional citations to acknowledge specific contributions. These are indicated in the description for each relevant function.
+
+For any questions, please feel free to contact us at: [support@malariagen.net](mailto:support@malariagen.net)
+
+
 ## Sponsorship
 
 This project is currently supported by the following grants:
--- a/malariagen_data/af1.py
+++ b/malariagen_data/af1.py
@@ -9,6 +9,7 @@ MAJOR_VERSION_NUMBER = 1
 MAJOR_VERSION_PATH = "v1.0"
 CONFIG_PATH = "v1.0-config.json"
 GCS_DEFAULT_URL = "gs://vo_afun_release_master_us_central1/"
+GCS_DEFAULT_PUBLIC_URL = "gs://vo_anoph_temp_us_central1/vo_agam_release/"
 GCS_REGION_URLS = {
     "us-central1": "gs://vo_afun_release_master_us_central1",
 }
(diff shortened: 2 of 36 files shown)


mohamed-laarej:
Hi @jonbrenas, thanks for creating this PR and incorporating my recent random generation updates!

Regarding your comment about "troubles with int casting," I believe the test failures are due to `rng.integers(...)` returning NumPy types `np.int64` which aren't always compatible where Python `int` is expected. Explicitly casting these to `int()` (e.g., `int(rng.integers(...))`) should resolve these TypeErrors.

Happy to help apply these `int()` casting fixes to this branch if that would be useful!

jonbrenas:
Thanks @mohamed-laarej. It would be great if you could address these errors but the idea, at least for now, is that you would make the change in your own PR and I would shadow them here (you can ping me when you think your changes are ready and I'll do the shadowing then). It is a very cumbersome process that should be made simpler once you are added to the team.

mohamed-laarej:
Hi @jonbrenas ,

Thanks for the clarification on the workflow and for your feedback on the int casting issues! I also wanted to share my thanks to @ahernank  for the recent team invitation I appreciate it.

Understood that for these current changes, the plan is for me to address the `int()` casting issues (where `rng.integers` results need to be Python `int`s to resolve the TypeErrors) in my original PR (#760). I will proceed with those fixes there.

I'll be sure to ping you on PR #760 once it's updated. Please let me know if you'd still prefer to shadow the changes, or if I should proceed more directly now that I have team access.

Thanks!

jonbrenas:
Thanks @mohamed-laarej. As I said in the email I just sent you, @ahernank added you to this repo so you should be able to push to this PR from now on if you so choose. 

I think it would be a good idea, for consistency's sake more than correctness of code, if all our `random` calls used the new `numpy` generator methods (instead of only `integers`). For instance, in `test_g123.py`, the test `test_g123_gwss_with_default_sites` uses `random.choice(...)` which has been substituted with `rng.choice(...)` in other tests (e.g., `test_g123_calibration`). Could you do a scan an update all the `random` functions in this PR?

mohamed-laarej:
Hi @jonbrenas,

While fixing the test failures, I ran into a remaining issue in notebooks/karyotype.ipynb:
CI fails with a ValueError: Not enough SNPs when trying to fetch 50,000 SNPs. I suspect this is due to the CI environment using a smaller dataset than local setups.

I'm considering two options to address this:

Add a try-except block to handle the error gracefully

Lower the n_snps value to fit available data

Would you prefer one of these approaches, or suggest a better alternative?

jonbrenas:
Hi @mohamed-laarej, I think lowering `n_snps` is the way to go. For the notebooks, the actual data is used so the fact that it stopped working is a little strange.

mohamed-laarej:
Hi @jonbrenas ,
Apologies for the confusion on the PR. It seems I accidentally pushed some commits related to PR #755 to this branch (GH756-mohamed-laarej-shadow) instead of the correct one (GH738-mohamed-laarej-fix-plot-haplotype-network-color).

I have now force-pushed to remove those commits from this PR. You should see that the commits ee3f144 and 260a0d3 are no longer present here.

My apologies again for the mix-up. I'm now proceeding with pushing the correct changes to the right branch for PR #755.

Thanks for your understanding!

leehart:
Looks like this PR has unresolved conflicts and test failures.

mohamed-laarej:
Yes @jonbrenas , I can take care of the clean-up.

Issue #775, opened 2025-05-14
More heterozygosity support


jonbrenas:
Currently, the way we handle heterozygosity and ROH is less than optimal. It is still part of `anopheles.py` which means that it doesn't have unit tests and it is not as well documented. There is also the fact that most functions are at the sample level making comparisons between cohorts quite difficult. Finally, and maybe most importantly, the heterozygosity for a sample can be ploted but not really accessed (one would need to use `_sample_count_het` which is private and doesn't have a public counterpart).

PR #724, opened 2025-02-07
Support unrestricted_use_only and surveillance_use_only

leehart:

PR #779, opened 2025-05-19
Fix/fake sample count het

Jiya873:
This PR addresses part of the existing issues #775 with sample-level heterozygosity handling by:

* Refactoring the heterozygosity functionality in the [AnophelesDataResource](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) class to expose a public API instead of relying solely on the private `_sample_count_het` method.
* Unit tests (with monkey-patching) for the heterozygosity method will be added to ensure the method returns the expected results.
* Setting up a dummy logging mechanism to avoid dependency issues during testing.

The handling of ROH and cohort-level comparisons remains suboptimal and will be addressed in a separate effort. This change focuses solely on improving heterozygosity accessibility for single samples.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ docs/source/generated
 # gcloud cli
 google-cloud-cli-*
 google-cloud-sdk
+.venv/
--- a/malariagen_data/anopheles.py
+++ b/malariagen_data/anopheles.py
@@ -11,6 +11,7 @@ import pandas as pd
 import plotly.express as px  # type: ignore
 import plotly.graph_objects as go  # type: ignore
 from numpydoc_decorator import doc  # type: ignore
+from .util import parse_single_region
 
 
 from .anoph import (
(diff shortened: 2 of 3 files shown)


jonbrenas:
Thank you @Jiya873. This is great but I think there are several steps that need to be done before we get to the public function:

1. All of the ROH and het functions currently in `anopheles.py` need to be moved to their own file under `anoph/`
2. The tests in `tests/anoph/` need to be structured similarly to what exists for other functions
3. The public functions need to be updated to meet with the newer requirements, including `surveillance_use_only` and `unrestricted_access` that are currently being implemented (see https://github.com/malariagen/malariagen-data-python/pull/724), and we need to make sure that the functions make sense for the analyses that they are connected too. In this case, we need a function that returns the heterozygosity for a single sample, another one that works at the cohort(s) level, and at least one plotting function. 

That's a lot of work (a lot of which is dependent on other things to complete first) so there is no pressure to proceed with this PR quickly.

leehart:
Thanks @jonbrenas . If there's no response after a short time (say, September) then I guess we'll close this. This PR can always be resurrected, if needs be.

Issue #765, opened 2025-04-28
PCA cohort size parameters


tristanpwdennis:
When I run a PCA and specify ```max_cohort_size```, the PCA is run on a random selection of samples equal to the number passed to that argument. E.g. if ```max_cohort_size = 20```,  the resulting PCA is of 20 individuals. 

This random downsampling makes it difficult to look at the placement of a specific population, or in cases where you have a very unbalanced cohort size (so your low _N_ populations will likely be thinned out).

It would be really useful to be able to downsample a PCA by a cohort or other factor (e.g. to no more than 20 samples by ```country```, or ```admin1_region```, or ```location```), but I wasn't sure if this was possible here - am I missing something?

DeepakSilaych:
Wondering if a weighted downsampling approach might be a better alternative to the current random selection, and also suggested cohort-based downsampling?

jonbrenas:
Thanks @DeepakSilaych. Can you explain what you mean by "weighted downsampling approach" in more details? My default interpretation would be to weigh cohorts such that proportions are kept (which would bias against smaller cohorts) but it really depends on how you choose your weights.

If I understand @tristanpwdennis request correctly, the idea would be to add a parameter, e.g. `cohorts`, to the `pca` function.  If it is absent, `max_cohort_size` is the maximum total number of samples that are going to be included in the computation. Otherwise, samples are grouped into cohorts following `cohorts` definition and each such cohort is downsampled to a maximum size equal to `max_cohort_size`.

PR #790, opened 2025-06-02
feat: add cohort-based downsampling to PCA

DeepakSilaych:
### Add cohort-based downsampling to PCA

This PR introduces a new `cohorts` parameter to the PCA API, enabling users to downsample samples by a specified metadata column (e.g., country, location, admin1_region). When `cohorts` is provided, each group is randomly downsampled to a maximum of `max_cohort_size` samples before PCA is performed.

**Key changes:**
- Added `cohorts` parameter to the PCA API and internal methods.
- Implemented per-group downsampling logic using Pandas.
- Updated docstrings and type hints for clarity.
- Added unit tests to verify correct cohort-based downsampling.
- Updated example notebook and API documentation to demonstrate usage.

#765

### DIFF
--- a/docs/source/Ag3.rst
+++ b/docs/source/Ag3.rst
@@ -141,6 +141,22 @@ Principal components analysis (PCA)
     plot_pca_coords
     plot_pca_coords_3d
 
+.. rubric:: Cohort-based downsampling
+
+The ``pca`` method supports a ``cohorts`` parameter. If provided, samples are grouped by the specified column in the sample metadata, and each group is downsampled to a maximum size equal to ``max_cohort_size``. If ``cohorts`` is not provided, ``max_cohort_size`` applies globally to all selected samples.
+
+**Example:**
+
+.. code-block:: python
+
+    df_pca, evr = ag3.pca(
+        region="3L:15,000,000-16,000,000",
+        sample_sets="AG1000G-BF-A",
+        n_snps=10_000,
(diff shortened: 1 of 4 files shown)


DeepakSilaych:
Thanks, will implement the reviews

leehart:
> The addition of the example code is great but I am not sure that it is in the right place in the `.rst`. I would have put it in the docstring of the function. @leehart, do you have an opinion on the matter? Also, it would have to be added (mutatis mutandis) to `Af1.rst` as well.

Adding such code usage examples directly to the `.rst` is an interesting idea, but I don't think we've done that with other functions yet. Instead it looks like we're currently sourcing documentation directly from the code, e.g. [sample_metadata](https://malariagen.github.io/malariagen-data-python/latest/generated/malariagen_data.ag3.Ag3.sample_metadata.html#malariagen_data.ag3.Ag3.sample_metadata),
which has the advantage of being maintained right alongside the function itself and also avoids duplication where functions are common to different classes, as pointed out.

Also, I suspect that we might prefer to maintain code usage examples in something like a Jupyter Notebook rather than an `.rst`. Another consideration is that the docs are automatically generated (via `.github/workflows/latest_docs.yml`) upon merging with `master`, so we might want to avoid adding anything too complicated to these `.rst` files, just to avoid broken docs.

DeepakSilaych:
Ok, will do that, thanks for notifying

Issue #789, opened 2025-05-30
Create functions to access phenotypic data


jonbrenas:
The phenotypic data from the GAARD project is now available on GCS. I followed the same organisation as what existed for the metadata, haplotypes and genotypes, so the data for the sample set `1237-VO-BJ-DJOGBENOU-VMF00050` can be found at `gs://vo_agam_release_master_us_central1/v3.2/phenotypes/all/1237-VO-BJ-DJOGBENOU-VMF00050/phenotypes.csv`. 

The idea is to create a function to access this data similar to what exists for genotypes or haplotypes (e.g., the function `haplotypes` which can probably be used as an example). This function would probably have similar parameters, specifically around `sample_sets`, `sample_queries` and so on. The definition of cohorts will need to be different from the one used elsewhere as the insecticide used, the dose and the phenotype are likely to be more of a discriminant than time or geography.  

For @leehart, all samples are, by definition not `is_surveillance` and `unrestricted_use` so these new functions ought not to raise new issues for #716. 

I think the main function would return an xarray `Dataset` (to be able to handle the metadata about `samples` and `variants` at the same time) but a function that returns a `DataFrame` of the metadata about the `samples` can be a good stepping stone. @mohamed-laarej, feel free to ask questions if things are not quite clear.

mohamed-laarej:
Hi @jonbrenas ,
I’ve been working on implementing phenotype data loading, and I wanted to check if this design direction makes sense before going too far:

I'm thinking of creating a dedicated `phenotypes.py` file under `anoph/` to organize phenotype-specific logic.

The function would be designed to handle multiple sample sets (e.g., Ag3 and Af1), and take either a string or list of sample sets.
I’d ensure it supports future expansion if new datasets are added.

Does this sound in line with how you’d want the project structured? Happy to revise if there's a preferred pattern.

jonbrenas:
Hi @mohamed-laarej,

Yes, creating a new `phenotypes.py` file under `anoph/` is probably the way to go. 

If the function is only accessing the phenotypes, I think it should have  `sample_sets`, `sample_query`, and `sample_query_options` (and maybe `cohort_size`, `min_cohort_size`, `max_cohort_size`) as parameters. If it also includes genotypes and/or haplotypes, `region`, `analysis` and/or a few others might also be needed. You can look at existing functions that access these data to see which parameters make sense.

PR #792, opened 2025-06-04
Feat: Add functions to access insecticide resistance phenotype data

mohamed-laarej:
**Feat: Add functions to access insecticide resistance phenotype data**

**Resolves:** #789

**Description:**

This PR introduces new functionality to access insecticide resistance phenotype data, as requested in issue #789 for GSoC 2025. The primary goal is to provide users with easy access to bioassay results (e.g., alive/dead status after insecticide exposure) alongside sample metadata and optionally linked genetic data.

**Changes Implemented:**

1.  **New Mixin Class:** Created `malariagen_data.anoph.phenotypes.PhenotypeDataMixin` located in the new file `malariagen_data/anoph/phenotypes.py`. This mixin encapsulates all logic related to fetching, processing, and merging phenotype data.
2.  **Integration with Ag3:** The main `Ag3` class (in `malariagen_data/ag3.py`) now inherits from `PhenotypeDataMixin`. This seamlessly integrates the new phenotype methods, making them directly available on `Ag3` instances (e.g., `ag3.phenotype_data(...)`).

**New Functionality Provided by `PhenotypeDataMixin`:**

The following public methods are now available via `Ag3` objects:

*   `phenotype_data(...)`: Loads insecticide bioassay data for specified sample sets from GCS. It merges this with sample metadata and returns a pandas DataFrame. Users can filter by `sample_sets`, `sample_query`, `insecticide`, `dose`, `phenotype`, and apply cohort size constraints (`cohort_size`, `min_cohort_size`, `max_cohort_size`).
*   `phenotype_binary(...)`: Similar to `phenotype_data`, but processes the results to return a pandas Series containing binary phenotype outcomes (1 = alive/resistant, 0 = dead/susceptible), indexed by `sample_id`. This is useful for downstream analyses requiring binary resistance status.
*   `phenotypes(...)`: Provides a comprehensive view by loading phenotype data and optionally merging it with genetic variant data (`snp_calls` or `haplotypes`) for a specified `region`. It returns an xarray Dataset containing both phenotype and genetic information for common samples. The function handles cases where genetic data is unavailable or does not overlap with the phenotype samples for the given query.
*   `phenotype_sample_sets()`: A utility method that returns a list of sample sets for which phenotype data files were found in GCS.

Several internal helper methods (`_load_phenotype_data`, `_merge_phenotype_with_metadata`, `_create_phenotype_binary_series`, `_create_phenotype_dataset`, etc.) support these public functions.

**Testing:**

Local testing using a script confirmed that `phenotype_data` and `phenotype_binary` load and process data correctly. The `phenotypes` function was also tested; the logic for merging phenotype data with genetic data appears sound and correctly handles cases where genetic data is unavailable or does not overlap with the available phenotype samples. However, fully verifying the merge output with actual overlapping data was challenging due to data availability limitations in the local test environment using the currently accessible sample sets.

Formal unit tests covering the new functions will be added as per project guidelines and mentor recommendations following this initial review.

**Next Steps:**

*   Requesting review of the implementation approach and code.
*   Plan to add formal unit tests covering the new functions as per project guidelines after initial feedback.

### DIFF
--- a/malariagen_data/anoph/phenotypes.py
+++ b/malariagen_data/anoph/phenotypes.py
@@ -1,0 +1,387 @@
+import pandas as pd
+import xarray as xr
+from typing import Optional, Union, List, Dict, Any
+import warnings
+
+
+class PhenotypeDataMixin:
+    """
+    Mixin providing methods for accessing insecticide resistance phenotypic data.
+    """
+
+    def _load_phenotype_data(
+        self,
+        sample_sets: List[str],
+        insecticide: Optional[List[str]] = None,
+        dose: Optional[List[float]] = None,
+        phenotype: Optional[List[str]] = None,
+    ) -> pd.DataFrame:
+        """
+        Load raw phenotypic data from GCS for given sample sets.
+        """
+        phenotype_dfs = []
+
+        for sample_set in sample_sets:
+            phenotype_path = self._phenotype_gcs_path_template.format(
+                sample_set=sample_set
+            )
+            try:
+                if not self._fs.exists(phenotype_path):
+                    warnings.warn(f"Phenotype data file not found for {sample_set}")
+                    continue
+
+                with self._fs.open(phenotype_path, "r") as f:
+                    df_pheno = pd.read_csv(f)
+
+                df_pheno["sample_set"] = sample_set
+                phenotype_dfs.append(df_pheno)
+
+            except FileNotFoundError:
+                warnings.warn(f"Phenotype data file not found for {sample_set}")
+                continue
+            except Exception as e:
+                warnings.warn(
+                    f"Unexpected error loading phenotype data for {sample_set}: {e}"
+                )
+                continue
+
+        if not phenotype_dfs:
+            raise ValueError(
+                "No phenotype data could be loaded for the specified sample sets"
+            )
+
+        df_combined = pd.concat(phenotype_dfs, ignore_index=True)
+
+        # Apply simple filters
+        if insecticide is not None:
+            df_combined = df_combined[df_combined["insecticide"].isin(insecticide)]
+        if dose is not None:
+            df_combined = df_combined[df_combined["dose"].isin(dose)]
+        if phenotype is not None:
+            df_combined = df_combined[df_combined["phenotype"].isin(phenotype)]
+
+        return df_combined
+
+    def _merge_phenotype_with_metadata(
+        self,
+        df_phenotypes: pd.DataFrame,
+        df_sample_metadata: pd.DataFrame,
+    ) -> pd.DataFrame:
+        """
+        Merge phenotypic data with sample metadata.
+        """
+        df_merged = df_phenotypes.merge(
+            df_sample_metadata,
+            on="sample_id",
+            how="left",
+            suffixes=("", "_meta"),
+        )
+
+        duplicate_cols = [col for col in df_merged.columns if col.endswith("_meta")]
+        for col in duplicate_cols:
+            base_col = col.replace("_meta", "")
+            if base_col in df_merged.columns:
+                df_merged[base_col] = df_merged[base_col].fillna(df_merged[col])
+            df_merged = df_merged.drop(columns=[col])
+
+        return df_merged
+
+    def _apply_phenotype_cohort_filtering(
+        self,
+        df: pd.DataFrame,
+        cohort_size: Optional[int] = None,
+        min_cohort_size: Optional[int] = None,
+        max_cohort_size: Optional[int] = None,
+    ) -> pd.DataFrame:
+        """
+        Apply cohort size filtering based on insecticide resistance experimental groups.
+        """
+        if all(
+            param is None for param in [cohort_size, min_cohort_size, max_cohort_size]
+        ):
+            return df
+
+        cohort_keys = ["insecticide", "dose", "location", "country"]
+        available_keys = [col for col in cohort_keys if col in df.columns]
+        if not available_keys:
+            warnings.warn("No suitable columns found for cohort grouping")
+            return df
+
+        grouped = df.groupby(available_keys)
+        filtered_groups = []
+
+        for name, group in grouped:
+            group_size = len(group)
+            if min_cohort_size is not None and group_size < min_cohort_size:
+                continue
+            if max_cohort_size is not None and group_size > max_cohort_size:
+                group = group.sample(n=max_cohort_size, random_state=42)
+            elif cohort_size is not None and group_size > cohort_size:
+                group = group.sample(n=cohort_size, random_state=42)
+            elif cohort_size is not None and group_size < cohort_size:
+                continue
+            filtered_groups.append(group)
+
+        if not filtered_groups:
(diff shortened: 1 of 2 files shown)


mohamed-laarej:
Hi @jonbrenas ,

Thank you very much for taking the time to review this PR and for your valuable feedback!
I've pushed new commits addressing the points you raised.
Please let me know if these changes look good or if there's anything else. Thanks again!

malariagen_data/anoph/phenotype_params.py
```diff
@@ -0,0 +1,13 @@
+"""
+Type aliases for phenotype-related parameters.
+"""
+from typing import TypeAlias, Optional, Union, List
+
+# Type alias for insecticide parameter
+insecticide: TypeAlias = Optional[Union[str, List[str]]]
```
jonbrenas, line 7: These should be 'Annotated` to help with the docs.

malariagen_data/anoph/phenotypes.py
```diff
@@ -0,0 +1,505 @@
...
+from . import base_params, phenotype_params
+
+
+class AnophelesPhenotypeData:
+
+    """
+    Provides methods for accessing insecticide resistance phenotypic data.
+    Inherited by AnophelesDataResource subclasses (e.g., Ag3).
+    """
+
```
jonbrenas, line 15: I think you are missing the constructor here.

jonbrenas (changes requested):
Thanks @mohamed-laarej. 

In addition to the tests that were mentioned during our call, I added a few comments here.

jonbrenas:
Thanks @mohamed-laarej. 

I ran a quick debug of your code and the reason why you don't get the `snp_calls` in `phenotypes` is that, in `_create_phenotype_dataset`, you use `variant_data.coords["samples"].values` as the list of sample ids from `variant_data` instead of `variant_data.coords["sample_id"].values`. I haven't looked at why you can't find the haplotypes yet but I am sure you will have it figured out before I get to it.

mohamed-laarej:
Thanks @jonbrenas for the detailed feedback! Really helpful stuff. Let me go through each point:
## Functions as Global Variables
Ah yeah, I can see why that looked weird! I ended up doing it that way because I was getting a bunch of MyPy errors when I first tried the mixin approach:
```console
malariagen_data\anoph\phenotypes.py:31: error: "AnophelesPhenotypeData" has no attribute "_url" [attr-defined]
malariagen_data\anoph\phenotypes.py:377: error: "AnophelesPhenotypeData" has no attribute "sample_metadata" [attr-defined]
```
To fix this, I added type annotations and made them constructor params when you suggested adding a constructor. But after reviewing how the other mixins work, I refactored it to use cooperative inheritance like this:
```python
class AnophelesPhenotypeData:
    # Type annotations for MyPy
    _url: str
    _fs: fsspec.AbstractFileSystem
    sample_metadata: Callable[..., pd.DataFrame]
    # etc...
    
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
```
I think it's much cleaner! Now it just calls `self.sample_metadata()` directly like it should.
Could you please confirm if this is the right approach? Or is there anything else I should consider here to make it more idiomatic or robust?

## Sample Query Filtering
Completely agree on this one! Consolidating everything into `sample_query` will definitely make the API cleaner and more consistent with how the rest of the codebase works.
## Clearer Data Type Selection
This is a great point, the current logic is definitely confusing. I'd love your thoughts on which approach would work best:
### Option A - Separate functions:

```python
def phenotypes_with_snps(self, region, sample_query=None, ...):
def phenotypes_with_haplotypes(self, region, sample_query=None, analysis="arab", ...):
def phenotypes_with_cnvs(self, region, sample_query=None, ...):
def phenotypes_only(self, sample_query=None, ...):
```
### Option B - Single function with parameter:
```python
def phenotypes(self, region=None, genetic_data="snps", sample_query=None, ...):
    # where genetic_data can be "snps", "haplotypes", "cnvs", or None
```
I'm personally leaning toward Option A since it makes the intent really clear and avoids any ambiguity, but I'd definitely defer to your judgment on what would be most intuitive for users.

## Next Steps
I'll get started on:
Cleaning up the inheritance structure based on your guidance
Moving all filtering logic to ` sample_query`
Implementing whichever genetic data selection approach you prefer
Putting together a demo notebook to show how everything works

Thanks again for taking the time to provide such detailed feedback, it's really helpful for understanding the codebase patterns better!

malariagen_data/anopheles.py
```diff
@@ -52,7 +52,7 @@
     parse_single_region,
     plotly_discrete_legend,
 )
-
+from .anoph.phenotypes import AnophelesPhenotypeData
```
jonbrenas, line 55: Shift to before the functions import to be consistent.

jonbrenas (changes requested):
Great work, @mohamed-laarej. In `phenotype_data`, `sample_query_options` isn't used properly: it is passed to `sample_metadata` without a query (i.e., it does nothing) and not used when the query is actually being evaluated.

malariagen_data/anopheles.py
mohamed-laarej: Thanks @jonbrenas ,
I believe that part was added due to a `mypy` issue or possibly a test failure related to `site_mask` values when using ["`gamb_colu_arab`", "`gamb_colu`", "`arab`"] across all fixture types. I don't recall the exact error message, but I think `mypy` (or the tests) was unhappy when some simulators were paired with incompatible or undefined site masks. That’s why I added the` isinstance` check to restrict values to valid options per simulator.

Happy to revisit this and simplify it if needed.

jonbrenas (approved):
LGTM

leehart:
I've always assumed that the blank line between imports at the top of `ag3.py` is used to visually separate external from internal imports. For comparison, there are also blank lines between the imports in `af1.py`, which I took to be visual separations for standard library modules (e.g. `sys`) and external modules (e.g. `plotly.express`) and internal modules (e.g. `.anopheles`).

This might be a style choice and does not need to be changed in this PR. Although the style does not seem to be applied consistently everywhere (e.g. `anoph/aim_data.py`), perhaps because it was never written as policy.

If we decide to police "don't have lines between imports", then that would require a lot of changes. Normally this wouldn't even be noticeable or noteworthy, but there is no other change in `ag3.py` in this PR.

mohamed-laarej:
Hi @jonbrenas, I just wanted to give you a quick heads-up
While reviewing the PR after approval (thank you again!), I realized that the `phenotype_binary` method had been inadvertently removed during the recent refactoring around `sample_query` and `phenotype_data`, and I unfortunately overlooked re-adding it.
Since this method was part of the original functionality and provides a convenient way to access binary phenotype outcomes, I’ve restored it in line with the new design principles, it now uses `sample_query` for filtering. I’ve also added integration tests and updated the demo notebook with relevant examples.
Apologies for the late addition to an already-approved PR. If you’d prefer this be handled in a separate PR instead, I’m happy to make that change just let me know.
Thanks again for your time and understanding!

jonbrenas:
That's good thanks @mohamed-laarej !

PR #793, opened 2025-06-16
Adir1.x

tristanpwdennis:
Hi guys

I've been working on plugging the API into the _Anopheles dirus_ data that has been sitting on GCP for a while now. So far a lot of the API runs fine. There are some modifications I've had to make to the API - mainly to do with commenting out features and bits that the data doesn't have yet that break the API or the metadata when you try and instantiate it. 

Will prob be a little time before this is ready to merge into the rest of the API and producing a public-ready release. I'm still working out what runs and what doesn't run yet, but this is a rough first draft if anyone is interested in having a look or taking it for a spin. 

Will request a review in due course!

PR #795, opened 2025-06-17
Adir1.x

tristanpwdennis:
This replaces PR #793. Previous discussion can be found there.

Continuing previous discussion....

Thanks @leehart - I'll keep working on this and adding features (e.g. cohorts analysis and fixing the metadata so it is compatible with Ag3/Af1), before I review a request.

In the meantime, happy to contribute to some of the re-engineering of functionality to allow more optional implementations of functionality to new classes. If that would be helpful, please point me at whatever I can do.


tristanpwdennis:
Great, thanks Lee. See Issue #798 and PR #799 regarding the caching, and will let you know about Adir1 as we go on.

jonbrenas:
Thanks @tristanpwdennis. I can confirm that the integration tests also pass locally for me so the issue seems to be with CI's access.

I'll look at the V/PUG tomorrow.

The partner study pages are created by the comms team. I can ask them to create the pages.  I'll check with @ahernank tomorrow if it goes under 'Vector Observatory - Asia' or if it is a different project for legal (or more likely financial) reasons.

leehart:
Re: issue #797

tristanpwdennis:
The discrepancy arises because we only did the variant calling on the X largest contigs in the genome for both species. Off top of my head I think it may have been contigs over 100kb?
I suppose an issue may arise if a user wanted to retrieve sequence data for a gene that was in the annotation GFF but wasn't in one of the contigs in the genome. Perhaps the only reason they would do that would be if they had a gene in mind already (e.g. an ortholog, from a paper, from a VB search) - in which case they would encounter an error anyway because either the GFF didn't contain the gene ID OR we didn't have the contig sequence for the region of that gene. 

Another possible case could be if a user specified a contig that wasn't included in the calling but was included in the GFF.

Either way - these are both only problematic if the user has a prior idea of the (tiny) contig they want to analyse first?

Either way the user will encounter an error. Perhaps we could introduce a check for if the contig label / gene ID are present in the GFF and throw an exception explaining the issue if not?

...but maybe I have misunderstood the "Dummy" point you make - what functions are you thinking, here?

jonbrenas:
Thanks @tristanpwdennis.

I agree with what you say. From the work on funestus, I would say there are 2 main ways to get at a gene of interest: after a genome scan showed a signal (which wouldn't be a problem), or searching for an ortholog. In the second case, I think there should be a difference in the error between a user who looks at "2R:28,480,000-28,510,000" (because that's CYP6AA/P in gambiae) and a user who looks at "KB672868:18,122,189-18,146,843" (CYP6P5 in dirus, it is not a great exemple because it is in the contigs but I didn't have time to find a better one). So, yes, I think there should be a check to discriminate between "absurd" contig/gene queries and ones that are "bad" because they point to very small contigs. I am not sure it will be any use so we may wait for an actual use case before we move in that direction, though.

leehart (dismissed):
Please amend the `url` of `setup_adir1` in `test_adir1.py` to use the `vo_adir_release_master_us_central1` bucket, which will then be consistent with `setup_ag3` and `setup_af1`.

jonbrenas:
The only real bug that I have found is to do with `adir1.site_annotations()`. It tries to access the annotations at `https://storage.googleapis.com/download/storage/v1/b/vo_adir_release_us_central1/o/reference%2Fgenome%2FidAnoFuneDA-416_04%2FAnopheles-funestus-DA-416_04_1_SEQANNOTATION.zarr%2F.zmetadata?alt=media`. I assume the value for `adir1._site_annotations_zarr_path` is not set properly and currently contains the path for funestus.

leehart:
Hi @tristanpwdennis - Regarding the docs, yes, those should go live under the "dev" version once the PR has been merged. We'll update the `docs/source/_static/switcher.json` after a new version has been released. There are GitHub workflows `latest_docs.yml` and `tagged_docs.yml` to handle that process. If there are unexpected errors from the new docs, we'll handle those in a subsequent PR. I haven't spotted any obvious problems with them yet and I'll let you know if I find my notes on how to test these docs before merging, if necessary.

malariagen_data/adir1.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Set up caching of some longer-running computations on the local file system,
+    in a directory named "results_cache":
+
+        >>> adir1 = malariagen_data.Adir1(results_cache="results_cache")
+
+    """
+
+    #    _xpehh_gwss_cache_name = XPEHH_GWSS_CACHE_NAME
+    #    _ihs_gwss_cache_name = IHS_GWSS_CACHE_NAME
```
leehart: We should probably remove these commented out lines.

tristanpwdennis:
It's hard to think of a value in the metadata that may apply in the same way as sex_call (e.g. by country or sample set may break?, whereas in all datasets we know the result will be M or F). I've added a tag to the cases so that the tests with those queries only run on ag3 and af1. 

Linting fixed.

Surveillance and restriction instantiated. The tests are failing, but I am unable to reproduce this error locally - thoughts?

jonbrenas:
The tests returned a SEGFAULT (which is not typically an error that you should be able to see in Python) so it seemed likely that it was a transient error with the script. Re-running the tests solved it.

An option for the query would be to build it as "country == C" where C is the country of the first sample in the metadata. Just skipping the test works, at least until we find a better solution, imho.

jonbrenas (dismissed):
LGTM. I will merge later today if no one objects.

PR #799, opened 2025-06-18
Partially fixes 798 by adding hashing support to roh_hmm function in anopheles.py

tristanpwdennis:
Will update this to add hashing support to cohort_diversity_stats too.


jonbrenas:
This is obsolete: ROH has been moved to `anoph/heterozygosity.py` with the hashing support.

Issue #815, opened 2025-09-01
Bug: Incorrect GCS path construction for phenotype data functions


mohamed-laarej:
There is a bug in the path construction logic within the `AnophelesPhenotypeData` class that prevents phenotype data from being loaded in cloud environments like **Bespin**.  

This issue was identified and diagnosed by @jonbrenas (thank you!).  

---

### The Problem  

The methods `_load_phenotype_data` and `phenotype_sample_sets` currently build their GCS paths like this:

```python
base_phenotype_path = f"{self._url}v3.2/phenotypes/all"
```
This leads to two problems:

* Incorrect Path
It produces a malformed path with a missing /, for example:
`gs://...us_central1v3.2/...`

* Hardcoded Version
The `v3.2` is hardcoded, which will break when the API is updated to a new data release.

These functions likely only worked in local testing environments due to file caching, which masked the path issue.

### The Solution
As suggested by @jonbrenas, the fix is to use the `self._base_path` attribute for path construction.
This attribute is correctly formatted and version-aware.

The line should be changed to:

```python
base_phenotype_path = f"{self._base_path}/phenotypes/all"
```
This change needs to be applied in both the `_load_phenotype_data` and `phenotype_sample_sets` methods.

PR #816, opened 2025-09-01
fix(phenotypes): Correct GCS path construction

mohamed-laarej:
Re: issue #815


jonbrenas:
Thanks @mohamed-laarej.

If type-checking works, go with that. We will have to modify the code to use inheritance at some point (to make sure that the package stays (somewhat) consistent) but there are better things for you to spend your time on.

jonbrenas (approved):
LGTM.

PR #817, opened 2025-09-05
add advanced haplotype clustering

sanjaynagi:
This PR adds advanced haplotype clustering, allowing users to cut the tree and assign clusters, and to plot haplotype data (amino acid mutations by default) below the tree. 

TODO 

- [x] Add tests, notebook
- [x] Currently the code loads the SNP genotypes to get SNP effects, this is unnecessary and ideally we can load the SNP effects straight from the haplotypes


sanjaynagi:
We now use the snp_effects() function... didnt know that existed, but makes sense

sanjaynagi:
I also add grey lines to the amino acid plot of the diplotype clustering functions - it looks a lot neater. 

And the haplotype function can now support multiple transcripts.

leehart:
Just adding a corresponding issue for this #819

malariagen_data/anoph/hapclust.py
```diff
@@ -287,8 +301,494 @@ def _haplotype_pairwise_distances(
...
+        chunks,
+        inline_array,
+    ):
+        """
+        Extract haplotype calls for a given transcript.
+        """
+
+        # Get SNP genotype allele counts for the transcript, applying snp_query
+        df_eff = (
+            self.snp_effects(
```
leehart, line 598: Pylint reports:
> Instance of 'AnophelesHapClustAnalysis' has no 'snp_effects' member

leehart (changes requested):

malariagen_data/anoph/hapclust.py
sanjaynagi: Ok. also not clear why the code was working - but i have added the Frequency class to hapclust class now

sanjaynagi: thanks Lee

leehart: Thanks @sanjaynagi . The `snp_effects` function is in `AnophelesSnpFrequencyAnalysis` rather than `AnophelesFrequencyAnalysis`, and `AnophelesFrequencyAnalysis` only inherits from `AnophelesBase`, so probably won't make `snp_effects` available to `AnophelesHapClustAnalysis`, if I understand correctly.

Also, since `AnophelesSnpFrequencyAnalysis` inherits from `AnophelesSnpData`, I suspect we might need to put it before `AnophelesSnpData` in the `AnophelesHapClustAnalysis` base classes list. That made we wonder whether we need `AnophelesSnpData` there at all, but when I tried removing it locally, I got a method resolution error.

Unfortunately, I'm starting to suspect that there might be a larger architectural issue here, around inheritance, namely https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem

PR #818, opened 2025-09-07
Fix phenotype data path construction and remove hardcoded release version

Aditya365x:
This PR fixes issues in phenotype data loading caused by malformed paths and a hardcoded release version:

Path construction
Previously, phenotype paths were built using string concatenation:

base_phenotype_path = f"{self._url}v3.2/phenotypes/all"

This produced malformed paths (missing /) in cloud environments and worked only locally due to file caching.
 Now uses posixpath.join to ensure proper path formatting across GCS and other filesystems.

Hardcoded release version (v3.2)
The hardcoded version caused breakage when new data releases (e.g., v3.3) became available.
 Now uses a dynamic self.release attribute (default fallback v3.2) to make the code forward-compatible.

Updated methods

_load_phenotype_data

phenotype_sample_sets


jonbrenas (approved):
LGTM, though will need to be modified once we get more phenotypes.

PR #825, opened 2025-10-21
Amin1.0

tristanpwdennis:
Hi guys

I've refactored the old amin1 API to include (I think) all the functionality of malariagen_data for unphased data - just like Adir1.0. This has involved:

- Reorganising the release bucket (under sample sets, updating metadata, re-org of filters, reference files, etc).
- Updating the amin1 class file - basically a copy of the Adir1 class file.
- Updating the docs.

Relates to PR #795

### DIFF
--- a/tests/anoph/test_pca.py
+++ b/tests/anoph/test_pca.py
@@ -6,10 +6,10 @@ import plotly.graph_objects as go  # type: ignore
 import pytest
 from pytest_cases import parametrize_with_cases
 
+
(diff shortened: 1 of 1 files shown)


tristanpwdennis:
I think this is running into the same issues as on Adir1 - GH actions can't access the bucket - but @ahernank suggests we need to move the bucket to us-central anyway.

ahernank:
@tristanpwdennis are you happy to close this PR in favour of #839?

Issue #847, opened 2026-01-22
ValueError errors during CI checks


leehart:
These recent CI errors are different to the segfault errors (issue #835) and also aren't reproduced locally.

I expect the segfault error is still an issue, but I guess the tests are hitting this issue before it gets to a segfault.

> 36 passed, 865 errors

There seem to be a  lot of:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
 
I suspect that this might be caused by some unhandled duplicated rows in sourced data files, but that's just an early guess.

Whatever it is, this appears to be very fundamental, affecting many areas, e.g.:
- test_aim_variants:
- test_aim_calls
- test_plot_aim_heatmap
- test_config
- test_releases
- test_client_location
- test_sample_sets_default
- test_sample_sets_release
- test_lookup_release
- test_prep_sample_sets_param
- test_lookup_study
- test_open_cnv_hmm
- test_snp_frq

Example from https://github.com/malariagen/malariagen-data-python/actions/runs/21252008375/job/61159587031
```
=========================== short test summary info ============================
ERROR tests/anoph/test_aim_data.py::test_aim_variants[gambcolu_vs_arab] - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
ERROR tests/anoph/test_aim_data.py::test_aim_variants[gamb_vs_colu] - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
ERROR tests/anoph/test_aim_data.py::test_aim_calls[gambcolu_vs_arab] - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
ERROR tests/anoph/test_aim_data.py::test_aim_calls[gamb_vs_colu] - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
ERROR tests/anoph/test_aim_data.py::test_aim_calls_errors - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
ERROR tests/anoph/test_aim_data.py::test_plot_aim_heatmap[gambcolu_vs_arab] - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
ERROR tests/anoph/test_aim_data.py::test_plot_aim_heatmap[gamb_vs_colu] - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
[...]

PR #850, opened 2026-01-23
Fix ValueErrors

leehart:
Re: issue #847


jonbrenas:
Made obsolete by #958

PR #854, opened 2026-02-02
Fix pandas chained indexing to restore NumPy 2.x compatibility

adilraza99:
### Summary

This PR fixes a NumPy 2.x compatibility issue caused by pandas chained indexing in the `plot_haplotype_network()` code path.

The change replaces chained `.iloc[rows][column]` access with a single-step `.loc[rows, column]` selection, avoiding intermediate object creation that can trigger NumPy 2.x `astype(order=...)` failures on string-backed arrays.

---

### Root Cause

The issue was triggered by the following chained indexing pattern:

```python
df_haps.iloc[list(s)]["_partition"]


jonbrenas (approved):
Minor change but LGTM.

PR #869, opened 2026-02-20
Improve plot_njt errors for insufficient data

### LINKED ISSUE #778: `plot_njt` raises a not very informative `ValueError` when there is no data
When using a "bad" query, as in the example below, `plot_njt` returns a scary error that is not very informative instead of a more useful 'ValueError` stating that no sample satisfy the query.

Image

It should probably be a different issue but it would be advantageous for the `ValueError("Not enough SNPs")` to indicate how many SNPs were found to avoid users having to test a bunch of random values until they find one that works.

Aryan-SINGH-GIT:
Fixes : #778 
This PR improves how plot_njt handles insufficient data. It updates the error message for "not enough SNPs" to show the exact number found versus requested, and adds a check for minimum sample size to prevent obscure failures when fewer than two samples are available. A new test file covers these cases.
image


Aryan-SINGH-GIT:
I have fix  the formating of files, please review  @jonbrenas

jonbrenas (approved):

PR #875, opened 2026-02-20
Fix hardcoded gene type and name attribute in test_plot_genes_with_ge…

### LINKED ISSUE #814: Genome features doesn't seem to find genes in `Af1`
For some reason, `api.genome_features(region=contig).query("type == api._gff_gene_type")` doesn't find any gene for `api == af1` in `test_genomes_features.py`. Somewhere else in the same file `api.genome_features(region=contig).query("type == 'gene'")` is used even for `api == af1`, which shouldn't return anything (but absence of genes is handled differently in this case).

31puneet:
Fixes #814

The test was querying `type == 'gene'` and accessing `["Name"]` directly, both hardcoded for ag3. For af1 and adir1, the gene type is `protein_coding_gene` and the name attribute is `Note`, so the query always returned empty — meaning genes were never actually found for af1/adir1 and the whole gene labels block was silently skipped.

Changed:
- `"type == 'gene'"` → `f"type == '{api._gff_gene_type}'"`  
- `["Name"]` → `[api._gff_gene_name_attribute]`

After the fix, genes are correctly found for af1 and adir1, and [plot_genes()](cci:1://file://wsl.localhost/Ubuntu/home/puneet_/malariagen-data-python/malariagen_data/anoph/genome_features.py:318:4-521:22) with gene labels is now actually tested for all datasets:

### DIFF
--- a/malariagen_data/anoph/genome_features.py
+++ b/malariagen_data/anoph/genome_features.py
@@ -552,5 +552,5 @@ class AnophelesGenomeFeaturesData(AnophelesGenomeSequenceData):
             return self._gene_name_overrides[parent_id]
         except KeyError:
             rec_parent = df_genome_features.loc[parent_id]
-            # Try to access "Name" attribute, fall back to "ID" if not present.
-            return rec_parent.get("Name", parent_id)
+            # Try to access gene name attribute, fall back to "ID" if not present.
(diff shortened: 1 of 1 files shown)


31puneet:
@jonbrenas, all  CI checks have passed! Ready to merge whenever you get a chance. Thanks!

PR #893, opened 2026-02-21
Migrate test_locate_region from integration tests to anoph test suite

GitGautamHub:
Migrated `test_locate_region` test from `tests/integration/test_amin1.py` to `tests/anoph/test_base.py`.

Changes:
- Added `test_locate_region()` function to `tests/anoph/test_base.py` using simulated data fixtures
- Created `amin1_sim_api_snp` fixture for SNP data capabilities
- Updated TODO comment in original test file to note migration
- Test now uses `genome_features()` instead of deprecated `geneset()` method

Benefits:
- Faster execution (no GCS access required)
- Uses modern test framework with simulated data
- Tests same functionality with dynamic data

Test passes: `pytest tests/anoph/test_base.py::test_locate_region`


jonbrenas:
Hi @GitGautamHub, would you mind resolving the conflict before I review this PR?

PR #896, opened 2026-02-21
Fix incorrect cache key in ihs_gwss to prevent cache collisions

### LINKED ISSUE #886: Bug: Incorrect cache key used in ihs_gwss causing potential cache collision
## Description

In `AnophelesDataResource.ihs_gwss`, the cache name is currently hardcoded as:

```
name = "roh"
```

However, this method computes **iHS genome-wide selection scan (GWSS)** statistics, not runs of homozygosity (ROH). Using `"roh"` as the cache key introduces a logical inconsistency and can lead to cache collisions.

The class already defines:

```
@property
@abstractmethod
def _ihs_gwss_cache_name(self):
```

So the method should use this property instead of a hardcoded string.

---

## Why this is a problem

- ROH-related methods also use caching.
- Cached results from different analyses (iHS vs ROH) may overwrite each other.
- `ihs_gwss` may incorrectly return ROH data (or vice versa).
- This could silently return incorrect scientific results without raising an error.

This is a data integrity issue rather than a runtime error.

---

## Expected behavior

`ihs_gwss` should use its own dedicated cache namespace:

---

## Proposed fix

Replace:

```
name = "roh"
```

With:

```
name = self._ihs_gwss_cache_name
```

---

## Impact

High — incorrect cache reuse may lead to silent corruption of analysis results.

--- 
## if its ok with maintainers i would like to open a pr and fix this right away @jonbrenas

blankirigaya:
## Fix incorrect cache key in `ihs_gwss` to prevent cache collisions

Closes #886

### Summary

This PR fixes an incorrect cache key used in the `ihs_gwss` method, which was previously hardcoded as `"roh"`.

Since `"roh"` is the cache namespace used for runs of homozygosity (ROH) analyses, this created a cache collision between ROH and iHS genome-wide selection scan (GWSS) results. Because caching is transparent to the user, this could lead to incorrect results being silently returned.

### Problem

In the current implementation:

```python
name = "roh"
```

However, this method computes **iHS GWSS**, not ROH. Reusing the `"roh"` cache key risks:

- Overwriting ROH results
- Returning cached ROH results for iHS calls
- Silent scientific inaccuracies

### Solution

Replace the hardcoded cache name with the correct dedicated cache key:

```diff
- name = "roh"
+ name = self._ihs_gwss_cache_name
```

This ensures that:

- iHS GWSS results use their own cache namespace
- No collision occurs with ROH analyses
- Cached outputs remain consistent and analysis-specific

### Impact

- Prevents silent cache corruption
- Ensures correctness of iHS results
- No API changes
- Fully backward compatible aside from correcting incorrect cached behavior

---

Let me know if additional test coverage should be added to explicitly guard against cache namespace collisions.


jonbrenas (approved):
LGTM

PR #908, opened 2026-02-22
fix: handle None variant in plot_frequencies_map_markers widget callback

blankirigaya:
## Summary

Fixes a crash in `plot_frequencies_map_markers` when `variant=None` is passed by the widget callback.

The function previously declared:

```python
variant: Union[int, str]
```

However, during widget initialization or state transitions, `None` may be emitted. This caused the assertion:

```python
assert isinstance(variant, str)
```

to fail, resulting in a runtime error.

---

### 1. Update function signature

```python
variant: Optional[Union[int, str]]
```

### 2. Add early guard

```python
if variant is None:
    return
```

### 3. Ensure correct import

```python
from typing import Optional, Union
```

---

## This change:

- Prevents an assertion failure during widget initialization
- Aligns the type hint with actual runtime behavior
- Preserves existing logic for valid `int` and `str` inputs

---

## Testing

- Verified that the widget no longer crashes when `variant=None`
- Confirmed existing behavior remains unchanged for valid variant selections


blankirigaya:
Hi @jonbrenas, thanks for the suggestion! I've updated the fix as requested. Instead of guarding for None in plot_frequencies_map_markers, the fix now lives in plot_frequencies_interactive_map where each dropdown is initialised with its first available value. I've also added a clear ValueError for each of the three parameters in case the dataset has no options available. Let me know if this looks good or if any further changes are needed!

jonbrenas:
If I understand the failures correctly (I have not looked into it yet), raising an error sounds correct. Which would mean that `test_cnv_frq.py` should expect and catch the ValueError. I'll try to look into it to check that I understand the situation correctly.

PR #911, opened 2026-02-22
Guard against invalid jackknife block sizing when n_jack exceeds site count

### LINKED ISSUE #907: n_jack validation missing when number of sites is smaller than jackknife blocks
### Summary

When `n_jack` is greater than the number of available variant sites (`n_sites`), the computed `block_length` becomes 0, which leads to downstream failures in blockwise calculations.

This situation can arise when analyzing small genomic regions or heavily filtered datasets while using the default `n_jack=200`.

---

### Where this occurs

This behavior originates from the block length calculation in:

- `fst.py`
- `anopheles.py` (cohort diversity statistics)

Both compute:

```python
block_length = n_sites // n_jack
```

If `n_jack > n_sites`, integer division produces `block_length = 0`.

This value is then passed to downstream blockwise functions (e.g. in scikit-allel), resulting in cryptic errors that do not indicate the root cause.

---

### Why this matters

- Produces non-intuitive errors that obscure the actual problem  
- Affects key workflows including Fst estimation and diversity statistics  
- Likely to occur when working with small genomic windows or filtered site sets  
- Default parameters work well for whole-chromosome analyses but can fail on smaller inputs  

---

### Suggested improvement

Add an early validation check before computing block lengths:

```python
if n_jack > n_sites:
    raise ValueError(
        f"n_jack ({n_jack}) exceeds the number of available sites ({n_sites}). "
        "Please reduce n_jack or expand the region."
    )
```

This would provide a clear and actionable error message while preventing downstream failures.

---

### Scope

- small input validation safeguard  
- no behavior changes for valid inputs  
- optional test to confirm a `ValueError` when `n_jack > n_sites`  

---

If this approach sounds appropriate, I’d be happy to open a PR.

adilraza99:
## Summary

This PR adds validation to ensure that `n_jack` does not exceed the number
of available variant sites when computing jackknife block lengths.

When `n_jack > n_sites`, integer division produces a block length of 0,
which leads to downstream failures inside blockwise computations
(e.g. scikit-allel) and results in unclear error messages.

This situation can arise when analysing small genomic regions or heavily
filtered datasets while using the default `n_jack=200`.

---

## Problem

Both implementations compute block length as:

```python
block_length = n_sites // n_jack
```

---

Closes #907

### DIFF
--- a/.github/actions/setup-python/action.yaml
+++ b/.github/actions/setup-python/action.yaml
@@ -1,0 +1,22 @@
+name: 'Setup python and poetry'
+inputs:
+  python-version:
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Install poetry
+      shell: bash
+      run: pipx install poetry
+
+    - name: Setup python
+      uses: actions/setup-python@v6
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: poetry
+
+    - name: Install packages
+      shell: bash
+      run: |
+        poetry env use ${{ inputs.python-version }}
+        poetry install --extras dev
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,16 +16,12 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup python
-        uses: actions/setup-python@v5
+        uses: ./.github/actions/setup-python
         with:
-          python-version: "3.10"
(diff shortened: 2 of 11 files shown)


jonbrenas:
Thanks @adilraza99. Unless I missed something, the repetitions might still end up with a failed test if a region that is too small is repeatedly selected.

PR #918, opened 2026-02-23
Fix: Use clip_min as Lower Bound for Y-Axis Range in plot_fst_gwss_track

### LINKED ISSUE #721: possible bug in plot_fst_gwss?
Apologies if I'm wrong about this. 

A colleague is running plot_fst_gwss with min_clip = -1, but is finding that the y axis range of the plot is always 0 to 1. 

Looking at the source code, it looks to me like this is driven by lines 238 and 255 in anoph/fst.py. This is the definition for plot_fst_gwss_track, where the value y_range is set to (0,1) and the value fig.yaxis.ticker is set to [0, 1]. 

I think the y_range should be set appropriately for the range of Fst that come out in the calculations. Otherwise, changing min_clip doesn't serve, because the values don't get displayed anyway. 

Perhaps the minimum value of the y axis could be np.min([0, np.min(fst)]. Max value can be kept at 1 for consistency if desired.

Thanks

Eric

blankirigaya:
# Summary

This PR fixes the y-axis range in plot_fst_gwss_track so that 
setting clip_min actually affects the plot display as expected.

# Problem

The y_range in plot_fst_gwss_track was hardcoded to (0, 1), meaning 
that even when a user passed a custom clip_min value such as -1, 
the y-axis would still display from 0 to 1 and the clipped values 
would never be visible on the plot.

## Changes

malariagen_data/anoph/fst.py — one change:

y_range updated from hardcoded (0, 1) to (clip_min, 1) so the 
y-axis lower bound reflects the clip_min parameter value.

```

python# BEFORE
y_range=(0, 1),

# AFTER
y_range=(clip_min, 1),
```

### Linked Issue

Closes #721


jonbrenas:
Thank you @blankirigaya. Yes, it would make sense to modify the Y ticks accordingly.

PR #932, opened 2026-02-23
fix: define missing Anopheles cache-name properties

### LINKED ISSUE #900: NotImplementedError on roh_hmm(), ihs_gwss(), and xpehh_gwss() across all subclasses
The base class AnophelesDataResource in anopheles.py declares three abstract properties — _roh_hmm_cache_name, _ihs_gwss_cache_name, and _xpehh_gwss_cache_name — which are never implemented in any of the four concrete subclasses (Ag3, Af1, Amin1, Adir1).
As a result, any call to roh_hmm(), ihs_gwss(), or xpehh_gwss() raises a NotImplementedError immediately at runtime, making all ROH and selection scan functionality completely unusable across every supported species.
Impact
High. This affects core analytical methods that researchers depend on for population genetics analysis. There is no workaround — the crash occurs unconditionally regardless of inputs or configuration.

## Steps to Reproduce
```
pythonimport malariagen_data
ag3 = malariagen_data.Ag3()
ag3.ihs_gwss(...)  # raises NotImplementedError
```

## Expected Behaviour
Methods execute and return results as documented.

## Actual Behaviour

NotImplementedError
Raised before any computation begins.

## Proposed Fix

Add the three missing cache name properties to each subclass, using a species-specific prefix to avoid cache collisions between subclasses:
```
python_xpehh_gwss_cache_name = "ag3_xpehh_gwss_v1"
_ihs_gwss_cache_name = "ag3_ihs_gwss_v1"
_roh_hmm_cache_name = "ag3_roh_hmm_v1"
```
See the linked PR for the full implementation across all subclasses.

(came across this bug while using a workflow)
```
import malariagen_data
ag3 = malariagen_data.Ag3()


ag3.ihs_gwss(region="3L", sample_sets="AG1000G-BF-A")

Flamki:
## What problem does this solve?
`roh_hmm()`, `ihs_gwss()`, and `xpehh_gwss()` can hit abstract cache-name properties that are not implemented in all Anopheles subclasses, leading to `NotImplementedError` during use.

## How does it solve it?
- Adds concrete cache-name constants/properties to all four subclasses:
  - `Ag3`: add `_roh_hmm_cache_name`
  - `Af1`: add `_roh_hmm_cache_name`
  - `Amin1`: add `_xpehh_gwss_cache_name`, `_ihs_gwss_cache_name`, `_roh_hmm_cache_name`
  - `Adir1`: add `_xpehh_gwss_cache_name`, `_ihs_gwss_cache_name`, `_roh_hmm_cache_name`
- Adds regression test `tests/test_anopheles_cache_names.py` to enforce these subclass property definitions.

## Relevant issue
Fixes #900

## Testing done
- `pytest -q tests/test_anopheles_cache_names.py -vv`
- Result: `4 passed`

## Breaking changes / migration notes
- None.


jonbrenas:
Obsolete

PR #934, opened 2026-02-24
Set lower default wheel zoom sensitivity for haplotype networks

### LINKED ISSUE #504: Dash set lower wheel zoom sensitivity
For haplotype networks we can now set a lower zoom wheel sensitivity thanks to https://github.com/plotly/dash-cytoscape/pull/203

blankirigaya:
Closes #504 

Exposes wheel_sensitivity as a parameter on plot_haplotype_network(), defaulting to 0.1 (vs cytoscape.js's default of 1). This prevents accidentally zooming out when scrolling over a haplotype network, while still allowing users to tune the sensitivity to their preference.
Made possible by the wheelSensitivity prop added in dash-cytoscape v1.0.0 (plotly/dash-cytoscape#203), which is already our minimum required version.

### Changes:

Add wheel_sensitivity: float = 0.1 parameter to plot_haplotype_network()


blankirigaya:
@jonbrenas Sir, the issue mentioned lower the sensitivity without a specify value so i adjusted it to a value usually used for large graphs to avoid confusion. let me know if the value need adjusting/fine tuning

PR #935, opened 2026-02-24
Fix simplecache storage location being ignored for chained GCS URLs

### LINKED ISSUE #773: Configuration of simplecache is not working in integration tests and notebooks
The integration tests and notebooks make use of the fsspec simplecache feature, which caches files from GCS on the local file system. E.g., this function from the tests/integration/test_ag3.py module by default sets up access to GCS via a simplecache, and tries to configure simplecache to use a folder called "gcs_cache" to store the cached files:

```python
def setup_ag3(url="simplecache::gs://vo_agam_release_master_us_central1/", **kwargs):
    kwargs.setdefault("check_location", False)
    kwargs.setdefault("show_progress", False)
    if url is None:
        # test default URL
        # This only tests the setup_af1 default url, not the Ag3 default.
        # The test_anopheles setup_subclass tests true defaults.
        return Ag3(**kwargs)
    if url.startswith("simplecache::"):
        # configure the directory on the local file system to cache data
        kwargs["simplecache"] = dict(cache_storage="gcs_cache")
    return Ag3(url, **kwargs)
```

Similarly, many of the notebooks do something like this:

```python
ag3 = malariagen_data.Ag3(
    "simplecache::gs://vo_agam_release_master_us_central1",
    simplecache=dict(cache_storage="../gcs_cache"),
)
ag3
```

...again trying to configure a specific folder to use for caching files locally, which is a folder called "gcs_cache" within the repo root dir.

This is configuration is no longer working, and simplecache is ignoring this. Instead, it is using the default temporary directory, usually /tmp, and files are being deleted once the python process is exited. This is not a serious problem, but does mean that file caching is basically not working, and so every test run or notebook run will read directly from GCS again.

I think this is due to changes in the malariagen_data.util.init_filesystem() implementation, that came in around the time support for Pf8 was being added.

blankirigaya:
Fixes #773

### summary

When using a chained URL like simplecache::gs://..., the simplecache options (e.g. cache_storage="gcs_cache") were being nested under the "gs" key in storage_options alongside the GCS credentials, rather than being passed at the top level where fsspec expects them. This caused fsspec to silently ignore the configured cache directory and fall back to the system temp directory (/tmp), meaning cached files were deleted on every process exit and every test run or notebook session would re-fetch data directly from GCS.

The fix pops simplecache out of kwargs before building the nested storage_options dict, then re-adds it at the top level just before calling url_to_fs().

### Changes:

In malariagen_data/util.py, pop simplecache from kwargs at the start of _init_filesystem()
Re-add it to storage_options at the top level before calling url_to_fs()


blankirigaya:
Hi @jonbrenas ,

Just following up on this PR when you have a moment. I’d appreciate it if you could take a look and let me know if any changes are needed from my side.

Thanks for your time!

PR #936, opened 2026-02-24
Unconstrained p_variance causing empty dataset crash + exon reversal typo

blankirigaya:
## Two bugs were identified via codebase review and dry-run scans:

NaN frequency crash — In simulate_cnv_hmm() and simulate_cnv_discordant_read_calls(), p_variance was set using np.random.random(), which samples uniformly from [0.0, 1.0]. When this value landed near 1.0, nearly all samples were flagged as high-variance and excluded from frequency computation. With no valid samples remaining, every gene received a NaN frequency. Any subsequent call using variant_query (e.g. "max_af > 0") would silently filter out all rows — since NaN > 0 is False in pandas — leaving an empty dataset and causing plot_frequencies_interactive_map() to crash with ValueError: No variants available in dataset. 

This only surfaced when variant_query was not None; the None path returned the NaN dataset silently and appeared to pass.
Exon strand reversal typo — In Gff3Simulator.simulate_exons(), the line exons == exons[::-1] used == (comparison) instead of = (assignment). The list was never actually reversed, causing incorrect 5'/3' UTR annotations on all minus-strand genes.

## Changes

simulate_cnv_hmm() (~line 406): replaced np.random.random() with 0.1
simulate_cnv_discordant_read_calls() (~line 509): replaced np.random.random() with 0.1
Gff3Simulator.simulate_exons() (~line 205): fixed exons == exons[::-1] → exons = exons[::-1]

# Why 0.1

Capping p_variance at a fixed 0.1 ensures ~90% of samples remain valid, frequency computation always succeeds, and tests are deterministic and stable. It also reflects realistic high-variance rates in CNV data without eliminating coverage of the exclusion path entirely.

## Testing

Dry-run scans confirmed the crash is reproducible with the old code when variant_query is not None
With p_variance = 0.1, frequency computation succeeds consistently and filtered queries return valid results


jonbrenas (approved):
LGTM

PR #946, opened 2026-02-25
Unconstrained p_variance causing empty dataset crash + exon reversal typo

### LINKED ISSUE #919: Simulated CNV HMM data fails 'variant_query' during tests
When, during the validation tests, a 'variant_query' is provided to `gene_cnv_frequencies_advanced`, the returned Dataset is always empty (to be more precise, there are no variants). This is probably a bug in `contest.py` where the simulated CNV HMM data is generated but I have not identified what the exact problem is. 

The function works on real data so the impact of this bug is limited.

blankirigaya:
What changed and why:- 

### Three targeted fixes across one file:

- simulate_cnv_hmm() (~line 406) — replaced np.random.random() with 0.1
- simulate_cnv_discordant_read_calls() (~line 509) — replaced np.random.random() with 0.1
- Gff3Simulator.simulate_exons() (~line 205) — fixed exons == exons[::-1] → exons = exons[::-1]

## Bug 1 — NaN frequency crash (primary)
 p_variance = np.random.random() sampled uniformly from [0.0, 1.0]. When the value landed near 1.0, nearly all samples were flagged as high-variance and excluded from frequency computation. With no valid samples left, every gene received a NaN frequency. Any call using variant_query (e.g. "max_af > 0") then silently dropped all rows — since NaN > 0 is False in pandas — producing an empty dataset and crashing plot_frequencies_interactive_map() with:
ValueError: No variants available in dataset

This only surfaced when variant_query was not None. The None path returned the NaN dataset silently and appeared to pass, masking the issue.

Fixing p_variance = 0.1 ensures ~90% of samples remain valid, frequency computation always succeeds, and tests are fully deterministic. It also reflects realistic CNV high-variance rates without eliminating coverage of the exclusion path.

## Bug 2 — Exon strand reversal typo (secondary)

exons == exons[::-1] used == (comparison) instead of = (assignment). The list was never reversed, silently producing incorrect 5'/3' UTR annotations on all minus-strand genes. Unrelated to the crash but a correctness fix.

closes #919


jonbrenas (approved):
LGTM

PR #948, opened 2026-02-25
Fix Fst Z-score calculation when standard error is zero

### LINKED ISSUE #947: Fst Z-score heatmap may display infinite values when SE = 0 (dead ZeroDivisionError handler)
### Summary
While reviewing the Fst heatmap output, I noticed that when the standard error (SE) is zero, the computed Z-score becomes `inf`. This happens because floating-point division by zero does not raise a `ZeroDivisionError`, so the current exception handler is never triggered.

As a result, infinite values may appear in the heatmap instead of being treated as missing values.

### Current behavior
If `se == 0.0`:

- `fst / se` returns `inf`
- the exception block is never executed
- the heatmap stores infinite values
- these values appear as valid results in the visualization

### Why this matters
This can produce misleading visual output, especially in low-variation regions or small cohorts where the standard error may be zero.

### Suggested fix
Replace the exception handling with an explicit check:

```python
if se == 0:
    fig_df.loc[cohort1, cohort2] = np.nan
else:
    fig_df.loc[cohort1, cohort2] = fst / se

adilraza99:
## Summary
Fix handling of zero standard error values when computing Fst Z-scores in the heatmap.

## Background
The previous implementation attempted to catch a `ZeroDivisionError` when computing:

    zs = fst / se

In Python, floating-point division by zero does not raise an exception — it returns `inf`.  
Because of this, the exception handler was never triggered and infinite values could be written to the heatmap.

## Impact
When `se == 0`, the heatmap could display infinite Z-scores.  
These appear as valid numeric values and may lead to misleading interpretation of results.

## Change
Replace the unused exception handler with an explicit zero check:

- return `NaN` when `se == 0`
- otherwise compute the Z-score normally

This ensures undefined values are represented safely and prevents misleading visualization output.

## Verification
- Ran the Fst test suite locally; all tests passed
- Confirmed no new warnings were introduced
- Verified that behaviour remains unchanged for non-zero SE values

## Notes
This change affects only the visualization layer and preserves existing statistical calculations.

Closes #947


adilraza99:
@jonbrenas  When you have time, could you review this to confirm the zero standard error handling is statistically appropriate?

adilraza99:
@jonbrenas

PR #949, opened 2026-02-25
fix: prevent logging handler accumulation in LoggingHelper

suhr25:
## **Summary**

This PR fixes a critical issue in `LoggingHelper` where logging handlers were unintentionally accumulating on a module-level logger across repeated instantiations.

The issue originates from the use of:

```python
logger = logging.getLogger(name)  # returns singleton
logger.addHandler(handler)        # unconditional
```

Since `logging.getLogger(name)` returns a persistent singleton, each instantiation of:

* `Ag3`
* `Af1`
* `Amin1`
* `Adir1`

adds a new handler to the same logger instance without removing previous ones.

Because `log=sys.stdout` is the default, this issue is triggered in standard notebook usage, especially when cells are re-executed.

This leads to:

* Duplicate log messages (growing linearly with reruns)
* File descriptor leaks when using `FileHandler`
* Potential crashes (`OSError: [Errno 24] Too many open files`) in long-running sessions

---

## **Fix**

Before adding a new handler, existing handlers on the logger are removed and closed:

```python
for existing_handler in logger.handlers[:]:
    logger.removeHandler(existing_handler)
    existing_handler.close()
```

Then the new handler is attached as intended:

```python
logger.addHandler(handler)
```

### Why this approach

* Ensures the logger remains **idempotent across instantiations**
* Properly releases resources (especially file descriptors)
* Keeps behavior unchanged for first-time initialization
* Minimal and localized change (no API impact)

---

### Added regression test

A unit test verifies that handlers do not accumulate:

```python
def test_no_handler_accumulation_on_repeated_instantiation():
    import logging
    import io

    for _ in range(5):
        LoggingHelper(name="test", out=io.StringIO())

    logger = logging.getLogger("test")
    assert len(logger.handlers) <= 1
```

### Manual verification

* Repeated instantiation in a notebook no longer produces duplicate log output
* `FileHandler` no longer leaks file descriptors across runs
* Logger state remains stable across multiple object creations

---

## **Impact**

* Prevents duplicate logging output in interactive workflows
* Eliminates file descriptor leaks and related crashes
* Improves stability of long-running notebook sessions
* No breaking changes or API modifications

---

## **Additional Notes**

* This fix targets only handler lifecycle management and does not alter logging configuration semantics
* Behavior remains identical for single instantiation scenarios

---

## **Checklist**

* [x] Tests pass locally (`poetry run pytest -v tests/anoph`)
* [x] Added regression test for the bug
* [x] Code passes pre-commit (`ruff` formatting/linting)
* [x] No breaking changes introduced
* [x] Follows existing project structure and conventions

### DIFF
--- a/LINUX_SETUP.md
+++ b/LINUX_SETUP.md
@@ -1,0 +1,67 @@
+# Developer setup (Linux)
+
+To get setup for development, see [this video if you prefer VS Code](https://youtu.be/zddl3n1DCFM), or [this older video if you prefer PyCharm](https://youtu.be/QniQi-Hoo9A), and the instructions below.
+
+## 1. Fork and clone this repo
+```bash
+git clone git@github.com:[username]/malariagen-data-python.git
+cd malariagen-data-python
+```
+
+## 2. Install Python
+```bash
+sudo add-apt-repository ppa:deadsnakes/ppa
+sudo apt install python3.10 python3.10-venv
+```
+
+## 3. Install pipx and poetry
+```bash
+python3.10 -m pip install --user pipx
+python3.10 -m pipx ensurepath
+pipx install poetry
+```
+
+## 4. Create and activate development environment
+```bash
+poetry install
+poetry shell
+```
+
+## 5. Install pre-commit hooks
+```bash
+pipx install pre-commit
+pre-commit install
+```
+
+Run pre-commit checks manually:
+```bash
+pre-commit run --all-files
+```
+
+## 6. Run tests
+
+Run fast unit tests using simulated data:
+```bash
+poetry run pytest -v tests/anoph
+```
+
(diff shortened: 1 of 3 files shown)


suhr25:
Hi @jonbrenas, 
This fixes duplicate logs and FD leaks caused by handlers piling up on the shared logger during repeated runs.
Added a small cleanup and regression test, would really appreciate your review.
Thanks!

jonbrenas (approved):
LGTM

suhr25:
Hi @jonbrenas,
Thanks for the approval! It looks like CI hasn’t started after the latest commit, could you please approve the workflows so the checks can run?

suhr25:
Hi @jonbrenas,
Just a follow up comment. 
Please take a look at it. Thanks !!

Issue #939, opened 2026-02-24
Install package dependencies with poetry in CI workflows


jwgarber:
Many of the CI workflows are currently installing dependencies using pip, which doesn't track the exact version numbers of the packages and can cause them to get out of sync with what's in poetry.lock. E.g. some jobs were using incorrectly using numpy 2 and had to be pinned to use an earlier version, see https://github.com/malariagen/malariagen-data-python/pull/855

The best solution to this is to install all packages using `poetry install --extras dev` and bypass pip entirely. This will ensure we have a reproducible baseline for debugging CI failures that matches a local development environment.

Definition of done:
- All workflows use poetry to install dependencies
- No more manual invocations of pip

Issue #835, opened 2025-12-01
Segmentation fault during CI checks


leehart:
E.g.
```
/home/runner/work/_temp/b19ab3ad-1112-4134-800b-d010aa6c715e.sh: line 1:  2267 Segmentation fault      (core dumped) pytest -v tests --ignore tests/integration --typeguard-packages=malariagen_data,malariagen_data.anoph
tests/anoph/test_sample_metadata.py::test_wgs_data_catalog[adir1_sim] 
Error: Process completed with exit code 139.
```

leehart:
The `test_wgs_data_catalog` function isn't particularly complicated:
```py
@parametrize_with_cases("fixture,api", cases=".")
def test_wgs_data_catalog(fixture, api):
    # Set up test.
    df_sample_sets = api.sample_sets().set_index("sample_set")
    sample_count = df_sample_sets["sample_count"]
    all_sample_sets = df_sample_sets.index.to_list()
    # sample_set = random.choice(all_sample_sets)

    for sample_set in all_sample_sets:
        # Call function to be tested.
        df = api.wgs_data_catalog(sample_set=sample_set)

        # Check output.
        assert isinstance(df, pd.DataFrame)
        expected_cols = [
            "sample_id",
            "alignments_bam",
            "snp_genotypes_vcf",
            "snp_genotypes_zarr",
        ]
        assert df.columns.to_list() == expected_cols
        assert len(df) == sample_count.loc[sample_set]

        # Compare with sample metadata.
        df_samples = api.sample_metadata(sample_sets=sample_set)
        # Don't enforce same order, but require same set.
        assert set(df["sample_id"]) == set(df_samples["sample_id"])
```

For each sample set, the function simply appears to try to get the `wgs_data_catalog` file into a DataFrame, check the columns names and the sample count (self-consistency), and then check the set of sample ids against the sample metadata.

I wonder if there is a problem with one or more of the sample sets' `wgs_data_catalog` files, except this set of tests always runs fine locally. Also, we wouldn't expect a segfault error if there was simply a missing file... 🤔 

Possibly unrelated, but I notice that the local tests report missing surveillance flags data for sets 1277-VO-KH-WITKOWSKI-VMF00151, 1276-AD-BD-ALAM-VMF00156 and 1175-VO-KH-STLAURENT.

jonbrenas:
Great investigative work, @leehart !

[...]

PR #958, opened 2026-02-26
Add github action for installing python and poetry consistently

jwgarber:
Some of the workflows were installing the dependencies using pip, which always grabs the latest allowed package version. This was causing subtle failures in some cases since the new versions may not be compatible (e.g. numpy 2 in https://github.com/malariagen/malariagen-data-python/issues/847) or have bugs (e.g. pandas in https://github.com/malariagen/malariagen-data-python/issues/835). This adds an action so all the workflows setup python in the same way, and install packages with poetry (which will follow the lock file and match local development).

Fixes https://github.com/malariagen/malariagen-data-python/issues/939

### DIFF
--- a/malariagen_data/anoph/cnv_data.py
+++ b/malariagen_data/anoph/cnv_data.py
@@ -642,9 +642,13 @@ class AnophelesCnvData(
 
                 ly.append(y)
 
-                if len(ly) == 0:
-                    # Bail out, no data for given sample sets and analysis.
-                    raise ValueError("No data found for requested sample sets.")
+            if len(ly) == 0:
+                # Bail out, no data for given sample sets and contig.
+                raise ValueError(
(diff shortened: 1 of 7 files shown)


jwgarber:
Hmm the [failure](https://github.com/malariagen/malariagen-data-python/actions/runs/22472053182/job/65090921923) in the coverage task seems to a flaky test, the other ones are passing and I haven't seen this one before. I have an idea about logging out the random seeds of the generated data so that these are easier to debug ... I might tackle that next week. In the mean time I rebased my PR to re-trigger the checks, hopefully it passes this time.

jwgarber:
The last pending check "tests (3.10, numpy==1.26.4)" is from the old testing matrix and seems to be stuck, you may have to update the branch protection rules to the new test names: https://github.com/orgs/community/discussions/51499

PR #963, opened 2026-02-26
Allow G123 to run without phasing analysis when no default phasing exists

SKM2227229725:
Fixes #795

## Summary

Previously, `g123_gwss()` required a default phasing analysis ID,
which caused an assertion error for datasets such as `amin1` and `adir1`
where no phasing analysis is available.

## What this PR changes

- Allows `sites=None` when no default phasing analysis exists
- Adds `None` to valid site options
- Maintains backward compatibility for datasets that do use phasing
- Avoids the need for placeholder strings like "noneyet"

## Rationale

G123 uses unphased genotype data for the main calculation.
Requiring a phasing analysis ID should not be mandatory for
datasets that do not provide phasing data.

This change enables G123 to run cleanly on such datasets
without breaking existing behaviour.


jonbrenas (changes requested):

jonbrenas:
Obsolete

PR #964, opened 2026-02-26
Add dataset-return options for SNP counts/diplotypes and remove duplicates

### LINKED ISSUE #472: Use xarray datasets for more return values
Currently we have a few situations where code could be simplified by returning xarray datasets instead of arrays or tuples of arrays.

A particular case motivating this issue is functions which require SNP allele counts. E.g., when computing biallelic SNP calls, we first need to compute allele counts, then use the counts to select sites. This results in two calls to `snp_calls()` under the hood, the first to get the allele counts, the second to then build the output dataset.

It would be simpler if we modified the `snp_allele_counts()` function to return a SNP calls dataset with the allele counts added as an additional variable, rather than just the allele counts array. This would avoid multiple calls to `snp_calls()` internally.

Similarly, `biallelic_diplotypes()` could return a SNP calls dataset with the diplotype array added as a variable. This would avoid having to return a tuple of diplotypes and sample identifiers, because the sample identifiers if needed could be accessed from the dataset.

Other cases where it also might be better to return a dataset include: `biallelic_snp_pairwise_distances()`, `haplotype_pairwise_distances()`.

31puneet:
Fixes #472 
- I have Added `return_dataset=True` to snp_allele_counts(), biallelic_diplotypes(), biallelic_diplotype_pairwise_distances(), and haplotype_pairwise_distances().
- Cached variant metadata alongside allele counts so `snp_allele_counts(return_dataset=True)` can build the Dataset from cache without a redundant snp_calls().
- Refactored `biallelic_snp_calls` to call snp_calls() and snp_allele_counts() separately
- Added tests for all new dataset-return paths and cleaned up duplicate assertions in `test_distance.py.`

### DIFF
--- a/tests/anoph/test_distance.py
+++ b/tests/anoph/test_distance.py
@@ -124,17 +124,6 @@ def check_biallelic_diplotype_pairwise_distance(*, api, data_params, metric):
     assert n_snps_used >= n_snps
     assert n_snps_used <= n_snps_available
 
-    assert isinstance(dist, np.ndarray)
-    assert isinstance(samples, np.ndarray)
-    assert isinstance(n_snps_used, int)
-
-    assert dist.ndim == 1  # condensed form distance matrix
-    assert dist.shape[0] == int((n_samples * (n_samples - 1)) / 2)
-    assert samples.ndim == 1
-    assert samples.shape[0] == n_samples
-    assert n_snps_used >= n_snps
-    assert n_snps_used <= n_snps_available
-
     ds = api.biallelic_diplotype_pairwise_distances(
         n_snps=n_snps,
         metric=metric,
--- a/tests/anoph/test_distance.py
+++ b/tests/anoph/test_distance.py
@@ -124,18 +124,29 @@ def check_biallelic_diplotype_pairwise_distance(*, api, data_params, metric):
     assert n_snps_used >= n_snps
     assert n_snps_used <= n_snps_available
 
-    # Check types.
     assert isinstance(dist, np.ndarray)
     assert isinstance(samples, np.ndarray)
     assert isinstance(n_snps_used, int)
 
-    # Check sizes.
     assert dist.ndim == 1  # condensed form distance matrix
     assert dist.shape[0] == int((n_samples * (n_samples - 1)) / 2)
     assert samples.ndim == 1
     assert samples.shape[0] == n_samples
     assert n_snps_used >= n_snps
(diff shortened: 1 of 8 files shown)


jonbrenas:
Thanks @31puneet. 

> I can update _snp_allele_counts to also cache the variant metadata so the Dataset can be built directly from cache without a second call.

Do you think that's what needs to be done? You chose not to do it before, did you have a reason? If you do it now, what's your reasoning? What has changed?

31puneet:
Hi @jonbrenas PTAL
I have made some changes since your last review:
- Removed extra snp_calls().
- Added tests for haplotype_pairwise_distances(return_dataset=True) and removed duplicate assertions.

jonbrenas (approved):

tests/anoph/test_distance.py
```diff
@@ -124,18 +124,19 @@ def check_biallelic_diplotype_pairwise_distance(*, api, data_params, metric):
...
-    assert isinstance(dist, np.ndarray)
-    assert isinstance(samples, np.ndarray)
-    assert isinstance(n_snps_used, int)
+    ds = api.biallelic_diplotype_pairwise_distances(
+        n_snps=n_snps,
+        metric=metric,
+        return_dataset=True,
+        **data_params,
+    )
+    import xarray as xr
```
ahernank, line 133: @jonbrenas could I double check why this import is inside the function rather than defined at the top?

tests/anoph/test_hapclust.py
```diff
@@ -124,3 +124,37 @@ def test_plot_haplotype_sharing_chord(fixture, api: AnophelesHapClustAnalysis):
             show=False,
         )
         assert fig is not None
+
+
+@parametrize_with_cases("fixture,api", cases=".")
+def test_haplotype_pairwise_distances(fixture, api: AnophelesHapClustAnalysis):
+    import xarray as xr
```
ahernank, line 131: @jonbrenas same as above, could I double check why this import is inside the function rather than defined at the top?

ahernank (commented):
@jonbrenas could I get your eye here please?

tests/anoph/test_distance.py
jonbrenas: Thanks, @ahernank. I believe the idea is that it saves some time as the default behaviour doesn't require `xarray` or `squareform`. I am not sure it is worth the loss in readability, though.

PR #972, opened 2026-02-27
refactor: replace manual cache pattern with cached_property for _snp_effect_annotator (#469)

### LINKED ISSUE #469: Simplify memory caching
Currently we manually manage in-memory caches for many methods on the Ag3 and Af1 APIs, where results are small but may take noticeable time to compute. 

This could potentially be simplified by using the [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) and [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) features of the `functools` standard library module. See also [how do I cache method calls](https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls).

Tanisha127:
Closes #469

## Summary
Replaces the manual None-check caching pattern with functools.cached_property 
for the _snp_effect_annotator private method in AnophelesSnpFrequencyAnalysis.

## Changes
- malariagen_data/anoph/snp_frq.py: converted _snp_effect_annotator from a 
  manual cache (self._cache_annotator = None in __init__ + None-check in method) 
  to @cached_property, removing boilerplate and making the caching intent explicit.

## What was NOT changed
After investigation, open_genome (in both genome_sequence.py and plasmodium.py), 
sample_metadata, and _open_variant_calls_zarr were kept as manual cache patterns 
because they are either part of the public API (called with () by tests and users) 
or are patched by mocks in the test suite — both incompatible with @cached_property.

## Tests
All existing tests pass with no modifications:
- tests/anoph/test_genome_sequence.py — 11 passed
- tests/anoph/test_snp_frq.py — 123 passed
- tests/test_plasmodium.py — 27 passed
Total: 161 passed

### DIFF
--- a/malariagen_data/anoph/dipclust.py
+++ b/malariagen_data/anoph/dipclust.py
@@ -13,7 +13,7 @@ from ..util import (
     _multiallelic_diplotype_mean_sqeuclidean,
     _multiallelic_diplotype_mean_cityblock,
 )
-from ..plotly_dendrogram import _plot_dendrogram
+from ..plotly_dendrogram import _plot_dendrogram, concat_clustering_subplots
 from . import (
     base_params,
     plotly_params,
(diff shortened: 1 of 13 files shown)


Tanisha127:
Hi @jonbrenas 
Those docstrings were accidentally removed during the refactor. I've added them back now. Please let me know if there's anything else to address. Thanks!

PR #974, opened 2026-02-27
fix: raise informative error when no cohorts match selected taxon in plot_frequencies_time_series

### LINKED ISSUE #671: Occasional test failure in test_plot_frequencies_time_series_with_taxa
> FAILED tests/anoph/test_snp_frq.py::test_plot_frequencies_time_series_with_taxa[ag3_sim] - ValueError: No objects to concatenate

I suspect this is happening when random taxa are selected for testing, but somehow results in an empty dataset:
```python
    # Pick a random taxon and taxa from valid taxa.
    sample_sets_taxa = (
        api.sample_metadata(sample_sets=sample_sets)["taxon"].dropna().unique().tolist()
    )
    taxon = random.choice(sample_sets_taxa)
    taxa = random.sample(sample_sets_taxa, random.randint(1, len(sample_sets_taxa)))
```

Tanisha127:
Fixes #671

## Problem
`plot_frequencies_time_series` was crashing with an unhelpful `ValueError: No objects to concatenate` when the selected taxon had no matching cohorts in the computed dataset. This happened because `pd.concat(dfs)` was being called with an empty list.

## Solution
Added a guard before the `pd.concat(dfs)` call that raises a clear, informative error message when no cohorts match the selected taxon, instead of letting pandas raise a confusing internal error.

## Changes
- `malariagen_data/anoph/frq_base.py`: Added check for empty `dfs` list before calling `pd.concat`


jonbrenas (changes requested):

jonbrenas:
Thanks @Tanisha127, I am still not sure why the absence of cohorts would be due exclusively to the taxon parameter. It could be that the areas are nonsensical.

Tanisha127:
Thank you for the feedback! 
I've updated the error message to mention both the taxa and areas parameters, since either could result in no matching cohorts. The message now also prints the actual values the user passed to help with debugging.

jonbrenas (approved):
LGTM

Tanisha127:
Thank you for the review @jonbrenas! 
Happy to hear it looks good. Just flagging that this PR is now approved and all 8 checks are passing — please feel free to merge when convenient.

PR #975, opened 2026-02-27
refactor: standardise biallelic diplotypes and handling of missing calls

### LINKED ISSUE #471: Standardise biallelic diplotypes and handling of missing calls
We now have a function `biallelic_diplotypes()` which computes an array of shape (variants, samples) representing genotypes at biallelic sites. 

The current implementation codes each call as an integer counting the count of the second allele. Missing calls do not receive any special coding, and so get a count of 0, which is the same value for a genotype homozygous for the first allele. This isn't ideal as it will cause a bias towards the reference allele in general, because this is most commonly the first allele. This bias can be minimised, by using the `max_missing_an` parameter to select only sites with low or zero missingness. However, in general it would be better to have missing calls represented in the data, and then explicitly dealt with in any functions that use the data.

Current functions downstream of `biallelic_diplotypes()` are `pca()` and `plot_njt()` (via `biallelic_pairwise_distances()`). Neither of these functions have any special handling of missing calls, and so would need to be adapted.

A planned function to be added in the next release is `biallelic_snp_calls_to_plink()` to write out data in plink binary format (#248). The convention for these data is to represent diplotypes as the count of the first allele (rather than the second) and to code missing calls as -127.

To allow for consistency across all functions downstream of `biallelic_diplotypes()` I propose to:

* Modify `biallelic_diplotypes()` to count the first allele, rather than the second, and code missing calls as -127.
* Modify the implementation of `biallelic_pairwise_distances()` to explicitly deal with missing calls. For each pair of diplotypes, this could be done by computing the mean distance at sites with non-missing calls in both diplotypes.
* Modify the implementation of `pca()` to add a simple imputation of missing values if `max_missing_an > 0`, e.g., mean or most common. 

This would allow a user to run `pca()`, `plot_njt()` and `biallelic_snp_calls_to_plink()` with the same parameters and know that exactly the same data had been used.

Tanisha127:
Closes #471

### `snp_data.py`
- Modified `_biallelic_diplotypes()` to count the **first (ref) allele** instead of the second (alt), using `to_n_ref()` instead of `to_n_alt()`
- Missing calls are now explicitly coded as **-127** instead of 0, preventing bias towards the reference allele
- Updated cache name from `biallelic_diplotypes` to `biallelic_diplotypes_v2` to invalidate previously cached data

### `distance.py`
- Updated `_biallelic_diplotype_pairwise_distances()` to handle missing calls (-127) by converting them to `np.nan` before computing distances
- This ensures distances are computed only at sites with non-missing calls for both samples

### `pca.py`
- Added mean imputation for missing calls (-127) in `_pca()` when `max_missing_an > 0`
- Missing values are replaced with the site mean before running PCA
- Updated cache name from `pca_v5` to `pca_v6` to invalidate previously cached data

### DIFF
--- a/malariagen_data/adir1.py
+++ b/malariagen_data/adir1.py
@@ -19,6 +19,10 @@ TAXON_COLORS = {
     "dirus": TAXON_PALETTE[0],
 }
 
+XPEHH_GWSS_CACHE_NAME = "adir1_xpehh_gwss_v1"
+IHS_GWSS_CACHE_NAME = "adir1_ihs_gwss_v1"
+ROH_HMM_CACHE_NAME = "adir1_roh_hmm_v1"
+
 
 class Adir1(AnophelesDataResource):
     """Provides access to data from Adir1.0 releases.
@@ -70,6 +74,10 @@ class Adir1(AnophelesDataResource):
         >>> adir1 = malariagen_data.Adir1(results_cache="results_cache")
 
     """
+
+    _xpehh_gwss_cache_name = XPEHH_GWSS_CACHE_NAME
+    _ihs_gwss_cache_name = IHS_GWSS_CACHE_NAME
+    _roh_hmm_cache_name = ROH_HMM_CACHE_NAME
 
     def __init__(
         self,
--- a/malariagen_data/af1.py
+++ b/malariagen_data/af1.py
@@ -21,6 +21,10 @@ TAXON_COLORS = {
     "funestus": TAXON_PALETTE[0],
 }
 
+XPEHH_GWSS_CACHE_NAME = "af1_xpehh_gwss_v1"
+IHS_GWSS_CACHE_NAME = "af1_ihs_gwss_v1"
+ROH_HMM_CACHE_NAME = "af1_roh_hmm_v1"
+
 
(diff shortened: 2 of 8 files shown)


Tanisha127:
Hi @jonbrenas, I've updated _pca() to use the imputation_method parameter. "mean" is the default (preserving existing behaviour), "zero" is also supported, and a ValueError is raised for unknown values. All CI checks passing. Please let me know if anything else needs changing!

jonbrenas (changes requested):
Thanks @Tanisha127. Could you add the most common imputation as an option? It think, after all, that it should be the default option too (or zero should be) as a float genotype doesn't make biological sense. It doesn't matter as it is purely for the sake of the PCA algorithm that the value is generated but still.

It would also be great if imputation method was added as a parameter to `pca_params.py` to make things slightly more consistent.

Tanisha127:
Hi @jonbrenas, I've added imputation_method to pca_params.py following the same TypeAlias + Annotated pattern as other parameters, with "most_common" as the default. Cache bumped to pca_v8. Please let me know if anything else needs changing!

PR #980, opened 2026-02-28
Fix: remove stray `f` in ValueError messages

suhr25:
### **SUMMARY**

This PR fixes a small formatting issue in error messages where a stray `f` was being shown to users. The issue affects validation in `snp_data.py` and `hap_data.py`, making the output look slightly broken.

---

### FIX / CHANGES

* Root cause

  * A typo (`f{...}` inside an f-string) caused a literal `f` to appear in the final message.

* Approach

  * Removed the extra `f` so the string renders correctly.
  * No functional or behavioral changes.

**Before**

```python
f"Invalid site mask, must be one of f{self.site_mask_ids}."
```

After

```python
f"Invalid site mask, must be one of {self.site_mask_ids}."
```
---

### VERIFICATION

**Steps**

* Call:

  ```python
  ag3.snp_sites(region="3L", site_mask="invalid")
  ```
* Observe the error message.

**Expected result**

* Error message cleanly lists valid options without the extra `f`.

### DIFF
--- a/malariagen_data/anoph/cnv_frq.py
+++ b/malariagen_data/anoph/cnv_frq.py
@@ -446,6 +446,7 @@ class AnophelesCnvFrequencyAnalysis(AnophelesCnvData, AnophelesFrequencyAnalysis
         chunks: base_params.chunks = base_params.native_chunks,
         inline_array: base_params.inline_array = base_params.inline_array_default,
         taxon_by: frq_params.taxon_by = frq_params.taxon_by_default,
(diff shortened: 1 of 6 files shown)


suhr25:
@jonbrenas,
Could you please review this when you have time?
Thanks!

suhr25:
@jonbrenas 
Thanks for approval!
Is it ready to merge or needs some changes , do let me know.

PR #981, opened 2026-02-28
fix: enforce bool dtype on pca_fit column in pca()

blankirigaya:
## Summary

Fixes incorrect `object` dtype on the `pca_fit` column returned by `pca()`.

## Root Cause

`loc_keep_fit` is a numpy `bool_` array produced in `_pca()`. After the
`.join()` call that merges sample metadata into `df_pca`, assigning this
array to the joined DataFrame can silently produce `object` dtype if pandas
encounters any index misalignment during the assignment.

## Change

Single one-word fix in `malariagen_data/anoph/pca.py`:
```python
# Before
df_pca["pca_fit"] = loc_keep_fit

# After
df_pca["pca_fit"] = loc_keep_fit.astype(bool)
```

## Files Changed

- `malariagen_data/anoph/pca.py` — enforce `bool` dtype on `pca_fit`

## Testing

The fix can be verified without GCS access:
```python
import numpy as np
import pandas as pd

# Simulate the post-join assignment that causes the bug
loc_keep_fit = np.ones(5, dtype=bool)
df = pd.DataFrame({"PC1": range(5)})
df = df.join(pd.DataFrame({"taxon": list("abcde")}, index=range(5)))

# Before fix — dtype may become object under certain index conditions
df["pca_fit"] = loc_keep_fit
print(df["pca_fit"].dtype)  # may print: object

# After fix — always bool
df["pca_fit"] = loc_keep_fit.astype(bool)
assert df["pca_fit"].dtype == bool  # always passes
```


jonbrenas:
Obsolete

PR #984, opened 2026-02-28
Gh542 separate fst plots

### LINKED ISSUE #542: Pairwise Fst plots - Only plotting one information on each plot
Currently, pairwise Fst plots give the option to have the standard error (or the Z score) in the upper-right half. The problem is that with only one colorbar, it ends up with one of the two statistics being uniform in colour which looks bad. It is also much easier to read a whole row (or column) than having to turn halfway through. The idea is thus to have several plots instead of just one with one statistic in each.

ritoban23:
## Summary

Replaces the `annotation: Optional[Literal["standard error", "Z score"]]` parameter in `plot_pairwise_average_fst` with two boolean flags:

- `show_se` (default `False`) - adds a second heatmap panel for pairwise standard errors
- `show_zscore` (default `False`) - adds a third heatmap panel for pairwise Z scores

Each enabled panel is stacked below the FST panel with its own independent colour scale and colorbar. Column labels appear only on the bottom panel to avoid repetition.

Closes #542.

fst_and_se

## Changes

- `fst_params.py`: removed `annotation`, added `show_se` and `show_zscore` type aliases
- `fst.py`: rewrote `plot_pairwise_average_fst` using `go.Heatmap` + `make_subplots` for independent per-panel colorbars
- `test_fst.py`: updated tests to cover all four combinations of `show_se`/`show_zscore`

## Breaking change

The `annotation` parameter is removed. Replace `annotation="standard error"` with `show_se=True` and `annotation="Z score"` with `show_zscore=True`.

### DIFF
--- a/malariagen_data/anoph/fst.py
+++ b/malariagen_data/anoph/fst.py
@@ -1,4 +1,4 @@
-from typing import Tuple, Optional
+from typing import Tuple, Optional, List
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


jonbrenas:
Obsolete

PR #990, opened 2026-02-28
Fix handling of missing terms-of-use metadata in sample set manifest (#766)

### LINKED ISSUE #766: `sample_metadata` and `lookup_terms_of_use_info ` does not work when there is no terms-of-use info

adilraza99:
## Summary

Fix handling of sample set manifests that do not include terms-of-use metadata.

Some pre-release and legacy manifests legitimately omit the following fields:

- `terms_of_use_expiry_date`  
- `terms_of_use_url`  
- `unrestricted_use`  

The current implementation assumes these columns are always present, which results in a **KeyError** and prevents `sample_metadata()` from loading.

---

## Changes

- guard against missing `unrestricted_use` in `_sample_set_has_unrestricted_use()`
- make `lookup_terms_of_use_info()` tolerant of missing columns
- populate missing fields with null values to keep output consistent
- add tests covering manifests without terms-of-use metadata

---

## Why this change

Pre-release datasets may not yet have terms-of-use information.  
Handling this case explicitly allows metadata to load successfully while making it clear that usage status is unknown.

---

## Additional context

This update keeps behaviour unchanged for public releases and ensures
metadata outputs remain consistent across different release stages.

---
fixes #766

### DIFF
--- a/malariagen_data/anoph/h12.py
+++ b/malariagen_data/anoph/h12.py
@@ -625,7 +625,7 @@ class AnophelesH12Analysis(
         )
 
         # Plot H12.
-        for i, (cohort_label, (x, h12, contig)) in enumerate(res.items()):
+        for i, (cohort_label, (x, h12, contig_idx)) in enumerate(res.items()):
             fig.scatter(
                 x=x,
                 y=h12,
--- a/tests/anoph/conftest.py
+++ b/tests/anoph/conftest.py
@@ -1,8 +1,9 @@
+import hashlib
 import json
+import secrets
 import shutil
 import string
 from pathlib import Path
-from random import choice, choices, randint
 from typing import Any, Dict, Tuple
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


adilraza99:
Thanks for clarifying the intended behaviour.

This change allows metadata loading to succeed when terms-of-use columns are absent, while preserving existing behaviour when they are present.

The implementation is minimal and includes tests for the missing-column scenario.

Happy to refine if needed.

@jonbrenas

adilraza99:
All checks have passed.

@jonbrenas please take a look when you have time.
Let me know if anything should be adjusted.

adilraza99:
@jonbrenas
That explanation cleared things up.

I’ve updated the implementation so that missing terms-of-use columns now raise a clear error instead of being silently filled. This ensures incomplete metadata is surfaced as a data integrity issue while preserving traceability and keeping the API behaviour explicit.

Tests have been updated to reflect the strict validation behaviour.

Please let me know if this aligns with the intended design, or if you’d like any adjustments.

PR #994, opened 2026-03-01
Fix contig variable shadowing in H12 overlay plot

### LINKED ISSUE #993: Incorrect x-axis label in plot_h12_gwss_multi_overlay_track() due to variable shadowing
While reviewing the H12 GWSS overlay plotting function, I noticed that `plot_h12_gwss_multi_overlay_track()` reuses the name `contig` when
unpacking the results returned by `h12_gwss()`.

Inside the plotting loop:

    for i, (cohort_label, (x, h12, contig)) in enumerate(res.items()):

the loop variable `contig` represents an array of contig indices for each window. This shadows the function parameter `contig`, which is expected to remain a string identifier (e.g. "2L").

After the loop completes, the shadowed value is passed to:

    self._bokeh_style_genome_xaxis(fig, contig)

As a result, the x-axis label can render as:

    Contig [0 0 0 …] position (bp)

instead of the expected:

    Contig 2L position (bp)

The plotted data remain correct, but the axis label becomes misleading.

Suggested fix:

Rename the loop variable to avoid shadowing, for example:

    for i, (cohort_label, (x, h12, contig_idx)) in enumerate(res.items()):

This preserves the original `contig` parameter for axis labeling and aligns with naming used in related plotting functions.

adilraza99:
## Summary

This fixes an issue in `plot_h12_gwss_multi_overlay_track()` where the loop variable `contig` unintentionally shadows the function parameter of the same name.

The value unpacked from `h12_gwss()` is a NumPy array of contig indices. Because Python does not create a new scope inside loops, this overwrote the original contig string (e.g. `"2L"`). After the loop, the shadowed value was passed to `_bokeh_style_genome_xaxis()`, which could produce an incorrect axis label such as:

Contig [0 0 0 …] position (bp)

instead of:

Contig 2L position (bp)

The plotted data remain correct; only the axis label is affected.

## Fix

Rename the loop variable to avoid shadowing and preserve the original parameter value used for axis labeling.

## Notes

- This change only affects axis labeling and does not modify plotted data.
- The update is limited to a single variable rename with no behavioural impact.
- Existing functionality and outputs remain unchanged.

Fixes #993

### DIFF
--- a/malariagen_data/anoph/snp_data.py
+++ b/malariagen_data/anoph/snp_data.py
@@ -1872,9 +1872,10 @@ class AnophelesSnpData(
 
                 # Apply missingness condition.
                 if max_missing_an is not None:
-                    an_missing = (ds_out.sizes["samples"] * ds_out.sizes["ploidy"]) - an
+                    an_total = ds_out.sizes["samples"] * ds_out.sizes["ploidy"]
+                    an_missing = an_total - an
                     if isinstance(max_missing_an, float):
-                        an_missing_frac = an_missing / an
+                        an_missing_frac = an_missing / an_total
                         loc_missing = an_missing_frac <= max_missing_an
                     else:
                         loc_missing = an_missing <= max_missing_an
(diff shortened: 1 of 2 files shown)


adilraza99:
@jonbrenas could you please review this when you have time?

adilraza99:
All checks are passing and the patch coverage is 100%.

The overall coverage shows a small drop due to baseline comparison, but this change does not alter functionality.

Please let me know if you would like a test added.

adilraza99:
Hi @jonbrenas, CI checks are passing now.  
Could you please take a look when you have a moment?

Issue #938, opened 2026-02-24
Crash when running test_njt_with_algorithm


jwgarber:
Running the unit tests locally showed one of them was failing.

```
FAILED tests/anoph/test_distance.py::test_njt_with_algorithm[ag3_sim] - IndexError: index is out of bounds
debug: IndexError: index -1 is out of bounds for axis 0 with size 57
```

Unfortunately this only happened sporadically and was difficult to reproduce. Originally I suspected that the different randomly generated data was the cause of the non-reproducibility (set `np.random.seed()` with 100 different values and generate the simulated data with a different seed each time), but it turned out that even with the same random seed and simulated data the test would pass most of the time and only fail occasionally.

The exception itself was being thrown inside `anjl.dynamic_nj()`.

```
    def dynamic_nj(
        D: params.D,
        disallow_negative_distances: params.disallow_negative_distances = True,
        progress: params.progress = None,
        progress_options: params.progress_options = {},
        copy: params.copy = True,
        parallel: params.parallel = True,
    ) -> params.Z

       ...
    
        # Begin iterating.
        for iteration in iterator:
            # Perform one iteration of the neighbour-joining algorithm.
>           z = dynamic_iteration(
                iteration=np.uintp(iteration),
                distance=distance,
                R=R,
                Q=Q,
                previous_z=z,
                index_to_id=index_to_id,
                obsolete=obsolete,
                Z=Z,
                n_original=np.uintp(n_original),
                disallow_negative_distances=disallow_negative_distances,
                parallel=parallel,
            )
E           IndexError: index is out of bounds
```

On a lucky test run I was able to save the distance matrix and run it in a loop.

```python
import anjl
import numpy as np

D = np.load("crash.npy")

for i in range(100):
    print("Iteration", i)
    anjl.dynamic_nj(D=D)
```

This reliably crashes on some iteration. However, setting `parallel=False` makes the error go away, so this appears to be a race condition internally in anjl when it's running with multiple threads.

PR #1008, opened 2026-03-02
Make random anoph testing data reproducible

jwgarber:
The unit tests generate random anoph data, which is much faster than downloading real data from GCS. However, there have been many cases (https://github.com/malariagen/malariagen-data-python/issues/938, https://github.com/malariagen/malariagen-data-python/issues/835, and others) where the randomness makes it very difficult to reproduce flaky CI failures after the fact for local debugging. One has to "roll the dice" many times in the hopes of generating the same data that caused the failure, and then try to capture enough of it to debug the problem. Painful!

The usual solution to this is to use [pytest-randomly](https://github.com/pytest-dev/pytest-randomly) to generate a global random seed, which it then sets at the start of each test using `random.seed()`. However, there are two problems:

1. This doesn't work for our session fixtures that create the data before running tests.
2. It only works with the legacy numpy.random calls, which have since been replaced by the [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) API to avoid problems with setting a global seed variable (speed, thread safety, etc.)

To solve this I've adopted the following approach inspired by pytest-randomly, but ported to use the new RNG API.

1. A global seed is generated on startup from a 128 bit random integer (as recommended by Numpy).
2. The global seed is mixed with a seed derived from a context string to create context-specific RNGs where needed (e.g. for Ag3, Af1, etc.). The context is mixed into the seed to ensure that all RNGs will generate independent random numbers, but still be reproducible from the global seed.
3. All usages of the python random library and legacy numpy.random calls are replaced with calls to the corresponding RNG.
4. The global seed is logged on startup to allow reproducibility after the fact.

To test this out I reverted the pandas pin in https://github.com/malariagen/malariagen-data-python/pull/943, and after finding an appropriate global seed I am able to trigger the segfault reliably from that seed, which shows that the failure is reproducible. After this PR is merged I will create a second one to port the random calls in the unit tests to use the new RNGs, after which all of the pipeline failures will be reproducible locally from the seed that caused the failure.

### DIFF
--- a/malariagen_data/anoph/h12.py
+++ b/malariagen_data/anoph/h12.py
@@ -625,7 +625,7 @@ class AnophelesH12Analysis(
         )
 
         # Plot H12.
-        for i, (cohort_label, (x, h12, contig)) in enumerate(res.items()):
+        for i, (cohort_label, (x, h12, contig_idx)) in enumerate(res.items()):
             fig.scatter(
                 x=x,
                 y=h12,


jwgarber:
@jonbrenas This is a follow-up to https://github.com/malariagen/malariagen-data-python/pull/958 to make debugging easier.

PR #1017, opened 2026-03-02
feat: add generic Manhattan/QQ plotting helpers for association results

### LINKED ISSUE #772: Adding Manhattan and QQ plots in the API
I was preparing to create the Manhattan and QQ plots at the end of my GWAS example notebook using `plink` and I was confronted with a choice that (in my opinion) is not trivial. Marees et al. uses R but that seems very sub-optimal in our case. At this point, I am not even sure how much of the GWAS code is going to be part of the API so this question might be moot. 

The API doesn't have currently a way to create Manhattan and QQ plots and the first question is probably "Are there other use cases than GWAS that would justify adding the functionality to `malariagen_data`?". The answer might be different for the Manhattan plots and the QQ plots.

The next question is "If we were to add the functionality to create Manhattan/QQ plots, how would we do it?". We already have some nifty plots with a stats track and a gene track (e.g., H12) that are Manhattan plot adjacent so that would probably be the way I would go. `dash_bio` has a function that creates Manhattan plots but we it has the qualities and restrictions of using `dash`. The `sgkit` GWAS tuto  creates both plots (form an xarray) using `seaborn` which would probably be the way to go but there are also options using `matplotlib`.

Sharon-codes:
## Summary
Adds Manhattan and QQ plotting helpers to the API for association-style results, designed to work directly in Python workflows (including GWAS notebooks) without requiring R or Dash-specific tooling.

## Design choices (addressing issue discussion)
- **Not tied to GWAS only**: the functions operate on generic association result tables containing contig/position/p-values.
- **Python-native API support**: no R dependency.
- **No Dash requirement**: uses Plotly, already used in this project.
- **Supports multiple data backends**:
  - `pandas.DataFrame` (e.g. PLINK-exported tables loaded in pandas)
  - `xarray.Dataset` (aligned with sgkit-style workflows)

## Changes
- `malariagen_data/anoph/phenotypes.py`
  - Added `plot_manhattan(...)`
  - Added `plot_qq(...)`
  - Added internal normalization helper to accept both DataFrame and xarray Dataset inputs.
  - Added Manhattan controls for `contig_order`, `contig_spacing`, and threshold line.
- `tests/anoph/test_phenotypes_plotting.py`
  - Added tests for DataFrame and xarray inputs
  - Added validation test for missing columns

## Validation
- `poetry run pytest tests/anoph/test_phenotypes_plotting.py -q` (5 passed)
- `poetry run mypy malariagen_data/anoph/phenotypes.py tests/anoph/test_phenotypes_plotting.py --ignore-missing-imports`
- `poetry run ruff check malariagen_data/anoph/phenotypes.py tests/anoph/test_phenotypes_plotting.py`

Closes #772


Sharon-codes:
@jonbrenas Hey please look into this , i hope this addresses your question also which was there in the comment of the issue , if you need me to add anything do let me or else please check and allow it , thanks alot !!!

Sharon-codes:
Heyy @jonbrenas if all good , please accept it , thanks

malariagen_data/anoph/association.py
```diff
@@ -0,0 +1,296 @@
...
+
+    if TYPE_CHECKING:
+        phenotypes_with_snps: Callable[..., xr.Dataset]
+
+    @_check_types
+    def association_results(
+        self,
+        data: xr.Dataset,
+        *,
+        genotype_col: str = "call_genotype",
```
jonbrenas, line 25: Quite a few of these should be defined as params in a separate file for consistency, maybe base_params.field is enough.

jonbrenas (changes requested):

Issue #848, opened 2026-01-23
project pins numpy <2.2 but CI tests use numpy 2.4.1


leehart:
While investigating issue #847 it was noticed that CI tests for "3.11, numpy~=2.0" were using numpy v2.4.1, despite the `pyproject.toml` pinning the version of numpy "<2.2".

Example:
- https://github.com/malariagen/malariagen-data-python/actions/runs/21061927397/job/60574852161

The explanation appears to be a misunderstanding of what `numpy~=2.0` means. For instance, in the `tests.yml` workflow we currently have:
```yaml
      matrix:
        python-version: ["3.10", "3.11", "3.12"]
        numpy-version:
          - "numpy==1.26.4"  # current colab version
          - "numpy~=2.0"  # 2.0 series
          # numba (0.60.0) does not yet support numpy 2.1, disable this for now
          # - "numpy~=2.1"  # 2.1 series
[...]
      - name: Setup python
        uses: actions/setup-python@v5
        with:
          python-version: ${{ matrix.python-version }}
          cache: "pip"

      - name: Install package
        run: pip install "${{ matrix.numpy-version }}" .[dev]
```

This seems to intend for tests to run for Python versions `["3.10", "3.11", "3.12"]` and the numpy versions `1.26.4` (for Colab) and the "2.0 series" of numpy versions, but `numpy~=2.0` actually allows any version from 2.0.0 to 2.9.9, which obviously includes the v2.4.1 that is evidently being tested in practice.

if the test matrix is amended to test `"numpy>=2.0.0,<2.1.0" ` instead, there would still the contradiction with the `pyproject.toml` pinning numpy to "<2.2", so I expect the most consistent amend would be to change the testing matrix to "numpy>=2.0.0,<2.2.0".

The comment in the `tests.yml` says that "numba (0.60.0) does not yet support numpy 2.1, disable this for now", referring to the commented-out line for `"numpy~=2.1"`. However, `"numpy~=2.0"` already includes numpy 2.1., as well as anything up to v3.0. Additionally, the `pyproject.toml` currently pins numba to ">=0.60.0". So that comment doesn't seem to apply and should probably be removed.

Additionally, the current version of numpy in Colab appears to be `2.0.2`, so the current testing of `1.26.4` appears outdated, and we can probably remove that from the test matrix. (By the way, Colab appears to currently use Python 3.12.12.)

leehart:
[...]

PR #1023, opened 2026-03-02
ci: verify tests against locked and latest supported NumPy versions

Sharon-codes:
## Summary
Addresses #848 by making NumPy version verification explicit in CI.

## Changes
- Add `numpy-spec` to the tests matrix:
  - `==2.0.2` (locked baseline)
  - `>=2.0.2,<2.1` (latest supported range from `pyproject.toml`)
- Install the matrix-selected NumPy version before running tests.
- Assert the installed NumPy version satisfies the matrix spec.

## Why
This ensures CI continuously checks both:
- the known baseline version, and
- the newest compatible version within the declared support range.


jonbrenas (approved):
Thanks @Sharon-codes

PR #1028, opened 2026-03-03
Fix: raise an error when n_sites < n_jack in cohort diversity jackknife

suhr25:
**SUMMARY**

This PR fixes a missing validation check in `_block_jackknife_cohort_diversity_stats()` that caused small regions to return misleading zero width confidence intervals. It updates `malariagen_data/anopheles.py` to fail fast when there are not enough sites for the requested number of jackknife blocks.

---

**FIX**

* **Root cause:** When `n_sites < n_jack`, `block_length` becomes `0`, so every jackknife resample removes zero sites and all resamples are identical.
* **Fix:** Add a `block_length < 1` guard (matching `average_fst`) and remove a dead no-op line.
* **Scope:** Small, low-risk change; behavior only changes for invalid inputs.

**Before**

```python
block_length = n_sites // n_jack
```

**After**

```python
block_length = n_sites // n_jack
if block_length < 1:
    raise ValueError("Not enough sites for jackknife blocks.")
```

---

**VERIFICATION**

1. Call `cohort_diversity_stats()` on a small region with `n_jack=200`.
2. **Expected:** a clear `ValueError` telling the user to use a larger region or reduce `n_jack`, instead of returning zero width confidence intervals.

### DIFF
--- a/tests/anoph/test_base.py
+++ b/tests/anoph/test_base.py
@@ -1,3 +1,6 @@
+import io
+import logging
+
 import numpy as np
 import pandas as pd
 import pytest
@@ -8,6 +11,7 @@ from malariagen_data import af1 as _af1
 from malariagen_data import ag3 as _ag3
 from malariagen_data import adir1 as _adir1
 from malariagen_data.anoph.base import AnophelesBase
+from malariagen_data.util import LoggingHelper
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @jonbrenas,
This PR fixes a missing validation check in _block_jackknife_cohort_diversity_stats().
Can you please take a look at it.
Thanks!

Issue #725, opened 2025-02-25
Warnings in tests


leehart:
Just to note some warnings that appear during routine tests that should probably be addressed at some point:

```
tests/anoph/test_dipclust.py: 37 warnings
  site-packages/plotly/basedatatypes.py:2314: DeprecationWarning:
  
  The append_trace method is deprecated and will be removed in a future version.
  Please use the add_trace method with the row and col parameters.
```

```
WARNING  traitlets:interaction.py:250 Exception in interact callback:  Parameter 'variant' with value None in call to function
'plot_frequencies_map_markers' has incorrect type: found None,
expected int or str. See below for further information.

None did not match any element in the union:
  int: is not an instance of int
  str: is not an instance of str
Traceback (most recent call last):
  File "site-packages/ipywidgets/widgets/interaction.py", line 243, in update
    self.result = self.f(**self.kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^
  File "malariagen-data-python/malariagen_data/util.py", line 1163, in check_types_wrapper
    raise error from None
TypeError:  Parameter 'variant' with value None in call to function
'plot_frequencies_map_markers' has incorrect type: found None,
expected int or str. See below for further information.

None did not match any element in the union:
  int: is not an instance of int
  str: is not an instance of str
```

ZiadXI:
Hi @leehart, I'd like to work on fixing the plotly deprecation warnings from append_trace -> add_trace and scattermapbox -> scattermap. I'll leave the None type warning for a follow-up if needed

PR #1030, opened 2026-03-03
Fix: Resolve TypeError for None variant in plot_frequencies_map_markers (#725)

ArnabTechiee:
### Description
This PR addresses the lingering `TypeError` reported in **Issue #725** during the test suite runs (specifically originating from `test_dipclust.py`). 

**The Bug:**
When `plot_frequencies_interactive_map` initializes its `ipywidgets.Dropdown` for the `variant` parameter, the widget briefly holds a value of `None` before the data is fully populated. Because the `plot_frequencies_map_markers` function strictly enforced `variant: Union[int, str]`, the `typeguard` library threw a `TypeError`.

**The Fix:**
1. Updated the type hint in `plot_frequencies_map_markers` to `Optional[Union[int, str]]`.
2. Added an early return (`if variant is None: return`) right after the widget imports to safely catch the initial unpopulated state and prevent the downstream `AssertionError`.

All tests in `tests/anoph/test_dipclust.py` now pass with zero warnings.

**Resolves:** #725


jonbrenas:
Hi @ArnabTechiee,

Your solution does remove the error message but doesn't really solve the problem (which was that `plot_frequencies_map_markers` was called with 'variant == None' when it shouldn't be in tests). Can you explain in more details why you think it is the way yo go?

ArnabTechiee:
Hi @jonbrenas,

Thank you for the review, and that is a very fair point. You are absolutely right—my current fix treats the symptom (the TypeError) rather than the underlying root cause.

To answer your question on my original reasoning: Because ipywidgets (like Dropdowns) often experience brief transitional states—such as during initial rendering or when options are being asynchronously updated—the widget's internal value can temporarily default to None. I assumed the test suite (test_dipclust.py) was simply catching this transient UI initialization state. I added the early return in the core function to gracefully handle that temporary state without crashing the user's Jupyter cell.

The better approach moving forward:
After reading your feedback, I agree that relaxing the strict type constraints of the core plot_frequencies_map_markers function is the wrong place to handle this. The core backend function should strictly expect valid data.

Instead, we should catch this at the UI boundary. I propose removing my changes to the plotting function and instead putting the None check upstream inside the widget's callback handler (the observe function). If the widget temporarily shifts to None, the callback simply ignores the event and never invokes plot_frequencies_map_markers.

Does moving the logic upstream to the widget callback align better with how you'd like to handle this? If so, I will happily refactor this PR (and I will fix the linting failure that just popped up, too!).

Thanks for pushing me to find the cleaner architectural solution!

malariagen_data/anoph/frq_base.py
```diff
@@ -698,17 +699,28 @@ def plot_frequencies_interactive_map(
         taxa = ds["cohort_taxon"].to_pandas().dropna().unique()  # type: ignore
         periods = ds["cohort_period"].to_pandas().dropna().unique()  # type: ignore
 
-        if len(variants) == 0:
+if len(variants) == 0:
```
jonbrenas: This is clearly a typo.

jonbrenas (changes requested):

Issue #1029, opened 2026-03-03
Setup instructions in CONTRIBUTING.md fail with ModuleNotFoundError: pytest_cases on fresh clone


sanyam4321:
Hi team I am setting up my local environment to start contributing to the project and ran into a slight hurdle with the onboarding documentation.

Following the CONTRIBUTING.md guide, I ran poetry install and then attempted to run the recommended fast unit tests. However, the test suite instantly fails to collect, throwing 21 errors for a missing dependency.

The ERROR:
```
E   ModuleNotFoundError: No module named 'pytest_cases'
================================================== short test summary info ===================================================
ERROR tests/anoph/test_base.py
ERROR tests/anoph/test_cnv_data.py
ERROR tests/anoph/test_cnv_frq.py
ERROR tests/anoph/test_describe.py
ERROR tests/anoph/test_dipclust.py
ERROR tests/anoph/test_distance.py
ERROR tests/anoph/test_fst.py
ERROR tests/anoph/test_g123.py
ERROR tests/anoph/test_genome_features.py
ERROR tests/anoph/test_genome_sequence.py
ERROR tests/anoph/test_h12.py
ERROR tests/anoph/test_h1x.py
ERROR tests/anoph/test_hap_data.py
ERROR tests/anoph/test_hap_frq.py
ERROR tests/anoph/test_hapclust.py
ERROR tests/anoph/test_igv.py
ERROR tests/anoph/test_pca.py
ERROR tests/anoph/test_plink_converter.py
ERROR tests/anoph/test_sample_metadata.py
ERROR tests/anoph/test_snp_data.py
ERROR tests/anoph/test_snp_frq.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 21 errors during collection 
```

**Steps to Reproduce (Verified in an isolated environment):**
To ensure this wasn't just a misconfigured Python cache or a global package conflict on my own machine, I spun up a completely blank Ubuntu 22.04 Docker container, installed Python 3.10 and Poetry from scratch, and followed the exact commands in the guide. The exact same error occurred.

Here is the clean-room reproduction script:

```
docker run -it --rm ubuntu:22.04 bash

apt update && apt install -y software-properties-common git curl tzdata
add-apt-repository ppa:deadsnakes/ppa -y
apt update && apt install -y python3.10 python3.10-venv python3.10-dev pipx
export PATH="$PATH:/root/.local/bin"
[...]

PR #1031, opened 2026-03-03
fix: add pytest-cases note to setup instructions (#1029)te

sallykinyua:
Updated CONTRIBUTING.md setup instructions to fix the ModuleNotFoundError: No module named 'pytest_cases' error encountered by new contributors on a fresh clone.

New contributors following the setup instructions were hitting a confusingModuleNotFoundError with no guidance on how to resolve it. Using poetry install --extras dev installs all development dependencies including pytest-cases in one command.

Replaced poetry install with poetry install --extras dev
Added a Note clarifying why --extras dev is needed
Added a Python version note after Prerequisites
Added a Tip in the Testing section for contributors working
on new features like the NL interface
Documentation only — no functional code was changed.

Fixes https://github.com/malariagen/malariagen-data-python/issues/1029


jwgarber:
The best way to fix this is to use `poetry install --extras dev`, that will install all the pytest dependencies.

Sally-designs:
Thank you @jwgarber let me look into that and change

PR #1036, opened 2026-03-04
Make pytest functions reproducible (1/2)

jwgarber:
This converts about half of the pytest functions so their random number generation is reproducible. The name of the pytest function test case is used to seed the RNG to ensure all functions generate independent random numbers. There are a few instances where the Numpy RNG has slightly different behaviour than the Python ones (e.g. `random.randint()` includes the endpoint by default while `rng.integers()` does not), so that had to be adapted. There are also a few cases where the numpy types have to be cast to Python types to make the type checker happy (e.g. `np.int64 -> int`), otherwise it's quite similar. 

This was tested by trying to reproduce [this flaky test failure](https://github.com/malariagen/malariagen-data-python/actions/runs/22472053182/job/65090921923). After about 140 trials I was able to find a random seed that caused the crash (325228949307297790525745584766237006555 in this case), and I am able to reliably reproduce this crash with that seed. Once this is finished the next job will be debugging that.

There are a lot of functions to convert, so the other half will be done in another PR since this one is large already.

While testing this I noticed that the global seed wasn't being printed in the workflow logs, even though it is when you run tests individually. Apparently this is because the seed needs to be printed from the top-level `conftest.py`, otherwise it won't get picked up when running `pytest -v tests`. Anyway, I rearranged some of the files so it works now.

### DIFF
--- a/malariagen_data/anoph/pca.py
+++ b/malariagen_data/anoph/pca.py
@@ -455,8 +455,8 @@ class AnophelesPca(
         # Apply jitter if desired - helps spread out points when tightly clustered.
         if jitter_frac:
             rng = np.random.default_rng(seed=random_seed)
-            data[x] = _jitter(data[x], jitter_frac, random_state=rng)
-            data[y] = _jitter(data[y], jitter_frac, random_state=rng)
+            data[x] = _jitter(data[x], jitter_frac, rng)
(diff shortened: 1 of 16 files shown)


jwgarber:
@jonbrenas Any other comments on this?

jonbrenas (approved):
LGTM

PR #1043, opened 2026-03-05
Fix incorrect type hint for _aim_metadata_dtype

vshnvii:
The attribute "_aim_metadata_dtype" was annotated as "Optional[Mapping[str, Any]]"
but is initialized as a dictionary and later updated using ".update()".

Since "Mapping" does not support mutation, the correct type should be
"Dict[str, Any]".

This change updates the type hint to reflect the actual usage.


jonbrenas (approved):
LGTM

PR #1047, opened 2026-03-05
Fix LoggingHelper.set_level() not updating logger level

suhr25:
## SUMMARY

This PR fixes `LoggingHelper.set_level()` so it updates both the logger and handler levels. Previously, only the handler level changed, causing runtime log level updates (e.g., enabling DEBUG) to silently fail.

---

## FIX

* **Root cause:** `set_level()` updated only the handler level, while the logger level still filtered messages.
* **Fix:** update both logger and handler levels together.
* **Scope:** small fix + regression test.

**Before**

```python
def set_level(self, level):
    if self._handler:
        self._handler.setLevel(level)
```

**After**

```python
def set_level(self, level):
    self._logger.setLevel(level)
    if self._handler:
        self._handler.setLevel(level)
```

---

## VERIFICATION

1. Create a `LoggingHelper` with `debug=False`.
2. Call `set_level(logging.DEBUG)` and log a debug message.

### DIFF
--- a/malariagen_data/anoph/fst.py
+++ b/malariagen_data/anoph/fst.py
@@ -360,7 +360,7 @@ class AnophelesFstAnalysis(
     )
     def average_fst(
         self,
-        region: base_params.region,
+        region: base_params.regions,
         cohort1_query: base_params.sample_query,
         cohort2_query: base_params.sample_query,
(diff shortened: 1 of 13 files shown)


suhr25:
Hi @jonbrenas,
Kindly look at this PR.
Thanks!

PR #1053, opened 2026-03-05
Add cohort_geometries() to access cohort GeoJSON metadata

adilraza99:
## Summary
Add a new `cohort_geometries()` method to expose cohort boundary geometries through the API.

## Changes
- add `cohort_geometries()` to `AnophelesSampleMetadata`
- load GeoJSON cohort boundary data using the standard `json` library
- add caching consistent with existing metadata access patterns
- validate accepted `cohort_set` values
- add tests and minimal GeoJSON fixtures

## Impact
Provides access to cohort geometry metadata while keeping the API simple and dependency-free.

Part of  #435


jonbrenas:
Hi @adilraza99, could you create a notebook with an example of how to use this new function?

PR #1060, opened 2026-03-06
feat: add VCF export functionality resolves #1054

### LINKED ISSUE #1054: Adding other file formats option
All the data that the API produces and uses during its analyses is in the Zarr format (https://zarr.readthedocs.io/en/stable/). However, many common external tools (e.g., ADMIXTURE, bcftools, ...) use other file format. Additionally, many genomic research groups use R (https://www.r-project.org) for many of their analyses and Zarr doesn't have modules for use in R. 

In the interest of expanding the options for our partners, it would be helpful if some of our data could be exported into other commonly used formats (e.g., PLINK, VCF, ...). Some limited support for Plink-conversions already exists but it could certainly be extended to other types of data.

Tanisha127:
## Summary
Adds `biallelic_snps_to_vcf()` method to export Anopheles biallelic 
SNP data to VCF (Variant Call Format) file format, enabling use with 
external tools such as bcftools, GATK, and R packages like 
VariantAnnotation.

## Changes
- Added `VcfExporter` mixin class in `malariagen_data/anoph/to_vcf.py`
- Integrated `VcfExporter` into `AnophelesDataResource` in `anopheles.py`
- Follows existing `PlinkConverter` mixin architecture pattern
- Output follows VCF 4.1 specification

## Approach
- Reuses existing biallelic_snp_calls() API
- Genotype data written chunk-by-chunk (1000 variants at a time) directly from Dask arrays — the full genotype matrix is never loaded into memory
- Writes spec-compliant VCF 4.1 manually with proper GT fields, FILTER, INFO and FORMAT columns
- allel.write_vcf() was evaluated but does not support sample/genotype data, so a manual writer was used instead
- Supports same filtering parameters as biallelic_snps_to_plink()

## Why VCF?
VCF is the most widely used format in genomics and is supported by:
- bcftools and samtools
- GATK pipelines  
- R packages (VariantAnnotation, vcfR)
- ADMIXTURE and other population genetics tools

## TODO
- [ ] Add unit tests

### DIFF
--- a/malariagen_data/anoph/to_vcf.py
+++ b/malariagen_data/anoph/to_vcf.py
@@ -99,51 +99,33 @@ class VcfExporter(
             chunks=chunks,
         )
 
-        # Compute genotype array from dask
-        with self._dask_progress("Computing genotype array"):
-            gt_array = allel.GenotypeDaskArray(ds_snps["call_genotype"].data).compute()
-
         # Extract variant info
         with self._spinner("Preparing VCF output data"):
             chrom = ds_snps["variant_contig"].values
             pos = ds_snps["variant_position"].values
             alleles = ds_snps["variant_allele"].values
             ref = alleles[:, 0]
-            alt = alleles[:, 1]
+            alt = alleles[:, 1:]
             sample_ids = ds_snps["sample_id"].values
+
+        # Compute genotype array from dask
+        with self._dask_progress("Computing genotype array"):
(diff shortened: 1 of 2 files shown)


Tanisha127:
Thanks @saadte, you're correct — allel.write_vcf() does not support genotype data. I've replaced it with a chunked line-by-line writer that iterates 1000 variants at a time from Dask arrays, so the full genotype matrix is never loaded into memory. Updated now.

PR #1061, opened 2026-03-06
 Add VcfConverter  module to export biallelic SNPs to Variant Call Format

### LINKED ISSUE #1054: Adding other file formats option

shauryam2807:
Fixes #1054
## Problem
Currently, all genotype data accessed by the API is exclusively stored in Zarr format or exported minimally to PLINK format (via [to_plink.py](cci:7://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anoph/to_plink.py:0:0-0:0)). As outlined in #1054, there is a critical need to natively export these analytical genomic datasets into Variant Call Format (VCF), which is the overarching standardized data format overwhelmingly required by downstream tools like `bcftools`, `GATK`, and generic `R` pipelines. 
## Solution
Leveraging the existing backend design of [to_plink.py](cci:7://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anoph/to_plink.py:0:0-0:0), this PR acts as an architectural foundation for natively parsing internal genomic Zarr structures back into a structured `.vcf` file, opening up API usability. 
- Formats reference counts and variant arrays using `scikit-allel`'s highly optimized `allel.write_vcf()` method natively to efficiently format string components without large memory bloat.
- Maintains strict parameter conformity with existing methods (ex: `sample_sets`, `region`, `min_minor_ac`, `thin_offset`, etc.) so the API `.to_vcf()` signature feels identical to `.to_plink()`.
## Changes
- [malariagen_data/anoph/to_vcf.py](cci:7://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anoph/to_vcf.py:0:0-0:0) — New cooperative class [VcfConverter](cci:2://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anoph/to_vcf.py:15:0-136:28) holding the core [biallelic_snps_to_vcf](cci:1://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anoph/to_vcf.py:27:4-136:28) extraction logic utilizing `scikit-allel`. 
- [malariagen_data/anopheles.py](cci:7://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anopheles.py:0:0-0:0) — Registers [VcfConverter](cci:2://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anoph/to_vcf.py:15:0-136:28) into the main [AnophelesDataResource](cci:2://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anopheles.py:79:0-1938:29) object to natively expose the method to users.
## Testing
_Testing will be added directly into `tests/anoph/test_vcf.py` once the overarching [to_vcf()](cci:1://file:///c:/Users/hp/GSOC26/malariagen-data-python/malariagen_data/anoph/to_vcf.py:27:4-136:28) API signature is confirmed to meet analytical requirements._
@jonbrenas — Since this is the initial drafted architectural framework for the exporter, I would love some early feedback! Specifically:
1. Would you prefer we output these as uncompressed `.vcf` files initially, or pipe them directly into `.vcf.gz` using `bgzip` if partners primarily use compressed pipelines?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,9 +12,10 @@ This package provides Python tools for accessing and analyzing genomic data from
 
 You'll need:
 
-- Python 3.10.x (CI-tested version)
-- [Poetry](https://python-poetry.org/) for dependency management
-- [Git](https://git-scm.com/) for version control
+- [pipx](https://python-poetry.org/) for installing Python tools
+- [git](https://git-scm.com/) for version control
+
+Both of these can be installed using your distribution's package manager or [Homebrew](https://brew.sh/) on Mac.
 
 ### Initial setup
 
@@ -33,16 +34,25 @@ You'll need:
    git remote add upstream https://github.com/malariagen/malariagen-data-python.git
    ```
 
-3. **Install Poetry** (if not already installed)
+3. **Install Poetry**
 
    ```bash
    pipx install poetry
    ```
 
-4. **Install the project and its dependencies**
+4. **Install Python 3.12**
+
+   Python 3.12 is tested in the CI-system and is the recommended version to use.
 
    ```bash
-   poetry install
+   poetry python install 3.12
+   ```
+
+5. **Install the project and its dependencies**
+
+   ```bash
+   poetry env use 3.12
+   poetry install --extras dev
    ```
 
    **Recommended**: Use `poetry run` to run commands inside the virtual environment:
@@ -71,7 +81,7 @@ You'll need:
    python script.py
    ```
 
-5. **Install pre-commit hooks**
+6. **Install pre-commit hooks**
 
    ```bash
    pipx install pre-commit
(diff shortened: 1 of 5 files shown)


malariagen_data/anoph/to_vcf.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+        # Compute gt ref counts
+        with self._dask_progress("Computing genotype ref counts"):
+            gt_asc = ds_snps["call_genotype"].data  # dask array
+            gn_ref = allel.GenotypeDaskArray(gt_asc).to_n_ref(fill=-127)
+            gn_ref = gn_ref.compute()
+
+        # Ensure genotypes vary
+        loc_var = np.any(gn_ref != gn_ref[:, 0, np.newaxis], axis=1)
```
saadte: Can you please explain the approach you have taken here? You are setting missing calls to the value `-127`, right?
Then, just below, you do:
`loc_var = np.any(gn_ref != gn_ref[:, 0, np.newaxis], axis=1)`
How does this ensure that something like `[-127,0,0,0]` is not marked as varying?

PR #1063, opened 2026-03-06
Added LD pruning support and PLINK export for ADMIXTURE analysis 

### LINKED ISSUE #1049: Adding Admixture functionalities to the API
The ancestry of a population can significantly bias many association studies. It is also critical for analyses of population demographics, genetic drift, ...

ADMIXTURE (https://dalexander.github.io/admixture/publications.html, https://pmc.ncbi.nlm.nih.gov/articles/PMC2752134/) is a commonly used tool to estimate individual ancestry.

Adding Admixture functionalities to the API would be of great benefit. This can be done either by extracting the data from the API into a format that is compatible with ADMIXTURE (i.e., LD-pruned PLINK), or the algorithm can be reimplemented, ...

rehanxt5:
## Solution

I have added LD pruning capabilities to the API, allowing  the users to inspect and export admixture compatible data in a single call. The implementation uses scikit-allel's `locate_unlinked()` , to do LD pruning.

I created a new module `ld_params.py` this holds the parameters description of ld_pruning function. Created `ld.py` where the actual functions/methods live: 

- `biallelic_snps_ld_pruned()` : it is used to perform ld pruning on the biallelic SNPs
- `biallelic_snps_ld_pruned_to_plink()`: it is used to generate the data in plink format , which is ready to use with admixture

All the changes in detail are discussed below:

### New Module: ld_params.py
It Defines 3 configurable parameters for LD pruning:
- `size` (default 500): It is the window size in number of SNPs
- `step` (default 200): How many SNPs to advance the window each iteration
- `threshold` (default 0.1): r² threshold for linkage disequilibrium

### New Mixin: ld.py (`AnophelesLdAnalysis`)

**Public method `biallelic_snps_ld_pruned()`**
- it performs LD pruning on biallelic SNPs for a given region and sample set 
- it supports all standard filtering: `min_minor_ac`, `max_missing_an`, `n_snps`, cohort downsampling
- it returns an xarray Dataset with pruned variants
- The results are cached efficiently (stores only the pruning mask, not full genotypes)
- I also added a warning , it will warn the users about memory usage before computing

**Public method `biallelic_snps_ld_pruned_to_plink()`**
- It combines LD pruning with PLINK binary export in one call.
- Outputs .bed, .bim, .fam files compatible with ADMIXTURE
- It also supports file overwriting control
- To avoid redundant computation , it uses cached results.

### Updated anopheles.py
- Added `AnophelesLdAnalysis` to the `AnophelesDataResource` class MRO (between `AnophelesPca` and `PlinkConverter`)
- Both methods now available on all Anopheles species API objects

### Documentation
- Updated RST docs for all 4 species (Ag3, Af1, Adir1, Amin1)
- Added both new methods to API autosummary

### Tests
- `test_biallelic_snps_ld_pruned`: Basic functionality, caching, result validation
- `test_biallelic_snps_ld_pruned_with_n_snps`: Pre-filtering with SNP count limit
- `test_biallelic_snps_ld_pruned_with_cohort_size`: Individual downsampling
- `test_biallelic_snps_ld_pruned_to_plink`: PLINK export validation

## Usage Example

```python
import malariagen_data

ag3 = malariagen_data.Ag3()

# Export LD-pruned PLINK files ready for ADMIXTURE
plink_path = ag3.biallelic_snps_ld_pruned_to_plink(
    output_dir="./admixture_input",
    region="3L",
    size=500,
    step=200,
    threshold=0.1
)

# Then run: admixture ./admixture_input/3L.ld_pruned*.bed K
```

## Technical Details

- The LD pruning uses scikit-allel's `locate_unlinked()` method
- Caching stores only the boolean pruning mask (efficient).
- Memory warning before compute step alerts users to potential RAM usage.
- Lazy evaluation with dask arrays until final compute.
- Supports all existing SNP filtering and sample selection parameters.

## Testing

- All 8 new tests pass
- No regression in existing PLINK converter tests
- Ruff linting passes (code style compliant)
- Tested on ag3_sim and af1_sim fixtures

Closes #1049


saadte:
Hi @rehanxt5,

Previously, I had asked how you determine, as you have commented in your code:
>Define output file path using parameters that affect the content.

and had asked for some understanding on:

> Have you considered the possibility of a file collision if the parameters you skip for naming the file change the content? That might result in files with different content pointing to the exact same file, causing silent downstream problems.

This can prove to be critical if not addressed. Can you please have a look and help me understand how exactly do you determine the relevant parameters and prevent collisions?

rehanxt5:
> I will implement this into `biallelic_snps_ld_pruned_plink()` within this PR. However, fixing `biallelic_snps_to_plink()` in this PR would be scope creep. I will open a new issue and fix it there for `biallelic_snps_to_plink()`.

hey @jonbrenas @saadte i have adressed the File collision issue in the `biallelic_snps_ld_pruned_to_plink` in this PR , i have used the `_hash_params` utility to generate a hash using `sample_sets` , `sample_query` , `sample_query_options` etc , and add first 8 characters of these to postfix , 

In #1114 i have did the same for `biallelic_snps_to_plink`

rehanxt5:
thanks @jonbrenas 

1. the `_write_plink` function doesn't require if the data is biallelic_snps or ld_pruned , it just expects x.array dataset , so it can be used to write any plink file after any filteration (MAF-filtered , thinned, or raw ). However it does expect the snps to be biallelic , but we require the snps to be biallelic only as plink only supports two alleles, So we do not require to create a different writer function for every filteration

2. You're right best would be to allow the users to specify the filename , I will implement a hybrid approach , where the user can specify the filename , or else the function generates it.

PR #1065, opened 2026-03-06
Optimize quarter column computation using vectorized NumPy operations

vshnvii:
This PR replaces the row-wise pandas "apply()" used to compute the "quarter"
column with a vectorized NumPy implementation.

Previous implementation:

df["quarter"] = df.apply(
    lambda row: ((row.month - 1) // 3) + 1 if row.month > 0 else -1,
    axis="columns",
)

Row-wise "apply()" loops through rows in Python and can be slow for large datasets.

The new implementation uses NumPy vectorization:

df["quarter"] = np.where(
    df["month"] > 0,
    ((df["month"] - 1) // 3) + 1,
    -1
)

This improves performance and follows pandas best practices.

### DIFF
--- a/malariagen_data/anoph/sample_metadata.py
+++ b/malariagen_data/anoph/sample_metadata.py
@@ -181,9 +181,10 @@ class AnophelesSampleMetadata(AnophelesBase):
             df["release"] = release
 
             # Derive a quarter column from month.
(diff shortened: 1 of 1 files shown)


jonbrenas (approved):
LGTM

PR #1076, opened 2026-03-07
feat: allow snp_allele_frequencies to accept genomic regions

### LINKED ISSUE #805: Modify snp_allele_frequencies to allow investigation of any genomic region.
Currently SNP tables can be generated but require a transcript as an input parameter which restricts its use to coding regions. However, there are cases for looking regions lying outside of genes and therefore it would be useful to modify the function to look at any genomic region input. Since snp_calls can already take a region (or transcript) as a parameter, this would require changes to the code adding the effect annotations and the creation of the title only (could be generated only if a transcript is used).

joshitha1808:
### What problem does this solve:
Previously, `snp_allele_frequencies()` required a transcript as input. This meant analyses were limited to coding regions, and users couldn’t easily check SNP allele frequencies for arbitrary genomic regions.

### How does it solve it:
This PR adds an optional `region` parameter, so users can now specify any genomic region directly. The function will accept either a transcript or a region .When a region is used, SNP calls and allele frequencies are calculated for that region. Effect annotations are still only generated if a transcript is provided. Plot titles are automatically adjusted to work with both transcript- and region-based queries. 

### Issue number:
Closes #805 

### Testing done:

Added unit tests for region-based queries.

Verified that all existing transcript-based tests still pass.

Confirmed that effect annotations are applied correctly when using transcripts.

### Breaking changes or migration notes:
No.


jonbrenas (approved):

joshitha1808:
Hi @jonbrenas ,
Thanks for the update. Happy to hear the solution works for now.

Issue #837, opened 2025-12-03
Add support for Anopheles darlingi (Adar1.x)


jonbrenas:
This issue is twinned with https://github.com/malariagen/vobs-darlingi/issues/11 (for data) and https://github.com/malariagen/vector-data/issues/113 (for the docs).

varun-kht:
@jonbrenas sure when done with it ,pls tag would love to go through it .

PR #1086, opened 2026-03-09
Adding limited support for Adar

jonbrenas:
Addresses #837.

The only functions available from Adir and not from Adar are the ones working with IGV (no data on S3 that can be used by IGV) and `site_annotations` as the Zarr containing them as not been created


ahernank (approved):
Thanks @jonbrenas!

PR #885, opened 2026-02-20
docs: add macOS developer setup guide

PaulaYaniz:
Fixes #726

The existing developer setup guide uses Linux-specific tooling (apt, pipx, poetry), which doesn't work on macOS. This adds a dedicated macOS section covering:

- Miniconda installation with Apple Silicon / Intel detection
- Python version requirements (>=3.10, <3.13)
- pip-based dev install instead of poetry
- pre-commit hooks setup
- Running tests without poetry
- Google Cloud CLI via brew and authentication
- VS Code terminal integration

PR #860, opened 2026-02-12
Add comprehensive CONTRIBUTING.md guide for new contributors

adilraza99:
## Summary

Add a comprehensive CONTRIBUTING.md guide to standardize the contribution workflow and improve onboarding for new contributors.

This guide includes:

- Development environment setup (Poetry, pre-commit, Python 3.10)
- Recommended branching and pull request workflow
- Testing and code quality practices
- Documentation guidelines
- Communication channels for contributors

This is a documentation-only change. No application code was modified.

Fixes #859

PR #1099, opened 2026-03-11
Add DEBIAN_SETUP.md to support Debian and WSL environments (Fixes #1091)

### LINKED ISSUE #1091: Docs: LINUX_SETUP.md relies on Ubuntu-specific PPA
Hi team, nice to meet you! I am looking to contribute to the project and was going through the `LINUX_SETUP.md` guide. However, step 2 fails on my setup (Debian 12 Bookworm) when running `sudo add-apt-repository ppa:deadsnakes/ppa`.

```bash
sudo add-apt-repository ppa:deadsnakes/ppa
Traceback (most recent call last):
  File "/usr/bin/add-apt-repository", line 362, in <module>
    sys.exit(0 if addaptrepo.main() else 1)
                  ^^^^^^^^^^^^^^^^^
  File "/usr/bin/add-apt-repository", line 345, in main
    shortcut = handler(source, **shortcut_params)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/softwareproperties/shortcuts.py", line 40, in shortcut_handler
    return handler(shortcut, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/softwareproperties/ppa.py", line 86, in __init__
    if self.lpppa.publish_debug_symbols:
       ^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/softwareproperties/ppa.py", line 126, in lpppa
    self._lpppa = self.lpteam.getPPAByName(name=self.ppaname)
                  ^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/softwareproperties/ppa.py", line 113, in lpteam
    self._lpteam = self.lp.people(self.teamname)
                   ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'people'
```

The setup script relies on a Launchpad PPA, which is specific to Ubuntu. On Debian and other Linux distributions, this command fails because Launchpad is not natively supported. Because of this, the guide is currently Ubuntu-exclusive rather than general Linux.

I suggest updating the documentation to recommend a distribution-agnostic tool like `pyenv` for Python version management. It works consistently across Debian, Mint, Arch, etc.

The tradeoff with `pyenv` is that it requires compiling Python from source (which takes longer) and involves modifying the `PATH` ordering, which could introduce edge cases for some users.

I also thought about suggesting Docker as an isolated environment, but to be completely honest, I am not familiar enough with it to accurately assess if it's a good fit for this specific project.

Let me know your thoughts on this approach. If you agree that `pyenv` is a viable alternative, I would be happy to work on a PR to update the setup documentation! :D

Schronding:
Fixes #1091

**What problem does this solve?** The current `LINUX_SETUP.md` relies on Launchpad PPAs, which are Ubuntu-specific and fail on Debian and other Linux distributions (detailed in #1091). This creates a barrier for contributors using non-Ubuntu environments.

**How does it solve it?** This PR adds a dedicated `DEBIAN_SETUP.md` guide.

- **The "Why" behind the technical choices:** While I initially proposed using `pyenv` in the original issue, further research showed that `uv` is significantly faster and handles both Python installation and virtual environments seamlessly.
- I considered suggesting `uv` as a global replacement for package management across the project, but since the test suite and repository heavily rely on `poetry`, I decided to keep `uv` strictly scoped to this Debian setup guide. I didn't want to make an irresponsible or disruptive change to the existing test architecture.
    

**Testing done**

- Tested the setup instructions step-by-step natively on Debian 12 (Bookworm).
- Tested the setup instructions on a separate machine using Debian via WSL.
- Ran the local test suite successfully: `952 passed, 4 skipped, 241 warnings in 760.47s`.

**Breaking changes or migration notes** None. This is purely an additive documentation change. I am completely open to feedback or expanding this to cover more distributions if the team thinks it would be valuable!


jwgarber:
There are currently several setup guides, one for Mac, Linux, and the general contributing guide. It would be good to have a single guide that works for everyone as much as possible. The Contributing.md guide suggests installing python with poetry, do you know if that works on Debian? (Incidentally I agree that uv is better, but since we have poetry already it's probably simplest to stick to a single tool for now.)

Schronding:
@jwgarber Thank you so much for your comment! I have checked my solution again and I understand the source of the problem now. There were some things that I misunderstood.

1. I thought the `deadsnakes` package was a necessary evil for Ubuntu, but as I kept exploring the alternatives, I noticed that I can use the same commands for Debian and Mint that I use for Ubuntu, but I definitely need to remove `deadsnakes`.

2. When I did my research about Poetry, I saw that it didn't allow you to control the specific Python version, but after a second check, I found that newer versions do allow you to select it with the `poetry python install <python_version>` command.

As I know the MalariaGEN team is very busy, I went ahead and modified the file directly with the changes so you can tell me if there is another problem.

The reason why I am leaving the local installation in `LINUX_SETUP.md` instead of merging it into `CONTRIBUTING.md` is that @PaulaYaniz made a commit separating it about 3 weeks ago; I want to add to her contribution, not remove it.

Her commit says:

> Per review feedback on #885, moved Linux and macOS developer setup instructions into dedicated LINUX_SETUP.md and MACOS_SETUP.md files. README now links to both. These will be referenced from the upcoming CONTRIBUTING.md (#860).

jonbrenas (approved):

PR #1122, opened 2026-03-14
Improve PLINK filename hashing with canonicalized parameters and SHA256

### LINKED ISSUE #1113: Hash PLINK export filenames by content-affecting selection parameters to prevent collisions
## Summary

The PLINK export methods currently generate output filenames from only a subset of their input parameters. As a result, different export requests can resolve to the same filename prefix even when they produce different data.

This affects:

1. `biallelic_snps_to_plink()`
2. `biallelic_snps_ld_pruned_to_plink()`

## Problem

At the moment, the filename prefix is built from parameters such as region-level SNP filtering, but it excludes several parameters that also affect the exported dataset contents, including:

- `sample_sets`
- `sample_query`
- `sample_query_options`
- `sample_indices`
- `site_mask`
- `random_seed`

Because these parameters are not reflected in the filename, two calls with different sample-selection inputs can write to the same `.bed/.bim/.fam` prefix.

## Example collision

Two calls such as:

```python
api.biallelic_snps_to_plink(
    output_dir="/plink",
    region="2L",
    n_snps=1000,
    sample_sets=["set_a"],
    random_seed=42,
)
```

and

```python
api.biallelic_snps_to_plink(
    output_dir="/plink",
    region="2L",
    n_snps=1000,
    sample_sets=["set_b"],
    random_seed=42,
)
```

can currently generate the same output prefix, despite exporting different sample cohorts.

## Why this is a problem

- Different exports can silently overwrite each other.
- Cached/exported file paths are not a stable identifier of dataset contents.
- The risk increases when users run multiple exports into the same `output_dir`.

mithil27360:
Fixes #1113

This PR ensures that PLINK export filenames uniquely represent the dataset
selection by hashing all content-affecting parameters.

**Changes:**
- Introduce [_compute_plink_selection_hash()](cci:1://file:///Users/mithils/Desktop/malariagen-data-python/malariagen_data/anoph/to_plink.py:17:0-44:60) helper to compute a deterministic
  hash from selection parameters
- Canonicalize list inputs such as `sample_sets` via sorting to avoid
  order-dependent hashes
- Use truncated SHA256 for stable, collision-resistant identifiers
- Append the hash to the existing filename prefix while preserving human readability

**Example filename:** `2L.1000.2.8.0.959c5581b5`

**Tests added to verify:**
- Deterministic hashing (same params → same hash)
- Different parameters produce different hashes
- Ordering of `sample_sets` does not affect the result
- Hash format and length


mithil27360:
Could a maintainer please approve the workflows to run CI? Thanks!

PR #1124, opened 2026-03-15
# Feat: Add `get_metadata_schema` Helper for Sample Metadata Introspection

### LINKED ISSUE #1123: # Feature Request: Introduce a Structured Query Abstraction Layer to Support API Extensions and Natural Language Interfaces
## Problem Description

Currently, filtering datasets via `sample_query` within `AnophelesDataResource` methods relies entirely on raw pandas evaluation strings, for example:

```python
sample_query = "country == 'Uganda' and year > 2014"

rajat552:
### The Problem
Currently, when constructing a `sample_query` to filter dataset metadata or analysis methods (like PCA or Diversity calculations), users (or programmatic clients like an LLM) have no straightforward programmatic way to know what columns are available or what the possible valid values are (e.g., discovering all unique `country` or `taxon` values).

### The Solution
This PR introduces a small, non-breaking helper method `get_metadata_schema()` to the `AnophelesSampleMetadata` mixin class. It programmatically introspects the sample metadata dataframe and exposes a dictionary of the schema.

It automatically categorizes columns by parsing `df.columns` and `df.dtypes` to build an intelligent dictionary mapping:

- **Low-cardinality Categoricals (Strings/Objects with ≤ 100 unique values):** Returns the `type` as `"categorical"` and enumerates the possible `"values"` (e.g., for `taxon`, `country`).
- **High-cardinality Strings:** Returns `"type": "string"`.
- **Numeric columns (Int/Float):** Returns `"type": "numeric"` and calculates the `"min"` and `"max"` bounds (e.g., for `year`).
- **Booleans:** Returns `"type": "boolean"` and `"values": [True, False]`.

Closes #1123 

## Example Usage

```python
import malariagen_data

ag3 = malariagen_data.Ag3()
schema = ag3.get_metadata_schema(sample_sets="AG1000G-UG")

print(schema["country"])
# Output: {'type': 'categorical', 'values': ['Uganda']}

print(schema["year"])
# Output: {'type': 'numeric', 'min': 2012, 'max': 2012}

### DIFF
--- a/malariagen_data/anoph/sample_metadata.py
+++ b/malariagen_data/anoph/sample_metadata.py
@@ -847,7 +847,7 @@ class AnophelesSampleMetadata(AnophelesBase):
                 schema[col] = {"type": "boolean", "values": [True, False]}
             else:
                 schema[col] = {"type": "other"}
-                
+
         return schema
 
     @_check_types


jonbrenas:
Thanks @rajat552. I think you have an interesting idea. However, this PR doesn't cover the first line of your stated problem at all (i.e., PCA and diversity stats) as it only applies to the sample metadata. I am pretty sure there are pandas functions that do most of what you are doing already.

rajat552:
@jonbrenas 
Thanks for the feedback!

You’re right that the current implementation focuses on sample metadata rather than directly covering PCA or diversity statistics. My intention with this PR was to introduce a small helper that exposes metadata schema information programmatically, which could act as a building block for safer query construction.

While pandas can indeed provide much of this functionality, the goal here was to expose a simple and consistent API-level method that surfaces this information directly within malariagen_data. This could help users (and programmatic systems such as automated analysis tools or natural-language interfaces) discover valid metadata fields and values before constructing queries.

If you think this functionality should instead be implemented differently, or placed elsewhere in the API, I’d be happy to adapt the approach based on your guidance.

PR #1149, opened 2026-03-18
Fix #940: Replace truthiness checks with explicit None checks for reg…

### LINKED ISSUE #940: if region.start or region.end` silently returns entire contig when `start=0
## Summary

A truthiness bug in 7 region boundary checks causes region slicing to be silently
skipped when `region.start = 0`. Instead of returning the requested subregion, the
full contig is returned with no error or warning.

## Background

Several functions in `snp_data.py` and `hap_data.py` gate their region-slicing logic
on the condition:
```python
if region.start or region.end:
    # slice data to region
```

The intent is to check whether a region boundary has been specified. However, Python
evaluates `0` as falsy, so when `region.start = 0` and `region.end` is `None`, the
entire condition evaluates to `False` and the slicing block is skipped entirely.

## Affected locations

| File | Line | Function |
|---|---|---|
| `snp_data.py` | 244 | `_site_filters_for_region` |
| `snp_data.py` | 344 | `_snp_sites_for_region` |
| `snp_data.py` | 497 | `snp_genotypes` |
| `snp_data.py` | 661 | `snp_variants` |
| `snp_data.py` | 1134 | `_cached_snp_calls` |
| `hap_data.py` | 146 | `_haplotype_sites_for_region` |
| `hap_data.py` | 388 | `haplotypes` |

## Expected behavior

When `region.start = 0` is specified, data should be sliced from position 0 to
`region.end`. The value `0` is a valid and meaningful genomic coordinate.

## Actual behavior

Slicing is skipped. The entire contig is returned as if no region was specified.

## Why this is hard to catch

- The failure is **silent** — no exception is raised, no warning is logged
- Most real-world region queries don't start at position 0, so routine usage and
  typical test cases don't exercise this path
- The returned data looks plausible — it's valid, well-formed data, just the wrong
  amount of it
- Any downstream analysis will run to completion on the inflated dataset, potentially
  producing incorrect results that are never traced back to this root cause

## Proposed fix

Replace all 7 instances with an explicit `None` check:
```python
# Before
if region.start or region.end:

# After
if region.start is not None or region.end is not None:
```

This correctly distinguishes between "no boundary was specified" (`None`) and
"boundary is zero" (`0`).

## Suggested additional test coverage

Add test cases for each affected function where `region.start = 0`, asserting that
the returned data is correctly sliced rather than spanning the full contig.

khushthecoder:
# Fix `region.start = 0` truthiness bug in region slicing (#940)

**Fixes:** [#940](https://github.com/malariagen/malariagen-data-python/issues/940)

## Description

A truthiness bug in 7 region boundary checks causes region slicing to be **silently skipped** when `region.start = 0`. Instead of returning the requested subregion, the full contig is returned — with no error or warning.

### The bug

Several functions in `snp_data.py` and `hap_data.py` gate their region-slicing logic on:

```python
if region.start or region.end:
    # slice data to region
```

The intent is to check whether a region boundary has been specified. However, Python evaluates `0` as **falsy**, so when `region.start = 0` and `region.end` is `None` (or any integer), the entire condition short-circuits incorrectly:

```python
# region.start = 0, region.end = 1000
if 0 or 1000:  # evaluates to 1000 (truthy) ✅ works by accident

# region.start = 0, region.end = None
if 0 or None:  # evaluates to None (falsy) ❌ slicing is SKIPPED
```

### Affected locations (all 7 fixed)

| File | Line | Function |
|------|------|----------|
| `snp_data.py` | 256 | `_site_filters_for_region` |
| `snp_data.py` | 356 | `_snp_sites_for_region` |
| `snp_data.py` | 509 | `snp_genotypes` |
| `snp_data.py` | 673 | `snp_variants` |
| `snp_data.py` | 1144 | `_cached_snp_calls` |
| `hap_data.py` | 146 | `_haplotype_sites_for_region` |
| `hap_data.py` | 388 | `haplotypes` |

## Impact

This bug is **particularly dangerous** because:

1. **Silent failure** — No exception is raised, no warning is logged. The function returns valid, well-formed data — just the **wrong amount** of it.

2. **Incorrect scientific results** — Any downstream analysis (allele frequency calculations, diversity statistics, haplotype clustering) will run to completion on the inflated dataset, potentially producing **incorrect results that are never traced back to this root cause**.

3. **Hard to catch in practice** — Most real-world region queries don't start at position 0, so routine usage and typical test cases don't exercise this path. The returned data looks plausible.

4. **Wide blast radius** — The bug affects **7 functions** across the two core data-access modules (`snp_data.py` and `hap_data.py`), meaning every major data retrieval pathway is affected: SNP sites, site filters, genotypes, variants, cached SNP calls, haplotype sites, and haplotypes.

5. **Coordinate system relevance** — Position 0 is meaningful in genomic computing (0-based coordinate systems, start-of-contig queries, internal slicing conventions). Users who adopt 0-based indexing will silently get wrong data.

### Root cause

The root cause is a classic Python truthiness trap. The original code uses `or` to check if a boundary was specified:

```python
if region.start or region.end:
```

But this conflates "not specified" (`None`) with "specified as zero" (`0`). The correct check is an **explicit `None` comparison**.

### Fix applied

Replaced all 7 occurrences with:

```diff
- if region.start or region.end:
+ if region.start is not None or region.end is not None:
```

And similarly for the variable `r`:

```diff
- if r.start or r.end:
+ if r.start is not None or r.end is not None:
```

### Why this fix is correct

- `is not None` correctly distinguishes between "no boundary was specified" (`None`) and "boundary is zero" (`0`)
- The fix is **minimal and focused** — only the boolean condition changes; no other logic is touched
- **Zero behavior change** for the common case where `start` and `end` are `None` (returns full contig as before)
- **Zero behavior change** for the common case where `start > 0` (works correctly already — just by accident)
- **Fixes the bug** for the `start = 0` case

### Why I didn't use a helper function

The issue reporter suggested defining a `_region_has_bounds()` helper. I considered this but chose the inline `is not None` approach because:

- It's the **most minimal** change, reducing review burden
- The pattern is immediately readable — Python developers recognize `is not None` as the idiomatic None-check
- Adding a helper would touch a third file (wherever region utilities live) and expand the diff unnecessarily
- If the maintainers prefer a helper, it can be added in a follow-up refactor

---

### DIFF
--- a/malariagen_data/anoph/genome_sequence.py
+++ b/malariagen_data/anoph/genome_sequence.py
@@ -117,9 +117,13 @@ class AnophelesGenomeSequenceData(AnophelesBase):
 
         # Deal with region start and stop.
         slice_start = slice_stop = None
-        if resolved_region.start:
+        if resolved_region.start is not None:
+            if resolved_region.start < 1:
+                raise ValueError("Region start must be >= 1 or None.")
             slice_start = resolved_region.start - 1
-        if resolved_region.end:
+        if resolved_region.end is not None:
+            if resolved_region.end < 1:
+                raise ValueError("Region end must be >= 1 or None.")
             slice_stop = resolved_region.end
         loc_region = slice(slice_start, slice_stop)
 
--- a/malariagen_data/anoph/snp_data.py
+++ b/malariagen_data/anoph/snp_data.py
@@ -1719,7 +1719,7 @@ class AnophelesSnpData(
 
         # Access SNP site positions.
         pos = self.snp_sites(region=resolved_region, field="POS").compute()
-        if resolved_region.start:
+        if resolved_region.start is not None:
             offset = resolved_region.start
         else:
             offset = 1
--- a/malariagen_data/anoph/snp_frq.py
+++ b/malariagen_data/anoph/snp_frq.py
@@ -671,8 +671,10 @@ class AnophelesSnpFrequencyAnalysis(AnophelesSnpData, AnophelesFrequencyAnalysis
 
             # Check for no SNPs remaining after applying variant query.
             if np.count_nonzero(loc_variants) == 0:
-                raise ValueError(
-                    f"No SNPs remaining after applying variant query {variant_query!r}."
+                warnings.warn(
+                    f"No SNPs remaining after applying variant query {variant_query!r}. "
+                    "Returning dataset with zero variants.",
+                    stacklevel=2,
                 )
 
             # Convert boolean mask to integer indices for NumPy 2.x compatibility
@@ -813,8 +815,10 @@ class AnophelesSnpFrequencyAnalysis(AnophelesSnpData, AnophelesFrequencyAnalysis
 
             # Check for no SNPs remaining after applying variant query.
             if np.count_nonzero(loc_variants) == 0:
-                raise ValueError(
-                    f"No SNPs remaining after applying variant query {variant_query!r}."
+                warnings.warn(
+                    f"No SNPs remaining after applying variant query {variant_query!r}. "
+                    "Returning dataset with zero variants.",
+                    stacklevel=2,
                 )
 
             # Convert boolean mask to integer indices for NumPy 2.x compatibility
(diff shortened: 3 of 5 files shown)


khushthecoder:
Hi @jonbrenas , all test cases for this PR have passed. Please review it, and let me know if any further changes are needed.

khushthecoder:
Hi @jonbrenas, just a gentle reminder — all tests have passed successfully and the PR is ready. Please let me know if anything else is needed from my side.

PR #1152, opened 2026-03-18
fix: raise clear error when window_size is too large

suhr25:
## SUMMARY

Fixes a crash when `window_size` is too large in Fst GWSS by raising a clear error instead of returning empty data.

---

## FIX

**Before**

```python
x = allel.moving_statistic(pos, statistic=np.mean, size=window_size)
return dict(x=x, fst=fst)
```

**After**

```python
x = allel.moving_statistic(pos, statistic=np.mean, size=window_size)

if len(x) == 0:
    raise ValueError("window_size is larger than available SNPs")

return dict(x=x, fst=fst)
```

---

## VERIFICATION

Using a large `window_size` on a small region used to crash with an `IndexError`. Now it raises a clear `ValueError`. A test confirms this behavior.

### DIFF
--- a/malariagen_data/anoph/fst.py
+++ b/malariagen_data/anoph/fst.py
@@ -123,6 +123,17 @@ class AnophelesFstAnalysis(
             is raised if the number of available SNP sites is below
             min_snps_threshold.
         """,
+        parameters=dict(
+            min_snps_threshold="""
+                Minimum number of SNP sites required. If fewer sites are
+                available a ValueError is raised.
+            """,
+            window_adjustment_factor="""
+                If window_size is >= the number of available SNP sites,
+                window_size is automatically set to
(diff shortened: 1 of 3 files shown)


suhr25:
Thanks, that makes sense. I've updated the implementation to automatically adjust window_size when possible and only raise an error for very small SNP counts. Let me know if you prefer different defaults.

suhr25:
Hi @jonbrenas,
Gentle follow-up on this PR.
Thanks!

jonbrenas:
Hi @suhr25,

Both _min_snps_threshold and _window_adjustment_factor should be parameters of the function (with these values by default as most people will not set values for them).

suhr25:
Hi @jonbrenas,
Made changes which you asked for ,kindly take a look.
Thanks!

PR #1157, opened 2026-03-19
refactor: tighten type annotations for cache variables in sample_metadata.py

faizavk:
This PR tightens the type annotations for two cache variables in AnophelesSampleMetadata.__init__. _cache_sample_metadata and _cache_cohort_geometries were previously typed as bare Dict with no key or value types. They have been updated to Dict[Tuple[str, ...], pd.DataFrame] and Dict[Tuple[str, str], Dict[str, Any]] respectively, based on the actual types stored and retrieved in the codebase. This follows the same pattern as PR #1129.


faizavk:
I noticed _cache_cohorts on line 87 is also a bare Dict in the same block. I've updated the PR to include this annotation as well.

PR #1161, opened 2026-03-19
Fix #1160: Replace broad except Exception with specific exception typ…

### LINKED ISSUE #1160: Bug: phenotypes.py swallows all exceptions via broad except Exception, silently masking data and logic errors
# Fix: Replace Broad `except Exception` Anti-Pattern in `phenotypes.py`
 
## Description
 
While exploring the API and testing robust error handling for a GSoC 2026 proposal, a recurring anti-pattern was identified in `malariagen_data/anoph/phenotypes.py`. The code uses broad `except Exception:` clauses to catch all errors and convert them into mere warnings, causing actual bugs to be silently ignored and returning empty or corrupted data instead of failing loudly.
 
---
 
## Where the Problem Is
 
There are **5 instances** of `except Exception:` in `phenotypes.py` doing exactly this:
 
### 1. Lines 73–75: Loading phenotype data
 
```python
except Exception as e:
    warnings.warn(f"Unexpected error loading phenotype data for {sample_set}: {e}")
```
 
### 2. Lines 310–313: Merging phenotype and variant data
 
```python
except Exception as e:
    warnings.warn(f"Unexpected error selecting/merging variant data: {e}")
return ds
```
 
### 3. Lines 351–353: Fetching sample metadata
 
```python
except Exception as e:
    warnings.warn(f"Error fetching sample metadata: {e}")
return pd.DataFrame()
```
 
### 4. Lines 376–378: Evaluating `sample_query`
 
```python
except Exception as e:
    warnings.warn(f"Error applying sample_query '{sample_query}': {e}")
return pd.DataFrame()
```
 
### 5. Lines 495–498: Checking if phenotype path exists
 
```python
if self._fs.exists(phenotype_path):
    phenotype_sample_sets.append(sample_set)
except Exception:
    continue
```
 
---
 
### 1. Silently Masked Bugs
 
A simple typo in a query string, a `KeyError` from a missing `sample_id` column, or a `TypeError` from Pandas will all trigger the `except Exception` block. Instead of raising a traceback to tell the user what went wrong, the code silently warns and returns an empty `pd.DataFrame()` or an unmodified dataset.
 
### 2. Data Integrity Risk
 
Users might continue their downstream scientific analysis without realizing their phenotype query failed or returned truncated results. In bioinformatics pipelines, an empty result caused by a `TypeError` **should fail the pipeline**, not just print a warning and proceed with `0` records.
 
### 3. Debugging Nightmare
 
If a bug originates deep inside `Dask`, `Pandas`, or the `fsspec` filesystem layer, catching `Exception` destroys the stack trace, making it much harder for maintainers to fix issues that users report.
 
---
 
## Proposed Fix
 
Replace these broad catch-all exceptions with the **specific exception classes** we actually expect to handle (e.g., `FileNotFoundError`, `pd.errors.UndefinedVariableError`), and let unexpected errors crash naturally with a full traceback.
 
### Example 1: Applying a query (Lines 376–378)
 
**Before:**
```python
except Exception as e:
    warnings.warn(f"Error applying sample_query '{sample_query}': {e}")
return pd.DataFrame()
```
 
**After:**
```python
except pd.errors.UndefinedVariableError as e:  # Catch specifically bad query syntax
    warnings.warn(f"Invalid sample_query '{sample_query}': {e}")
return pd.DataFrame()
# Let unexpected TypeErrors/KeyErrors propagate naturally
```
 
### Example 2: Checking if phenotype path exists (Lines 495–498)
 
**Before:**
```python
try:
    if self._fs.exists(phenotype_path):
        phenotype_sample_sets.append(sample_set)
except Exception:
    continue
```
 
**After:**
```python
# Rather than try/except Exception, let network/auth errors propagate
if self._fs.exists(phenotype_path):
    phenotype_sample_sets.append(sample_set)
```
 
---
 
## Summary
 
| Location | Current Behaviour | Proposed Fix |
|---|---|---|
| Lines 73–75 | Catches all errors on load, warns | Catch `FileNotFoundError`, `IOError` only |
| Lines 310–313 | Catches all merge errors, returns partial `ds` | Catch `KeyError`, `ValueError` only |
| Lines 351–353 | Catches all metadata errors, returns empty DataFrame | Catch `KeyError`, `AttributeError` only |
| Lines 376–378 | Catches all query errors, returns empty DataFrame | Catch `pd.errors.UndefinedVariableError` only |
| Lines 495–498 | Swallows all filesystem errors silently | Remove try/except, let errors propagate |
 
This change will make the codebase safer, more debuggable, and better aligned with the principle of **failing loudly** when something unexpected goes wrong.

khushthecoder:
## Fix #1160: Replace broad `except Exception` with specific exception types in `phenotypes.py`

**Fixes:** #1160

### Summary

`phenotypes.py` had 5 instances of `except Exception` that silently converted real bugs (TypeError, KeyError, AttributeError, etc.) into warnings, returning empty or partial data. This is dangerous for scientific analysis where data correctness is critical.

### Changes

| Site | Fix |
|------|-----|
| `_load_phenotype_data()` outer loop | Narrowed to `(KeyError, FileNotFoundError, IOError)` |
| `_create_phenotype_dataset()` variant merge | Removed redundant catch-all — `KeyError` and `ValueError` already caught above |
| `phenotype_data()` metadata fetch | Removed `try/except` — let `sample_metadata()` errors propagate naturally |
| `phenotype_data()` query evaluation | Narrowed to `(pd.errors.UndefinedVariableError, SyntaxError)` |
| `phenotype_sample_sets()` path check | Narrowed to `(KeyError, FileNotFoundError)` |

### Design rationale

- **Fail loudly on unexpected errors** — a `TypeError` or `AttributeError` is a real bug that should crash with a traceback, not get silently converted into a warning
- **Keep expected error handling** — `FileNotFoundError` when a sample set has no phenotype data, `KeyError` when a sample set has no release mapping, and `UndefinedVariableError` for bad user queries are all expected and still handled gracefully
- **Net negative lines** — the fix removes more code than it adds (6 insertions, 17 deletions)

### DIFF
--- a/malariagen_data/anoph/sample_metadata.py
+++ b/malariagen_data/anoph/sample_metadata.py
@@ -880,6 +880,7 @@ class AnophelesSampleMetadata(AnophelesBase):
         sample_sets: Optional[base_params.sample_sets] = None,
         sample_query: Optional[base_params.sample_query] = None,
         sample_query_options: Optional[base_params.sample_query_options] = None,
+        sample_indices: Optional[base_params.sample_indices] = None,
         index: Union[str, Sequence[str]] = (
             "country",
             "admin1_iso",
@@ -894,6 +895,7 @@ class AnophelesSampleMetadata(AnophelesBase):
             sample_sets=sample_sets,
             sample_query=sample_query,
             sample_query_options=sample_query_options,
+            sample_indices=sample_indices,
         )
 
         # Create pivot table.
@@ -928,6 +930,7 @@ class AnophelesSampleMetadata(AnophelesBase):
         sample_sets: Optional[base_params.sample_sets] = None,
         sample_query: Optional[base_params.sample_query] = None,
         sample_query_options: Optional[base_params.sample_query_options] = None,
+        sample_indices: Optional[base_params.sample_indices] = None,
         basemap: Optional[map_params.basemap] = map_params.basemap_default,
         center: map_params.center = map_params.center_default,
         zoom: map_params.zoom = map_params.zoom_default,
@@ -947,6 +950,7 @@ class AnophelesSampleMetadata(AnophelesBase):
             sample_sets=sample_sets,
             sample_query=sample_query,
             sample_query_options=sample_query_options,
+            sample_indices=sample_indices,
         )
 
         # Pivot taxa by locations.
@@ -1256,6 +1260,7 @@ class AnophelesSampleMetadata(AnophelesBase):
         sample_sets: Optional[base_params.sample_sets] = None,
         sample_query: Optional[base_params.sample_query] = None,
         sample_query_options: Optional[base_params.sample_query_options] = None,
+        sample_indices: Optional[base_params.sample_indices] = None,
         template: plotly_params.template = "plotly_white",
         width: plotly_params.fig_width = 800,
         height: plotly_params.fig_height = 600,
@@ -1268,6 +1273,7 @@ class AnophelesSampleMetadata(AnophelesBase):
             sample_sets=sample_sets,
             sample_query=sample_query,
             sample_query_options=sample_query_options,
+            sample_indices=sample_indices,
         )
 
         # Special handling for plotting by year.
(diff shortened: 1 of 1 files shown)


jonbrenas:
Thanks @khushthecoder, `phenotypes.py` is a work in progress (or maybe dead code) as the phenotypic data doesn't exist yet. Making it work better is thus not a priority but I will look at your PR as it may come in handy later if the code actually gets used.

jonbrenas (approved):

Issue #706, opened 2025-01-14
Inclusion of Anopheles stephensi API


leehart:

kebarr:
Closing based on merged [PR 1257](https://github.com/malariagen/malariagen-data-python/pull/1257)

PR #1257, opened 2026-03-31
Add As1 support

tristanpwdennis:
See #706


tristanpwdennis:
Good spot - thanks @adilraza99, I will fix these

tristanpwdennis:
@jonbrenas @ahernank - testing has hit an issue. As discussed we didn't generate contamination QC data for the samples. The tests fail as they are expecting comtam_pc and contam_llc columns.

Rather than restructure the tests, I've added dummy data to the columns in the fixture. My rationale for this is that in the future, we will be able to include contamination data in sample set QC. Let me know if there are any issues around this.

jonbrenas:
Hi @tristanpwdennis, yes, adding dummy data is the correct way to go. To be fair, all the data in the fixture is dummy data.

tristanpwdennis:
Looks like tests are passing (for now)

malariagen_data/as1.py
```diff
@@ -0,0 +1,237 @@
...
+            config_path=CONFIG_PATH,
+            cohorts_analysis=cohorts_analysis,
+            aim_analysis=None,
+            aim_metadata_dtype=None,
+            aim_ids=None,
+            aim_palettes=None,
+            site_filters_analysis=site_filters_analysis,
+            discordant_read_calls_analysis=discordant_read_calls_analysis,
+            default_site_mask="stephensi",
+            default_phasing_analysis="stephensi",
```
adilraza99: I noticed this while going through the phasing setup, `default_phasing_analysis` is set to `"stephensi"`, but the config has `PHASING_ANALYSIS_IDS: []`.

In that case, calling `haplotypes()` with the default would skip validation and eventually hit a `FileNotFoundError` when trying to access phasing data.

In Amin1, this is set to `None`, which makes it fail earlier with a clearer signal that phasing isn’t available. Would something similar make sense here?

tristanpwdennis: Another good spot, thank you @adilraza99

tristanpwdennis:
Hmm. Unable to reproduce this offline. Will investigate.

adilraza99:
Hi @tristanpwdennis,

It looks like the CI failure in `test_allele_frequencies_with_dict_cohorts[adir1_sim]` might be happening when `aa_allele_frequencies()` returns an empty DataFrame (i.e. no amino acid change SNPs for the selected transcript). In that case, the result eventually reaches `np.isnan`, which then raises the TypeError.

I noticed a few other test blocks in the same file already guard this case, for example [L437](https://github.com/malariagen/malariagen-data-python/blob/38fe863929e9c3f112f5aa50bb6469c4dc921899/tests/anoph/test_snp_frq.py#L437), [L572](https://github.com/malariagen/malariagen-data-python/blob/38fe863929e9c3f112f5aa50bb6469c4dc921899/tests/anoph/test_snp_frq.py#L572), [L653](https://github.com/malariagen/malariagen-data-python/blob/38fe863929e9c3f112f5aa50bb6469c4dc921899/tests/anoph/test_snp_frq.py#L653).

It seems this block (around [L709–L716](https://github.com/malariagen/malariagen-data-python/blob/38fe863929e9c3f112f5aa50bb6469c4dc921899/tests/anoph/test_snp_frq.py#L709-L716)) does not have that check yet. Maybe it would make sense to handle it the same way here?

tristanpwdennis:
@adilraza99 - thank you! I think this issue may be random and unrelated to As1 class - probably something to do with the annotations in the dirus assembly but we should open a separate issue about this and possibly implement the checks you suggest.

jonbrenas (changes requested):
I don't think that the fixture needs the curation or the cohorts for all the sample sets, only the 4 (which is already a lot, Ag has 3) that are actually listed.

Also only 1365-VO-DJ-ADBI-VMF00318 has surveillance flags (and the file is empty) which sounds like a missed opportunity to test that they work fine.

The rest seems correct. @tristanpwdennis, did you test every function in the `rst`? Because the functions are tested on simulated data, some functions that have their input from GCS directly fail even when all tests pass.

tristanpwdennis:
Thanks @jonbrenas!

I've had a tidy of the the fixture data - deleting the ghost surv flags (see sep discussion). I also added a flag allowing the dip clustering tests to be performed on classes without cnv data (eg as1, adir1, amin1). Think the .rst is covered now.

README.md
```diff
@@ -1,89 +1,15 @@
-# `malariagen_data` - analyse MalariaGEN data from Python
+# Curation metadata
```
jonbrenas: What happened here?

tristanpwdennis: No idea but it's been banished now

malariagen_data/as1.py
```diff
@@ -0,0 +1,237 @@
...
+        If True, use ipinfo to check the location of the client system.
+    **kwargs
+        Passed through to fsspec when setting up file system access.
+
+    Examples
+    --------
+    Access data from Google Cloud Storage (default):
+
+        >>> import malariagen_data
+        >>> adir1 = malariagen_data.As1()
```
jonbrenas: `adir1` should be `as1`. Same on the next line.

jonbrenas (dismissed):
Only a few copy/paste issues left, I think.

tristanpwdennis:
Is there anything outstanding left on this @jonbrenas and @ahernank ?

PR #1278, opened 2026-04-08
refactor: migrate karyotype tag loading to GCS and add simulated tests

### LINKED ISSUE #689: Refactoring `karyotype`
@cclarkson agreed with me that it might be a good idea to refactor `karyotype` so I move my comments from #652 to here.

The main issue, in my opinion, is that `karyotype` is a very special case because its code is in `ag3.py` (because we only have the inversion tags for ag), which means that its tests are in the legacy section (that we would like to get rid of). It also doesn't show up in the docs for the API. My instinct would be to rework it so that:

1. The tags are stored on gcs instead of within a `resources` folder of `malariagen_data`, this would enable us to have them more readily available to users/partners, to version and update them (as we do with site_filters, for instance) as they get refined with more data (and the notebook already shows that they can be improved), and to have the structure in place if and when a similar list of tags is generated for funestus (which definitely has inversions), stephensi and darlingi (which probably have some inversions as well).
2. The function is moved to `anoph`, same reason as above for future-proofing, even if for a while the function is hidden in funestus or even returns a `NotImplementedError: the tags for the inversion {inversion} have not yet been generated`   
3. It actually shows up in the docs
4. It has tests on simulated data to be compliant with the rest of the package  
5. I think it would make sense to have multiple inversions in the same heatmap (the question then is do we modify `karyotype` to take a list of inversions or do we loop the function in the `karyotype_frequencyies[_advanced]" functions)

Most of that should be a different issue but the frequency functions are going to go in the same folder, if not file, as the base function, in my opinion.

Any opinions?

adilraza99:
This PR implements the karyotype refactor outlined in #689.

The main change is aligning karyotype tag loading with existing data access patterns used across the codebase (e.g., `site_filters`). Instead of loading a bundled CSV via `importlib.resources`, tag SNP data is now accessed from GCS using a config-driven analysis version and the shared filesystem interface.

Changes:
- Replace `importlib.resources`-based loading with GCS-backed loading via `self._fs`
- Introduce `DEFAULT_KARYOTYPE_ANALYSIS` config key with optional constructor override
- Move tag data path to `{version}/karyotype/{analysis}/karyotype_tag_snps.csv`
- Add explicit validation for inversion names (raises `ValueError` for unknown inputs)
- Use contig from tag data directly instead of `inversion[0:2]` string slicing
- Add simulated tag SNP generation (`init_karyotype_tags`) and test coverage under `tests/anoph/`
- Remove bundled `karyotype_tag_snps.csv` from package resources
- Document `load_inversion_tags` in `Ag3.rst`

Note:
Simulated unit tests pass independently. Integration tests may require the tag SNP data to be available on GCS along with the corresponding config update.

Fixes #689


adilraza99:
Hi @jonbrenas, I've made the changes. Take a look when you have a moment.

tests/anoph/conftest.py
```diff
@@ -1150,6 +1154,7 @@ def init_config(self):
             "SITE_ANNOTATIONS_ZARR_PATH": "reference/genome/agamp4/Anopheles-gambiae-PEST_SEQANNOTATION_AgamP4.12.zarr",
             "DEFAULT_AIM_ANALYSIS": "20220528",
             "DEFAULT_SITE_FILTERS_ANALYSIS": "dt_20200416",
+            "DEFAULT_KARYOTYPE_ANALYSIS": "20231213",
```
jonbrenas: Where does this value come from?

tests/anoph/conftest.py
```diff
@@ -1519,6 +1524,34 @@ def init_snp_sites(self):
             path=path, contigs=self.contigs, genome=self.genome
         )
 
+    def init_karyotype_tags(self):
+        analysis = self.config["DEFAULT_KARYOTYPE_ANALYSIS"]
+
+        # Generate tag SNP data using positions from simulated SNP sites.
+        tags = []
+        for contig, inversion in [("2R", "2Rb"), ("2L", "2La")]:
```
jonbrenas: Why are the inversions hard coded here? If we ever implement karyotypes for ever species they won't have the same inversions.

malariagen_data/anoph/karyotype.py
```diff
@@ -36,37 +36,59 @@ def _karyotype_tags_n_alt(gt, alts, inversion_alts):
...
-        import importlib.resources
-        from .. import resources
+        self._require_karyotype_analysis()
 
-        if self._inversion_tag_path is None:
-            raise NotImplementedError(
-                "No inversion tags are available for this data resource."
+        path = (
+            f"{self._base_path}/{self._major_version_path}"
+            f"/karyotype/{self._karyotype_analysis}/karyotype_tag_snps.csv"
```
jonbrenas: The actual path would probably be `snp_karyotype` to be consistent with the others (e.g., `snp_haplotypes`). Also, the name of the actual file shouldn't be hard-coded. It should probably be added to the config.

jonbrenas (changes requested):

tests/anoph/conftest.py
adilraza99: This is a simulated placeholder used in the test config. I’ve added a comment to make that clearer.

tests/anoph/conftest.py
adilraza99: These are defined in the Ag3 simulator, so they are specific to that dataset. When karyotype support is added for other species, their simulators would define their own inversions, similar to how other simulator methods are structured.

malariagen_data/anoph/karyotype.py
adilraza99: Updated the path to follow the `snp_karyotype` convention and made the filename configurable via the config.

adilraza99:
Hi @jonbrenas, I’ve updated this to follow the existing patterns and addressed the review points. Let me know if this looks good.

adilraza99:
Could you please take a look when you have a moment?
@jonbrenas

tests/anoph/conftest.py
jonbrenas: The date 13/12/2023 still seems fairly random.

jonbrenas (changes requested):
Hi @adilraza99, I am not sure why there is an empty MANIFEST.in.

tests/anoph/conftest.py
adilraza99: Good catch , the previous value was arbitrary and ended up looking like a real date, which is confusing. I have replaced it with "simtest" so it is clearly just a test value and not mistaken for an actual analysis version on GCS.

adilraza99:
Hi @jonbrenas, thanks a lot for the review.

I've addressed the changes:
- Replaced the previous date-like value with "simtest" so it’s clearly just a test value
- Removed the empty MANIFEST.in

Please let me know if anything else needs adjustment. Thank you!

PR #1284, opened 2026-04-10
Fix #1283: Warn and handle missing FORMAT fields in snp_calls_to_vcf()

### LINKED ISSUE #1283: snp_calls_to_vcf() silently drops requested FORMAT fields when underlying data arrays don't exist
## Description
When a user requests optional FORMAT fields like GQ, AD, or MQ via the fields parameter of `snp_calls_to_vcf()`, the method silently falls back to writing "." (missing) for every sample if the corresponding data array doesn't exist in the dataset. There is no warning, log message, or exception to inform the user that their requested field is absent from the underlying data.

## Impact
- **Data integrity**: Users may unknowingly base analyses on "empty" FORMAT fields, leading to incorrect quality filtering or allele depth-based conclusions.
- **Misleading VCF headers**: The output file's header declares fields that contain no real data, violating the VCF specification's expectation that declared FORMAT fields carry meaningful values.
- **Debugging difficulty**: When a downstream tool (e.g., `bcftools filter -e 'GQ<20'`) behaves unexpectedly because all GQ values are missing, the user has no indication that the field was silently absent from the source data.

## Steps to Reproduce
1. Open `malariagen_data/anoph/to_vcf.py`, lines 136–150 and 169–183.
2. Note the `except KeyError: pass` pattern — no warnings.
3. Attempt to export VCF with `fields=("GT", "GQ", "AD")` for a sample set that doesn't contain `call_GQ` or `call_AD` arrays.
4. Inspect the output VCF: header declares GQ and AD, but all values are ".".

## Proposed Solution
Add `warnings.warn()` calls when a requested field's data array is not found.
Additionally, ensure the VCF header only declares fields that actually have data available in the dataset.

## Impact After Resolution
- Users receive clear feedback when requested FORMAT fields are absent, enabling informed decision-making.
- VCF output integrity improves — headers accurately reflect available data.
- Consistent with the project's existing `warnings.warn()` pattern for missing metadata files.
- Prevents silent downstream analysis errors.

Yashsingh045:
## Description
This PR fixes an issue #1283 , 
 In the snp_calls_to_vcf() method where requested FORMAT fields (GQ, AD, MQ) were silently dropped if the underlying data was missing or unreadable. It adds user-facing warnings and ensures VCF headers are dynamically updated to accurately reflect available data.

## Problem Statement
When a user explicitly requested optional FORMAT fields via the fields parameter, the exporter would silently fallback to missing values (.) if the data arrays were absent from the dataset or if computation failed for specific chunks.

## Impact: 
This led to "silent failure" where downstream tools would behave unexpectedly due to empty fields, and VCF files were generated with misleading headers that declared fields which contained no real data.

## Solution / Changes
- **Explicit Warnings**: Added warnings.warn() calls using the standard UserWarning category to notify users when a requested field is omitted due to missing data.
- **Header Filtering**: Reorganized the initialization logic to extract data arrays before writing the VCF header. This allows the tool to filter the output_fields list so that only fields with valid data arrays are declared in the ##FORMAT header lines.
- **Fail-Safe Computation**: Updated the chunk iteration loop to emit a warning if specific data blocks cannot be read (e.g., due to missing Zarr files on disk), while maintaining VCF structure by writing sentinel . values.
- **Imports**: Added import warnings to malariagen_data/anoph/to_vcf.py.

## Closes
Fixes #1283

### DIFF
--- a/malariagen_data/anoph/to_vcf.py
+++ b/malariagen_data/anoph/to_vcf.py
@@ -3,6 +3,7 @@ import os
 from datetime import date
 from typing import Optional
 
+import warnings
 import numpy as np
 from numpydoc_decorator import doc  # type: ignore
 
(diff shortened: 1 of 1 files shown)


Yashsingh045:
Hey @jonbrenas ,
Could you please review this PR when you have time?
Thanks

Yashsingh045:
Hey @jonbrenas ,
Just a gentle reminder, 
could you please review this PR when you have time?
Thanks

malariagen_data/anoph/to_vcf.py
```diff
@@ -95,11 +96,60 @@ def snp_calls_to_vcf(
...
+            except KeyError:
+                warnings.warn(
+                    "Requested FORMAT field 'MQ' not found in dataset. "
+                    "MQ values will be omitted from the output VCF.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                output_fields.remove("MQ")
+
+        include_gq = "GQ" in output_fields
```
jonbrenas, line 149: These should come before the if-statement.

jonbrenas (changes requested):
Also, Copilot has some comments.

PR #1306, opened 2026-04-15
fix: restore original recursion limit after dendrogram computation

### LINKED ISSUE #1305: sys.setrecursionlimit(10_000) in hapclust/dipclust Permanently Alters Global Python State
## Description
 
Both `hapclust.py:89` and `dipclust.py:92` call `sys.setrecursionlimit(10_000)` as a side effect of plotting dendrograms, but never restore the original limit. This permanently modifies the global Python recursion limit for the entire process, affecting all subsequent code — not just `malariagen_data`.
 
---
 
## How to Reproduce
 
```python
import sys
print(sys.getrecursionlimit())  # 1000 (default)
 
import malariagen_data
ag3 = malariagen_data.Ag3()
 
# Call any function that triggers hapclust or dipclust:
ag3.plot_haplotype_clustering(region="2L:28,545,000-28,550,000", analysis="gamb_colu")
 
print(sys.getrecursionlimit())  # 10000 — permanently changed!
 
# Now ALL code in this notebook/session uses the elevated limit,
# potentially masking stack overflow bugs in unrelated code.
```
 
---
 
## Why It Is Important
 
- **Side-effect pollution:** A plotting function should not permanently modify global interpreter state. Users working in long-running Jupyter sessions have no idea their recursion limit changed.
- **Masks bugs:** Legitimate `RecursionError` bugs in user code or other libraries will be silently suppressed because the limit is now 10x higher.
- **Memory risk:** A deeper call stack consumes more memory per thread. In environments with many threads (e.g., dask distributed workers), this compounds.
---
 
## Proposed Fix
 
The fix is trivial — wrap the recursion limit change in a `try/finally` block to restore the original limit:
 
```python
# Before (current behaviour)
sys.setrecursionlimit(10_000)
# ... dendrogram computation ...
 
# After (proposed fix)
_original_limit = sys.getrecursionlimit()
try:
    sys.setrecursionlimit(10_000)
    # ... dendrogram computation ...
finally:
    sys.setrecursionlimit(_original_limit)
```
 
This pattern should be applied in both:
- `hapclust.py` line 89
- `dipclust.py` line 92
---
 
The recursion limit is temporarily elevated only for the duration of the dendrogram computation and immediately restored afterward, eliminating global state pollution.

kunal-10-cloud:
## Summary

- Wrapped `sys.setrecursionlimit(10_000)` calls in `hapclust.py` and `dipclust.py` with `try/finally` blocks to restore the original recursion limit after dendrogram computation
- Previously, calling `plot_haplotype_clustering()` or `plot_diplotype_clustering()` permanently elevated the global Python recursion limit from 1,000 to 10,000 for the entire process
- The original limit is now saved before elevation and restored in a `finally` block, ensuring cleanup even if an exception occurs

## Problem

Both `hapclust.py` (line 89) and `dipclust.py` (line 92) called `sys.setrecursionlimit(10_000)` as a side effect of plotting dendrograms but never restored the original limit. This:

1. **Pollutes global state** — a plotting function permanently modifies interpreter behaviour for the entire session
2. **Masks bugs** — legitimate `RecursionError` bugs in user code or other libraries are silently suppressed because the limit is 10x higher
3. **Increases memory risk** — deeper call stacks consume more memory per thread, compounding in multi-threaded environments (e.g., dask distributed workers)

## Changes

**`malariagen_data/anoph/hapclust.py`**
- Save `sys.getrecursionlimit()` before elevation
- Wrap all computation in `try/finally` to guarantee restoration

**`malariagen_data/anoph/dipclust.py`**
- Same pattern applied

## Test plan

- [x] All 36 hapclust/dipclust tests pass (`test_hapclust.py`, `test_dipclust.py`)
- [x] Full test suite passes (1041 passed, 8 skipped)
- [x] `ruff check` passes with no issues
- [x] `ruff format` applied — no formatting violations

Fixes #1305


kunal-10-cloud:
Hi @jonbrenas 

Just a friendly nudge on this PR  whenever you get a chance!

Quick summary: plot_haplotype_clustering() and plot_diplotype_clustering() currently call sys.setrecursionlimit(10_000) but never restore the original limit. This means a single plotting call permanently elevates the recursion limit from 1,000 to 10,000 for the entire Python session — silently masking RecursionError bugs in user code and increasing memory risk in multi-threaded environments (e.g., dask workers).

The fix simply wraps the existing code in try/finally to restore the original limit after computation. No new dependencies, no API changes, no behaviour change for the dendrogram itself.

It's a small, self-contained fix — just two files changed with the same pattern applied to each. Full test suite passes (1041 passed). Should be a quick review!

Happy to address any feedback. Thanks!

jonbrenas (approved):

PR #1309, opened 2026-04-17
Fix #1308: Remove dead dask.config.set() at module import time

### LINKED ISSUE #1308: dask.config.set() at Module Import Time in ag3.py Silently Modifies Global Dask Configuration
## Description:
 
`ag3.py` line 10 executes the following statement at **module import time**:
 
```python
dask.config.set(**{"array.slicing.split_native_chunks": False})
```
 
This runs the moment any code imports `malariagen_data`, since `ag3.py` is loaded during package initialization. It silently disables a Dask performance optimization **globally**, affecting all Dask operations in the entire Python process — not just those belonging to `malariagen_data`.
 
---
 
## Steps to Reproduce
 
```python
import dask
print(dask.config.get("array.slicing.split_native_chunks"))  # True (default)
 
import malariagen_data  # Side effect triggered here!
 
print(dask.config.get("array.slicing.split_native_chunks"))  # False — globally changed
 
# Now ANY dask array operation in this process uses non-split behavior,
# even operations completely unrelated to malariagen_data:
import dask.array as da
x = da.random.random((10000, 10000), chunks=(1000, 1000))
result = x[::2]  # Uses non-split slicing due to the global config change
```
 
---
 
## Root Cause
 
The call to `dask.config.set()` is made at **module scope** (top-level of `ag3.py`), rather than inside a context manager scoped to the specific operations that require it. As a result, importing `malariagen_data` is sufficient to permanently alter the Dask configuration for the entire Python session.
 
---
 
### 1. Unexpected Side Effects
Importing a library should not modify the global configuration of other libraries. This violates the **principle of least surprise** and breaks the assumption that `import` statements are side-effect-free.
 
### 2. Performance Impact
`split_native_chunks` was introduced to improve performance for certain slicing operations. Disabling it globally may **degrade performance** for non-`malariagen_data` Dask workloads running in the same session (e.g., `xarray`, `pangeo`, custom pipelines).
 
### 3. Difficult to Debug
Researchers combining `malariagen_data` with other Dask-based tools will observe changed Dask behaviour with **no visible indication of the cause**. The configuration change is silent, persistent, and attributed to nothing in tracebacks or logs.
 
---
 
## Expected Behavior
 
The `dask.config.set()` override should be scoped **only to the specific operations that require it**, using the context manager form:
 
```python
with dask.config.set(**{"array.slicing.split_native_chunks": False}):
    # perform the specific operation that needs this config
    ...
```
 
This ensures the configuration is restored to its previous value immediately after the operation completes, with no lasting effect on the rest of the process.
 
---
 
## Proposed Fix
 
**Before (current behaviour — `ag3.py`, line 10):**
 
```python
# Module-level — runs on import, permanently changes global config
dask.config.set(**{"array.slicing.split_native_chunks": False})
```
 
**After (proposed fix — scoped to specific methods):**
 
```python
# Inside the method(s) that need this behaviour
def some_method(self, ...):
    with dask.config.set(**{"array.slicing.split_native_chunks": False}):
        # operation that requires this config
        ...
```
 
---
 
- Importing `malariagen_data` will no longer modify the global Dask configuration.
- The `split_native_chunks` override will be active **only** during the specific operations that require it.
- Third-party Dask-based tools used alongside `malariagen_data` (e.g., `xarray`, `pangeo`) will no longer have their performance characteristics silently altered.
- The fix is **fully backward-compatible** — no changes to public API or user-facing behaviour are required.
---
 
## Affected File
 
| File | Line | Code |
|------|------|------|
| `ag3.py` | 10 | `dask.config.set(**{"array.slicing.split_native_chunks": False})` |
 
---

khushthecoder:
Closes #1308.

## Summary

`ag3.py:10` executed the following at **module import time**:

```python
# silence dask performance warnings
dask.config.set(**{"array.slicing.split_native_chunks": False})
```

This line was added in an older dask era to silence a `PerformanceWarning` emitted by certain slicing operations. On current dask it is dead code — removing it directly fixes the bug (`import malariagen_data` no longer mutates global dask configuration).

## Why just remove the line

Verified empirically on dask `2025.11.0` (what the project resolves to since `dask = "*"` in `pyproject.toml`):

- `dask.config.get("array.slicing.split_native_chunks", "DEFAULT")` returns `"DEFAULT"` — the config key is no longer recognised.
- `from dask.array.slicing import PerformanceWarning` raises `ImportError` — the warning class has been removed.
- `da.take` / `da.compress` / fancy indexing with out-of-order indices emit **zero warnings** with the config set to `True`, `False`, or unset.

So the line suppresses a warning that no longer exists, gated by a config key that no longer exists. Scoping it to context managers (the previous approach in this PR) preserved dead code unnecessarily; deleting it is the cleaner fix.

## Diff

1 file changed, 4 deletions.

```diff
-import dask
 import pandas as pd
 ...
-# silence dask performance warnings
-dask.config.set(**{"array.slicing.split_native_chunks": False})
```

## Test plan

- [x] `python -c "import malariagen_data; import dask; assert dask.config.get('array.slicing.split_native_chunks', 'DEFAULT') == 'DEFAULT'"` — passes (import no longer mutates dask config).
- [x] `pytest tests/anoph/test_snp_data.py` — 193 passed, no `PerformanceWarning` emitted.
- [x] `pytest tests/anoph/test_frq.py tests/anoph/test_cnv_data.py tests/anoph/test_hapclust.py tests/anoph/test_dipclust.py tests/anoph/test_hap_data.py tests/anoph/test_aim_data.py` — 133 passed.
- [x] `pre-commit run --files malariagen_data/ag3.py` — ruff, ruff-format, trailing-whitespace, end-of-file all pass.
- [x] `mypy malariagen_data/ag3.py` — no issues.
- [ ] CI matrix on PR (linting, coverage, tests across Python 3.10 / 3.11 / 3.12 × numpy `==2.0.2` and `>=2.0.2,<2.1`).

## Backwards compatibility

- No public API changed.
- If anyone pins a very old dask version where the warning still fires, they may see it again; the project does not currently pin a minimum dask version in `pyproject.toml` (`dask = "*"`), so modern dask is the supported target.

### DIFF
--- a/malariagen_data/ag3.py
+++ b/malariagen_data/ag3.py
@@ -1,13 +1,9 @@
 import sys
 
-import dask
 import pandas as pd  # type: ignore
 import plotly.express as px  # type: ignore
 import malariagen_data
 from .anopheles import AnophelesDataResource
-
-# silence dask performance warnings
-dask.config.set(**{"array.slicing.split_native_chunks": False})  # type: ignore
 
 MAJOR_VERSION_NUMBER = 3
 MAJOR_VERSION_PATH = "v3"


khushthecoder:
Hi @jonbrenas — gentle follow-up. I've force-pushed a simpler approach to this PR than the
original scoped-context-manager version; quick summary so the change isn't confusing.

**New approach:** just deletes the 4 lines in `ag3.py`. No scoped context managers added anywhere.

**Why this turned out to be the right shape:** on current dask (`2025.11.0`, what `poetry.lock`
resolves to), the config key `array.slicing.split_native_chunks` no longer exists in dask's
schema — `grep` across the installed dask package returns zero matches, and
`from dask.array.slicing import PerformanceWarning` raises `ImportError`. So the original line
in `ag3.py` was silently setting a dead config key to suppress a warning class that no longer
exists. Wrapping dead code in context managers (the earlier approach) felt like overkill once I
verified that, so the PR is now just the deletion.

**Diff:** 1 file, 4 deletions, 0 additions.

**Verified locally:**

- `import malariagen_data` no longer mutates `dask.config`.
- `pytest tests/anoph/test_snp_data.py` (193 tests) plus `test_frq`, `test_cnv_data`,
  `test_hapclust`, `test_dipclust`, `test_hap_data`, `test_aim_data` — all pass, zero
  `PerformanceWarning` emitted from any `da.take` / `da.compress` call site.
- pre-commit + ruff + mypy clean.

CI is green on the new commit across all 10 checks. No rush on review — whenever it fits
your queue.

PR #1314, opened 2026-04-18
Make the sex-linked contig configurable in CNV frequency analysis

### LINKED ISSUE #1313: cnv_frq.py Hardcodes X-Chromosome Sex-Linked Ploidy Assumption — Breaks CNV Frequency Analysis for Non-Gambiae Species
## Description
 
The CNV frequency analysis in `gene_cnv_frequencies()` and `gene_cnv_frequencies_advanced()` contains a hardcoded assumption at `cnv_frq.py:293` and `cnv_frq.py:560`:
 
```python
if region.contig == "X":
    is_male = (df_samples["sex_call"] == "M").values
    expected_cn = np.where(is_male, 1, 2)[np.newaxis, :]
else:
    expected_cn = 2
```
 
This assumes:
 
- The sex-linked chromosome is always named `"X"` — true for *An. gambiae* (Ag3) but not guaranteed for other species assemblies.
- Male mosquitoes are always hemizygous (CN=1) on the sex chromosome — true for gambiae complex but *An. farauti* and other Pacific vectors may use different chromosome naming or sex-determination architecture.
- The `sex_call` column is always present — As1 samples may not have `sex_call` populated.
As MalariaGEN expands to Pacific vectors (*An. farauti*) for GSoC Project 1, this hardcoding will silently produce incorrect amplification/deletion frequencies for any species whose genome assembly uses different contig naming conventions.
 
---
 
## Why It Is Important
 
- **CNV frequencies directly inform insecticide resistance surveillance** — incorrect values lead to wrong public health decisions.
- **GSoC Project 1 explicitly requires** building cloud-native analysis functions for *An. farauti* that are interoperable with existing resources.
- This is a **correctness bug**, not a style issue.
---
 
## Steps to Reproduce
 
1. Examine `malariagen_data/anoph/cnv_frq.py` lines `293` and `560`.
2. Note the hardcoded `"X"` string comparison.
3. Attempt to run `gene_cnv_frequencies()` on a species where the sex chromosome is not named `"X"` — the expected copy number will default to `2` for all samples regardless of sex.
---
 
## Proposed Approach
 
**1. Add a `sex_contig` property to the species configuration.**
 
Each species class defines which contig is sex-linked:
 
```python
# e.g., in the species-specific resource class
@property
def _sex_contig(self) -> str:
    return "X"  # override per species as needed
```
 
**2. Replace the hardcoded string comparison.**
 
```python
# Before (current behaviour)
if region.contig == "X":
    ...
 
# After (proposed fix)
if region.contig == self._sex_contig:
    ...
```
 
**3. Add a safe fallback when `sex_call` is missing.**
 
Rather than crashing or silently computing incorrect values, log a warning and default to diploid:
 
```python
if "sex_call" not in df_samples.columns:
    warnings.warn(
        "sex_call column not found; defaulting to diploid (CN=2) for all samples.",
        UserWarning,
    )
    expected_cn = 2
else:
    is_male = (df_samples["sex_call"] == "M").values
    expected_cn = np.where(is_male, 1, 2)[np.newaxis, :]
```
 
**4. Pass `sex_contig` through `AnophelesDataResource.__init__()`** alongside existing species-specific params like `virtual_contigs`:
 
```python
def __init__(
    self,
    ...,
    sex_contig: str = "X",  # new param, defaults to "X" for backward compatibility
    virtual_contigs: Optional[Mapping] = None,
    ...
):
    self._sex_contig = sex_contig
    ...
```
 
---
 
## Expected Outcome After Fix
 
- CNV frequency analysis produces correct sex-adjusted copy numbers for any *Anopheles* species, not just gambiae.
- New species (*An. farauti*, *An. darlingi*) can be onboarded by simply specifying their sex contig name in the constructor.
- No behavior change for existing Ag3 / Af1 / As1 users.
- Missing `sex_call` data raises an explicit warning rather than silently returning incorrect frequencies.

khushthecoder:
## Summary

Fix #1313.

`gene_cnv_frequencies()` and `gene_cnv_frequencies_advanced()` hardcoded `region.contig == "X"` when deciding the expected copy number per sample. That is correct for every currently supported assembly, but bakes in a gambiae-complex assumption: any future species whose sex-linked contig uses a different name would silently fall through to the diploid branch and report incorrect amplification / deletion frequencies for hemizygous males.

Because CNV frequencies directly inform insecticide resistance surveillance, silent misclassification of the sex-linked contig is a correctness issue, not a cosmetic one.

### Configurable `sex_contig`

- Added a `sex_contig` constructor parameter to `AnophelesGenomeSequenceData`, defaulting to `"X"`, and exposed a matching public `sex_contig` property. This mirrors the existing pattern used for `virtual_contigs`.
- Threaded `sex_contig` through `AnophelesDataResource.__init__` so species subclasses can override it at construction time without touching the frequency analysis code.

### Single source of truth for expected copy number

- Replaced both hardcoded `"X"` sites in `cnv_frq.py` with a shared `_expected_copy_number()` helper, which consults `self._sex_contig` and encapsulates all ploidy logic in one place.

### Explicit handling of missing `sex_call`

Previously, requesting the sex-linked contig with a sample frame that happened to lack `sex_call` (e.g. an As1-style resource) would raise a bare `KeyError` from deep inside pandas. Now it emits a `UserWarning` and falls back to diploid for all samples, which is the safest interpretation when sex is unknown. Users get a clear message rather than an opaque crash, and the fallback behaviour is documented in the helper's docstring.

## Backwards compatibility

- Default `sex_contig="X"` keeps behaviour identical for Ag3, Af1, As1, and every other existing resource.
- No public API removed or renamed.
- The new `sex_contig` property is additive.
- The new `UserWarning` only fires on a code path that previously raised `KeyError`, so nothing that used to succeed now warns.

## Test plan

- [x] `pre-commit run --files …` — passed (ruff, ruff-format, trailing-whitespace, end-of-file)
- [x] `poetry run mypy …` — passed (no issues)
- [x] `poetry run pytest tests/anoph/test_cnv_frq.py` — 68 passed
- [x] `poetry run pytest tests/anoph/test_genome_sequence.py tests/anoph/test_base.py` — 62 passed
- [x] `poetry run pytest tests/anoph/test_dipclust.py` — 64 passed (exercises `AnophelesCnvFrequencyAnalysis` via the diplotype clustering path)
- [ ] CI matrix on PR (linting, coverage, tests across Python 3.10 / 3.11 / 3.12 × numpy `==2.0.2` and `>=2.0.2,<2.1`)

## Files

| File | Change |
| --- | --- |
| `malariagen_data/anoph/genome_sequence.py` | Add `sex_contig` init param + public property. |
| `malariagen_data/anopheles.py` | Thread `sex_contig` through `AnophelesDataResource`. |
| `malariagen_data/anoph/cnv_frq.py` | Introduce `_expected_copy_number()` helper; remove hardcoded `"X"`; add `sex_call` fallback warning. |


khushthecoder:
Hi @jonbrenas  — thanks again for merging #1304.

When you have a spare moment, I'd appreciate a look at this pr as well. It closes #1313 (the hardcoded "X" sex-contig in cnv_frq.py) by threading a configurable sex_contig through AnophelesGenomeSequenceData with a default of "X", so behaviour is unchanged for Ag3/Af1/As1 but non-gambiae assemblies can opt in. Local checks pass and CI is green. No rush — whenever it fits your queue.

khushthecoder:
Hi @jonbrenas — gentle follow-up on this one whenever you get a chance.

Quick status:

All CI checks are green (linting, coverage job, and tests across Python 3.10 / 3.11 / 3.12 × both numpy pins).
The diff is small and fully backwards-compatible: sex_contig defaults to "X", so Ag3, Af1, As1, and every existing resource keep identical behaviour. Non-gambiae assemblies can simply override it at construction time.
Codecov flagged patch coverage at 78.57% (target 80%). The uncovered lines are the defensive branches inside _expected_copy_number() — happy to add a small unit test targeting them directly if you'd prefer the patch to clear the threshold before merge, just let me know.
No rush at all — thanks for your time.

jonbrenas:
I don't think there is a species for which the sex chromosomes are known and not named 'X' and 'Y'.

khushthecoder:
> I don't think there is a species for which the sex chromosomes are known and not named 'X' and 'Y'.

Thanks for the quick reply, @jonbrenas — that's a fair point, and I agree: for every assembly
this library currently targets (and realistically any that would ever ship here), the sex
chromosome is `"X"`. So the "future non-X contig" framing in the original PR description is
weaker than I made it out to be.

That said, there are two pieces in the diff that stand on their own regardless of the
configurability:

1. **`sex_call` safety guard** — the current code accesses `df_samples["sex_call"]`
unconditionally inside the X branch (`cnv_frq.py:294` and `:561`). If a sample set ever lacks
that column — plausible for a newer species where sex calling hasn't been run yet — it surfaces
as a bare `KeyError` from inside pandas. The PR turns that into an explicit `UserWarning` with a
diploid (`CN=2`) fallback, which I think is the safer default when sex is unknown.

2. **Single source of truth** — the literal `region.contig == "X"` is duplicated across two
methods (`cnv_frq.py:293` and `:560`). Extracting a small `_expected_copy_number()` helper
removes the duplication and makes the ploidy logic explicit in one place, without adding real
complexity.

Since `sex_contig` defaults to `"X"`, there's **zero behaviour change** for Ag3, Af1, or As1 —
existing notebooks and downstream code keep working identically.

Happy to go in whichever direction you prefer:

- Keep only **(1) and (2)** — drop the `sex_contig` constructor parameter and public property
  entirely, leaving just the helper refactor and the `sex_call` guard. Much smaller diff.
- **Keep as-is** if the configurability is worth having as defensive scaffolding.
- **Close the PR** if you don't think the motivation justifies the surface-area change — totally
  your call, no hard feelings.

Let me know which you'd like and I'll push the change (or close it) today.

PR #1316, opened 2026-04-18
fix: classify in-frame complex variants (MNP+INDEL) as CODON_CHANGE in veff.py

### LINKED ISSUE #1315: veff.py: in-frame complex variants (MNP+INDEL) return literal "TODO" string as effect — invalid scientific output
## Summary

In `malariagen_data/veff.py`, the function `_get_within_cds_effect()` has an incomplete branch for in-frame complex variants — variants that are simultaneously a multi-nucleotide polymorphism (MNP) AND an indel, where the net length change is a multiple of 3 (no frameshift).

## Current broken behaviour

Lines 447-449 currently contain:

    effect = base_effect._replace(
        effect="TODO in-frame complex variation (MNP + INDEL)",
        impact="UNKNOWN"
    )

This means a researcher who calls `snp_effects()` on such a variant 
receives a DataFrame where:
- The `effect` column contains the literal string "TODO in-frame complex variation (MNP + INDEL)"
- The `impact` column contains "UNKNOWN"

## Why this is serious

1. "UNKNOWN" is not a valid impact level anywhere else in this codebase.
   All other values are "HIGH", "MODERATE", "LOW", or "MODIFIER".
   This silently breaks any downstream filtering such as df[df["impact"] == "HIGH"].

2. The TODO string is a developer note leaking directly into scientific output with no warning or error raised.

## How to reproduce

Call snp_effects() on any variant where:
- len(ref) > 1 and len(alt) > 1 (not a simple insertion or deletion)
- len(ref) != len(alt) (not a pure MNP)
- (len(alt) - len(ref)) % 3 == 0 (in-frame, not a frameshift)

Example: ref="GCC", alt="GCCATG" at a CDS position.

## Proposed fix

Replace the TODO branch with:

    effect = base_effect._replace(effect="CODON_CHANGE", impact="MODERATE")

This is consistent with how pure MNPs are already handled in the elif branch directly above. Both cases represent in-frame changes to one or more codons with no frameshift.

## Related

This is related to issue #1180. I will submit a PR with the fix and a regression test.

Tanisha127:
## What this PR does

Fixes the incomplete `else` branch in `_get_within_cds_effect()` in `veff.py` that previously leaked a literal "TODO" developer note and an invalid "UNKNOWN" impact value into scientific output.

Closes #1315

## The problem

When a variant was an in-frame complex variant (MNP + INDEL combined, net length change is a multiple of 3), the code at lines 447-449 returned:

    effect="TODO in-frame complex variation (MNP + INDEL)", impact="UNKNOWN"

This caused two problems:
- A developer TODO string appearing in researcher output DataFrames
- "UNKNOWN" impact is not valid — breaks any downstream filtering like df[df["impact"] == "HIGH"]

## The fix

Replaced the TODO branch with:

    effect="CODON_CHANGE", impact="MODERATE"

This follows the same logic as the elif len(ref) == len(alt) branch directly above (pure MNPs). Both cases are in-frame codon changes with no frameshift, so CODON_CHANGE / MODERATE is the correct SnpEff-compatible classification.

## Changes

- malariagen_data/veff.py: replaced TODO branch with proper classification
- tests/test_veff.py: added TestInFrameComplexVariant class with 3 tests

## Testing

    python -m pytest tests/test_veff.py -v 44 passed in 3.91s


Tanisha127:
Hi @jonbrenas , 
I have opened a PR. Could you please review it? If any changes are required, please let me know.

PR #1319, opened 2026-04-22
Add pydantic Field constraints

### LINKED ISSUE #484: Migrate to pydantic validate_call() for public API parameter type validation
Pydantic v2 has the [validate_call()](https://docs.pydantic.dev/latest/concepts/validation_decorator/) decorator which provides support for validating types of input parameters, we could use this instead of the home-grown check_types() decorator.

While we're at it, we could also make use of pydantic's [Field](https://docs.pydantic.dev/latest/concepts/fields/) to express some additional constraints on parameters, e.g., numeric constraints.

31puneet:
This PR solves the second part of the issue #484
-  Adds `Field()` numeric constraints to 38 parameters across `13 *_params.py files`.
- Adds `test_field_constraints.py` to verify enforcement.

Closes #484

### DIFF
--- a/malariagen_data/anoph/base_params.py
+++ b/malariagen_data/anoph/base_params.py
@@ -2,6 +2,7 @@
 
 from typing import Final, List, Mapping, Optional, Sequence, Tuple, Union
 
+from pydantic import Field
 from typing_extensions import Annotated, TypeAlias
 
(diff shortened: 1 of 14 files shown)


31puneet:
PTAL @jonbrenas, Thanks

Issue #1279, opened 2026-04-08
S3 references in the docstring for Pf8


jonbrenas:
The docstring for the class `Pf8` in `pf8.py` has not been updated and still says that S3 is the default path (though it is no longer true in fact). @nw20, could you correct that?

Tanisha127:
Hey @jonbrenas  
I came across this issue and would love to help if @nw20 needs a hand.

Here is how I am planning to approach it :
Locate the `Pf8` class docstring in `pf8.py`
Check the current default path being used in the actual code
Update the docstring to reflect the correct default path instead of the outdated S3 reference
Do a quick scan of any related docstrings nearby to make sure nothing else is outdated

It is important fix to keep the docs accurate for users. Happy to submit a PR if you'd like just let me know.

PR #1320, opened 2026-04-27
remove s3 from pf8 docstrings

nw20:
For issue #1279 - I have removed mentions of S3 data storage paths from `pf8.py` and the docstring now references GCS only.


jonbrenas (approved):
LGTM

ahernank (approved):

Issue #841, opened 2026-01-04
Pip dependency error during installation


ahernank:
Currently, when installing using pip in Colab, a dependency error rises:

> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> panel 1.8.4 requires bokeh<3.9.0,>=3.7.0, but you have bokeh 3.6.3 which is incompatible.

malariagen_data & other dependencies (inc. bokeh for our current purposes) install successfully, but this error can be misleading when running a quiet installation, e.g. `!pip install -qU malariagen_data`, as the succesful installation message won't print after the error above.

We might want to assess bringing up bokeh's version but this will need testing against the known linting errors in https://github.com/malariagen/malariagen-data-python/issues/734. Alternatively, we could take out the quiet installation from public docs, which might need assessing, particularly against training NBs.

jonbrenas:
Either this has been solved or there is another error upstream that masks it because the error message when using pip in Colab is now:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
google-colab 1.0.0 requires pandas==2.2.2, but you have pandas 2.3.2 which is incompatible.
```

kebarr:
Closing based on merged [PR 1326](https://github.com/malariagen/malariagen-data-python/pull/1326)

PR #1326, opened 2026-05-26
Pip dependency error during installation

jonbrenas:
Addresses #841.

Reverted pandas to previous minor version (2.2.2) to prevent conflict with Colab's basic set up.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3640,54 +3640,41 @@ files = [
 
 [[package]]
 name = "pandas"
-version = "2.3.2"
+version = "2.2.2"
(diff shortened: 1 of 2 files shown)


jwgarber:
Google lists all their colab package versions [here](https://github.com/googlecolab/backend-info/blob/main/pip-freeze.txt), so we can check that what we have in pyproject.toml is compatible.

ahernank (approved):
Thanks @jonbrenas, seems very sensible, as we can tackle 2.3 when we start getting ready to include 3.0.

PR #1329, opened 2026-06-01
handle Mapping cohorts in pca

### LINKED ISSUE #1328: `pca`'s `cohorts` argument does not work the way it is expected to
A `cohorts` parameter was added to `pca` but it is a bit confusing.

Most functions using a `cohorts` parameter, such as the `diversity_stats` which only works at the cohort level or `snp_allele_frequencies` which needs to split the data into cohorts, have a different use for `cohorts` than `pca` which uses it to downsample the data. Arguably, because the 2 parameters are semantically different, they should have different definitions and the current `cohorts` parameter works fine if it is set to a column name and only really fails if given a dictionary (or any other mapping, I would guess).

There are two options: 
- change the `cohorts` parameter to be more restrictive
- change the code to cope with the more general definition of `cohorts`

31puneet:
Fixes #1328

### DIFF
--- a/malariagen_data/anoph/pca.py
+++ b/malariagen_data/anoph/pca.py
@@ -8,6 +8,7 @@ from numpydoc_decorator import doc  # type: ignore
 
 from ..util import CacheMiss, _check_types, _jitter
 from . import base_params, pca_params, plotly_params
+from .sample_metadata import _locate_cohorts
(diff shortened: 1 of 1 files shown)


31puneet:
PTAL @jonbrenas, Thanks

Issue #705, opened 2025-01-14
Admixture stats (f3, f4, …)


leehart:
To do: add info

KellyLBennett:
Functions were written in colab to generate F3 and F4 statistics. These return the F statistic, SE value and Z statistic (determines whether comparison is significant). 

The F3 function was previously modified by @alimanfoo to squeeze the allele counts to ensure that all biallelic sites would be considered in analysis and to remove missing data. I also have a version of the F4 statistic function also modified with these enhancements. To implement these, the functions would require the addition of the remap_observed function in the example provided below. Would this be placed with the function in say, admixture_stats.py?

`@numba.njit
def remap_observed(ac_full):
    """Remap allele indices to remove unobserved alleles."""

    n_variants = ac_full.shape[0]
    n_alleles = ac_full.shape[1]

    # Create the output array - this is an allele mapping array,
    # that specifies how to recode allele indices.
    mapping = np.empty((n_variants, n_alleles), dtype=np.int32)
    mapping[:] = -1

    # Iterate over variants.
    for i in range(n_variants):

        # This will be the new index that we are mapping this allele to, if the
        # allele count is not zero.
        j_out = 0

        # Iterate over columns (alleles) in the input array.
        for j in range(n_alleles):

            # Access the count for the jth allele.
            c = ac_full[i, j]
            if c > 0:

                # We have found a non-zero allele count, remap the allele.
                mapping[i, j] = j_out
                j_out += 1

    return mapping`

`def patterson_f3(
    recipient_query,
    source1_query,
    source2_query,
    region,
    site_mask,
    n_jack,
    max_missing_an,
    segregating_mode,
    # TODO add min_cohort_size and max_cohort_size when these are supported
    # in the snp_allele_counts() function in malariagen_data.
):

    # Purely for conciseness, internally here we use the scikit-allel convention
    # of labelling the recipient population "C" and the source populations as
    # "A" and "B".

    # Find cohort sizes.
    nc = len(ag3.sample_metadata(sample_query=recipient_query))
    na = len(ag3.sample_metadata(sample_query=source1_query))
    nb = len(ag3.sample_metadata(sample_query=source2_query))
    print(f"Cohort sizes: {nc}, {na}, {nb}")
    n_chroms = (nc + na + nb) * 2

    # Compute allele counts for the three cohorts.
    acc = ag3.snp_allele_counts(
        sample_query=recipient_query,
        region=region,
        site_mask=site_mask,
    )
[...]

PR #1330, opened 2026-06-10
Admixture f3 and f4 statistics function.

KellyLBennett:
New branch with f3 and f4 admixture stats added including tests and example notebook. Resolves historical issue #705 . Please note this was created in VS code but due to a WSL issue new files and changes were then added to workbench to push from there.


malariagen_data/anoph/admixture_stats.py
```diff
@@ -0,0 +1,288 @@
...
+            sample_query=d_query,
+            region=region,
+            site_mask=site_mask,
+            cohort_size=cohort_size,
+            min_cohort_size=min_cohort_size,
+            max_cohort_size=max_cohort_size,
+        )
+
+        # Locate biallelic variants and deal with missingness.
+        ac = acc + aca + acb
```
jonbrenas, line 233: Why no `acd`?

malariagen_data/anoph/admixture_stats.py
```diff
@@ -0,0 +1,288 @@
...
+        loc_bi = allel.AlleleCountsArray(ac).is_biallelic()
+        ana = np.sum(acc, axis=1)
```
jonbrenas, line 235: That's probably not going to cause an error, but that's very confusing!

malariagen_data/anoph/admixture_stats.py
```diff
@@ -0,0 +1,288 @@
...
+        else:
+            print(f"Using {loc_seg.sum()} segregating variants for F4 calculation.")
+        aca_seg = acc_sqz[loc_seg]
+        acb_seg = aca_sqz[loc_seg]
+        acc_seg = acb_sqz[loc_seg]
+        acd_seg = acd_sqz[loc_seg]
+
+        # Compute f4 statistic.
+        blen = acc_seg.shape[0] // n_jack
+        f4, se, z, _, _ = allel.average_patterson_d(
```
jonbrenas, line 280: I am not an expert here, but I think D and f4 are slightly different. The code for `patterson_d` states that you can get f4 by ignoring the `den` but the `average_patterson_d` doesn't do that, and thus returns (as the name implies) D and not f4.

malariagen_data/anoph/admixture_stats.py
```diff
@@ -0,0 +1,288 @@
...
+        summary="""
+            Compute Patterson's f3 statistic for a specified recipient population and two source populations.
+        """,
+        returns="""
+            A NumPy float of the f3 value, standard error (SE) and Z-score.
+        """,
+    )
+    def patterson_f3(
+        self,
+        recipient_query: base_params.sample_query,
```
jonbrenas, line 62: Given the lack of symmetry between the various queries, giving them different parameter descriptions might be helpful.

malariagen_data/anoph/admixture_stats.py
```diff
@@ -0,0 +1,288 @@
...
+def _remap_observed(ac_full):
+    """Remap allele indices to remove unobserved alleles."""
+    n_variants = ac_full.shape[0]
+    n_alleles = ac_full.shape[1]
+    # Create the output array - this is an allele mapping array,
+    # that specifies how to recode allele indices.
+    mapping = np.empty((n_variants, n_alleles), dtype=np.int32)
+    mapping[:] = -1
+    # Iterate over variants.
+    for i in range(n_variants):
```
jonbrenas, line 23: This doesn't seem optimised but I don't have a better approach in mind right now.

tests/anoph/test_admixture_stats.py
```diff
@@ -0,0 +1,131 @@
...
+
+@parametrize_with_cases("fixture,api", cases=".")
+def test_patterson_f3(fixture, api: AnophelesAdmixtureAnalysis):
+    # Set up test parameters.
+    all_taxon = api.sample_metadata()["taxon"].dropna().unique().tolist()
+    taxa = np.random.choice(all_taxon, size=2, replace=False).tolist()
+
+    params = dict(
+        recipient_query=f"taxon == {taxa[0]!r}",
+        source1_query=f"taxon == {taxa[1]!r}",
```
jonbrenas, line 84: If `source1_query == source2_query`, isn't it technically testing f2?

tests/anoph/test_admixture_stats.py
```diff
@@ -0,0 +1,131 @@
...
+        max_cohort_size=50,
+    )
+
+    # Run patterson f3 function under test.
+    f3, se, z = api.patterson_f3(**params)
+    # Check results.
+    assert isinstance(f3, np.float64)
+    assert isinstance(se, np.float64)
+    assert isinstance(z, np.float64)
+    assert -0.1 <= f3 <= 1
```
jonbrenas, line 100: Patterson 2012 says that "The observation of a significantly negative value of f3(C; A, B) is thus evidence of complex phylogeny in C." which implies that f3 is not always >= -0.1, does it not?

jonbrenas (changes requested):
A lot of it is just comments, but the confusion around 'ana = np.sum(acc, axis=1)' needs changing, IMHO.

tests/anoph/test_admixture_stats.py
jonbrenas: Actually, now that I think about it, it circles back to my comment on f2 being tested as f2 cannot be significantly negative.

malariagen_data/anoph/admixture_stats.py
KellyLBennett: This is because D is the outgroup, its been a very long time since the stats function was written so my memory is hazy but I remember considering this point and I believe there would much less or not enough variation for analysis if the outgroup was also taken into account. We do not except them to share anywhere near as much variation as the other populations.

malariagen_data/anoph/admixture_stats.py
KellyLBennett: Please could you give some more explanation here/a suggested change. Thanks.

malariagen_data/anoph/admixture_stats.py
KellyLBennett: You are correct, the description is slightly off and should be revised to D.

malariagen_data/anoph/admixture_stats.py
KellyLBennett: Although it is called recipient query here, it is basically the same (just a sample query) and is more to avoid confusion over which population to test.

tests/anoph/test_admixture_stats.py
KellyLBennett: Unfortunately this is the only way to run tests on this function, since we do not have three taxa we can access to generate the simulated data in the repository.

tests/anoph/test_admixture_stats.py
KellyLBennett: I can change this to -1 to 1 to be safe.

malariagen_data/anoph/admixture_stats.py
jonbrenas: Well, `ana` should be `np.sum(aca, axis=1)`, not `np.sum(acc, axis=1)`, I believe.

malariagen_data/anoph/admixture_stats.py
jonbrenas: I think a "naive" user asking for these function's help should be given more info on the assumptions associated to each query than the current: 
"""
A pandas query string to be evaluated against the sample metadata, to
    select samples to be included in the returned data. E.g., "country ==
    'Uganda'". If the query returns zero results, a warning will be
    emitted with fuzzy-match suggestions for possible typos or case
    mismatches.
"""
For instance, while it makes sense that cohort D is the outgroup in f4/D, I think it is worth telling the user explicitly, in particular when it is treated differently.

tests/anoph/test_admixture_stats.py
jonbrenas: Couldn't you use `cohort_admin2_month` instead of taxa? There are at least 3, I believe.

PR #1346, opened 2026-08-26
GH1345 Update terms of use text and URL to Vector Observatory 

moylstmed360:
* Resolves #1345

PR #1347, opened 2026-09-01
Correct NumPy version in CONTRIBUTING.md CI note

moylstmed360:
CONTRIBUTING.md stated that CI tests run on Python 3.10 with NumPy 1.26.4. The Matrix in test.yml pins numpy >= 2.0.2 < 2.1, so the note was out of date. Also opened as a second PR while still a first-time contributor, to test whether the workflow approval issue Jon hit on #1346 recurs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -234,7 +234,7 @@ poetry run pytest -v tests --typeguard-packages=malariagen_data,malariagen_data.
 ### Review process
 
 - PRs require approval from a project maintainer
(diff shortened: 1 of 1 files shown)


jonbrenas:
It looks like the tests ran without any issue.

jonbrenas (changes requested):
I guess it would more accurate to describe each test that needs to pass.

moylstmed360:
Thanks Jon, that's fair. Updated to describe all eight required checks rather than picking one example. Let me know if you'd word it differently.

Issue #1349, opened 2026-09-07
`wgs_run_accessions()` not found in docs


ahernank:
The function `wgs_run_accessions` is not currently documented in the public docs of the Ag3 class. It would be great to get this linked to the regular docs (https://malariagen.github.io/malariagen-data-python/latest/Ag3.html).

ahernank:
Just to note that @kebarr will work on this one.

kebarr:
This appears to be an omission from the rst files used to configure Sphinx, I have added this command to the docs for each top level data class and will look for other such omissions and add them to the rst files for the data classes they are implemented for.

PR #1351, opened 2026-09-14
Documentation update- fix to issue 1349

kebarr:
Fix for[ issue 1349](https://github.com/malariagen/malariagen-data-python/issues/1349)

Added to other rst files data sets for which "wgs_run_accessions" can be run. Also added similarly missing methods for which "@doc" is present in the code, for the datasets to which they can be run. This was limited to the "Basic Data Analysis," "Reference genome data access" and "Sample metadata access" sections, there may be further analysis-specific methods for each dataset missing but I decided it would be better to address this in a separate issue. 

Whilst fixing this I noticed that the A darlingi photo was missing from the index page, this is now showing on the index. 

There were also some warnings from Sphinx, which I have also fixed. 

At some point I ran "ruff format" to fix a failing commit hook, this touched some of the notebooks and tests in the repo. I have checked the tests and they are still passing, the notebook changes apply to ld_pruning.ipynb and  plot_haplotype_clustering.ipynb and effect string escaping only. 

I have attached the html source obtained by running "make html" with this version. 

[html.zip](https://github.com/user-attachments/files/32194309/html.zip)


kebarr:
Closing as will be subsumed by PR hopefully coming tomorrow

PR #1354, opened 2026-09-16
Add burst clustering option to plot_haplotype_clustering_advanced

### LINKED ISSUE #1353: Haplotype clustering without a cut height in plot_haplotype_clustering_advanced
`plot_haplotype_clustering_advanced()` forms flat clusters by cutting the dendrogram at `cluster_threshold`. That works when you already know the right height, but the height is hard to choose and does not transfer:

- it confounds when a clade coalesced with how fast it did, so one height cannot suit both a young sweep and an old one in the same dendrogram;
- the clusters it returns change with sample size. On Ag3 data at one locus, a cut at a fixed fraction of the window's diversity gives ~5 clusters at 100 haplotypes and 22-25 at 2,200, on the same region.

Instead of a height, each clade is scored on whether its internal diversity is far below the neutral expectation for the window, and the best set of non-overlapping clades is selected in one bottom-up pass. Because the score sums a per-pair log-likelihood ratio against a background model, clades outside any cluster contribute zero and the selection is invariant to sample size by construction.

sanjaynagi:
Closes #1353.

Adds `cluster_method="burst"` to `plot_haplotype_clustering_advanced()`. The default stays `"cut"`, so nothing changes unless you ask for it.

### Why

Cutting the dendrogram needs a height, and the height is the problem. It confounds *when* a clade coalesced with *how fast* it did, so no single value suits a young and an old sweep in the same dendrogram, and the clusters it returns move with sample size. On Ag3 data at one locus, a cut at a fixed fraction of the window's diversity gives ~5 clusters at 100 haplotypes and 22–25 at 2,200 on the same region.

### What it does instead

Each clade is scored on how much better its internal pairwise distances fit a low-diversity cluster model than the window's background diversity, and the best set of non-overlapping clades is chosen in one bottom-up pass. Since the reference model is "every pair is background", pairs outside every cluster contribute exactly zero, so the objective is separable and the score scales with the number of pairs — which is what makes the selection invariant to sample size rather than merely robust to it.

Where a clade splits in two, the two daughters are kept as one cluster while describing them with a single diversity is cheap. For the geometric family that cost depends only on the *ratio* of the between-daughter to within-daughter distance, so the question asked is whether the daughters separated recently relative to their own age, not whether the split is deeper than some height.

### Evidence

Developed for the selection atlas and validated there against data the clustering never sees:

- copy-number alleles called from discordant read pairs map onto single clusters (sensitivity + specificity − 1 of 0.98 for Cyp6aap_Dup10, 0.96 Cyp9k1_Dup8, 0.95 Cyp9k1_Dup4, 0.90 Cyp9k1_Dup11);
- common non-synonymous variants reach within-cluster purity of 0.96–1.00 across six resistance loci and eight cohorts;
- at Vgsc the pooled sample splits into one L995F cluster spanning five West African countries and both species, with separate Ugandan and Cameroonian groups — the published picture;
- a clean control region yields no clusters.

### Notes for review

- `cluster_background` lets you supply the neutral expected pairwise distance. It is estimated from the data otherwise, by collapsing near-identical haplotypes, which assumes some unswept haplotypes are present. That assumption fails where a region is swept end to end (Vgsc in some cohorts), hence the argument.
- Requires `distance_metric="hamming"`; raises a clear error for `"dxy"` since the model is defined on counts of differing SNPs.
- The algorithm lives in a new module `hapclust_burst.py` (numpy/scipy only, no new dependencies) so it can be tested without the API.
- Tests: 8 unit tests on the algorithm, plus two through the public function using the simulated fixtures. `tests/anoph/test_hapclust.py` passes in full locally (10 tests), and `ruff` (the pinned 0.1.8) is clean.


kebarr:
Hi Sanjay, I've been assigned to review this, please bear with me whilst I familiarise myself with the project and this specific clustering method.

One immediate thing that jumps out is that this is quite a computationally intensive analysis- it would therefore be useful to cache the results in a similar manner to elsewhere in the code base, such as in `haplotype_pairwise_distances` using `results_cache_get/set` so that operations aren't repeated unnecessarily. 

I will hopefully get back to you with something more constructive probably early next week. 

Cheers,
Katie

sanjaynagi:
Hi @kebarr dont worry about reviewing this yet, I'm gonna do some further work on it soon. 

Have converted the PR to a draft for now.

